PR #32, opened 2015-01-09
Improve liveness analysis

serpilliere:
Support liveness analysis in incomplete execution flow graph.

### DIFF
--- a/miasm2/core/graph.py
+++ b/miasm2/core/graph.py
@@ -27,6 +27,14 @@ class DiGraph:
         self._nodes_to[n] = []
         self._nodes_from[n] = []
 
+    def del_node(self, n):
+        if n in self._nodes:
+            self._nodes.remove(n)
+        for pred in self.predecessors(n):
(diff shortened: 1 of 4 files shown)


miasm2/core/graph.py
```diff
@@ -27,6 +27,14 @@ def add_node(self, n):
         self._nodes_to[n] = []
         self._nodes_from[n] = []
 
+    def del_node(self, n):
```
commial: Can you use `node` instead of  `n`, as in `get_all_parents` ?
And add a (tiny) docstring ?

miasm2/ir/ir.py
```diff
@@ -52,6 +52,17 @@ def get_dst(self):
         self._dst = dst
         return dst
 
+    def set_dst(self, value):
```
commial: Because this is a public method (not an internal one), can you specify its behaviour in a (again tiny) docstring ?

PR #38, opened 2015-01-13
Graph: add dominators computation

serpilliere:
Add dominator computation
(see https://en.wikipedia.org/wiki/Dominator_(graph_theory))

### DIFF
--- a/miasm2/core/graph.py
+++ b/miasm2/core/graph.py
@@ -1,4 +1,7 @@
+from collections import defaultdict
+
 class DiGraph:
+    """Implementation of directed graph"""
 
     def __init__(self):
         self._nodes = set()
@@ -84,6 +87,14 @@ class DiGraph:
     def leaves(self):
         return [x for x in self.leaves_iter()]
 
+    def heads_iter(self):
+        for node in self._nodes:
+            if len(self._nodes_from[node]) == 0:
+                yield node
+
+    def heads(self):
+        return [node for node in self.heads_iter()]
+
     def roots_iter(self):
         for n in self._nodes:
             if len(self._nodes_from[n]) == 0:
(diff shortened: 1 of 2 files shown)


miasm2/core/graph.py
```diff
@@ -1,3 +1,5 @@
+from collections import defaultdict
+
 class DiGraph:
```
commial: A little *docstring* for this `class DiGraph` would be welcome.

miasm2/core/graph.py
```diff
@@ -84,6 +86,14 @@ def leaves_iter(self):
     def leaves(self):
         return [x for x in self.leaves_iter()]
 
+    def heads_iter(self):
+        for node in self._nodes:
+            if len(self._nodes_from[node]) == 0:
+                yield node
+
+    def heads(self):
+        return [x for x in self.heads_iter()]
```
commial: `x` stands here for a `node` ? If so, can you rename it ?

miasm2/core/graph.py
```diff
@@ -149,3 +159,44 @@ def dot(self):
                                                 self.edge2str(a, b))
         out += "}"
         return out
+
+
+
+    def compute_dominators(self):
+        """Compute dominators of the graph"""
+        nodes = self.nodes()
+        heads = self.heads()
```
commial: You have add the `heads_iter` methods, which returns an iterator. If you use `heads` method instead, the iterator will be consumed *before* returning to `compute_dominators`. That way, the iterator's interest is lost.

I think you can safely replace `heads` with `heads_iter` here.

`Difference` still working:
``` Python
>>> set("test").difference(xrange(3))
set(['s', 'e', 't'])
```

miasm2/core/graph.py
```diff
@@ -149,3 +159,44 @@ def dot(self):
...
+
+        dominators = defaultdict(set)
+        for n in self.nodes():
```
commial: Below, you use `node` as temporary variable. If it is relevant here too, can you develop `n` too ?

commial:
Can you add to your PR a regression test which use the dominator computation, for instance on the Wikipedia's example ? It seems to contain most of possible cases.

I add some notes to your code too (not very significant).

miasm2/core/graph.py
serpilliere: Is this a comment on code optimisation?
Do you think it is more interesting to use at least 4 times an iterator generation than generating an object once and referencing this non modified object?
The iterator will be called at least 4 times:
* `heads = self.heads()`
* `for n in heads:`
* `todo = nodes.difference(heads)`
* `if node in heads:` (in a loop)

commial: Ok, if you prefer to reduce the CPU footprint over the memory one, it sounds good for me in this case.
I wanted to ensure it was not a mistake.

PR #41, opened 2015-01-18
TranslatorZ3

fmonjalet:
Hi,

This is a Translator I wrote to translate a Miasm Expr into a Z3 Expr. It supports memory abstraction for any size of address. It also supports big and little endian.
One good thing with the Z3 python binding is that it can generate SMT expressions understandable by other SMT solvers if you need.

Here is a (silly) example of usage :

```python
expr = ExprRandom()                                                             
z3expr = Translator.to_language('z3').from_expr(expr)                             
print z3.simplify(z3expr)                                                       

# Solve a constraint                                      
solver = z3.Solver()                                                         
solver.add(z3expr == 10)                                                        
solver.check()                                                               
print solver.model()   
```

Some unit tests are included at the end of the file. This is clearly not the way to go. If the PR is merged, would you like include these into your existing set ? If yes, do not hesitate to give directions so that I can finalize this PR with a proper test file.


miasm2/ir/translators/z3_ir.py
```diff
@@ -0,0 +1,308 @@
...
+import miasm2.expression.expression as m2_expr
+from miasm2.ir.translators.translator import Translator
+
+log = logging.getLogger("translator_z3")
+console_handler = logging.StreamHandler()
+console_handler.setFormatter(logging.Formatter("%(levelname)-5s: %(message)s"))
+log.addHandler(console_handler)
+log.setLevel(logging.WARNING)
+
+class Z3Mem(object):
```
commial: It's a good thing that two accesses to a same address with different sizes return two different arrays.

But it would be interesting to abstract memory in a way which allow: @32[EAX] == @16[EAX] . @16[EAX + 2].
It seems to be a frequent case, especially in cryptography's algorithms (a nice Z3 use case).

A solution could be a memory abstraction using an array of bytes, a DWORD access becoming a 4 elements subarray.
Unfortunately, I'm not enough familiar with Z3 to offer you a better idea...

miasm2/ir/translators/z3_ir.py
```diff
@@ -0,0 +1,308 @@
...
+            # Clean cache and Z3Mem if this call is the root call
+            if del_cache:
+                cls._cache = None
+            if del_mem:
+                cls._mem = None
+
+# Register the class
+Translator.register(TranslatorZ3)
+
+if __name__ == '__main__':
```
commial: Miasm tests are located in the `test` directory.

This directory tries to follow the `miasm2` package architecture.
So, your tests could be moved to a `test/ir/translators/z3.py`file.

To add them to the test framework, you have to add the unit test command line in `test/test_all.py`.
For instance:
``` Python
## IR
for script in ["ir2C.py",
               "symbexec.py",
               ]:
    testset += RegressionTest([script], base_dir="ir")

testset += RegressionTest(["z3.py"], base_dir="ir/translators")
```

In addition, the z3 import fail should be handled.
To achieve this, you can define a *tag* for your z3 test, for instance `z3`, and assign it to your test (`tags=[TAGS["z3"]]`).

That way, as LLVM, you can disable the test case by using:
``` Python
try:
    import z3
except ImportError:
    # Print warning / sleep ...        
    if TAGS["z3"] not in exclude_tags:
            exclude_tags.append(TAGS["z3"])
```

miasm2/ir/translators/z3_ir.py
fmonjalet: In fact, it is actually the case : the implementation allows  @32[EAX] == @16[EAX] . @16[EAX + 2] for a big endian architecture.
For a little endian architecture, the equation becomes @32[EAX] == @16[EAX + 2] . @16[EAX].
This way, the memory model is coherent with memory accesses in real architectures, as far as I know.

You can verify it with this code (equiv is defined in the test section of z3_ir.py) :
```python
    mem = Z3Mem(endianness='<') # little endian                                  
    eax = z3.BitVec('EAX', 32)                                                                                                   
    assert equiv(                                                                
             # @32[EAX]                                                          
             mem.get(eax, 32),                                                   
             # @16[EAX+2] . @16[EAX]                                             
             z3.Concat(mem.get(eax+2, 16),                                       
                       mem.get(eax, 16)))
```

Any memory access is a *concatenation of bytes-sized memory access*, the direction of the concatenation depends on the endianness. To be totally precise, it is a slice of this, to handle non 8bit-aligned accesses, even though it could not make sense it is necessary to pass the automatic tests. That said, non 8bit aligned memory accesses are not tested yet, I suspect they could be wrong in some cases.

But @*16*[**EAX**] != @*16*[**AX**] even if EAX = ZeroExtend(AX), because as these addresses does not have the same size, they refer two different address spaces. This is quite a borderline case, but TranslatorZ3 should be able to translate any Expression, even if it does not necessarily make sense in a real architecture.
This can be verified with this code :

```python
    ax = z3.BitVec('AX', 16)                                                     
    assert not equiv(     
            # @16[EAX] with EAX = ZeroExtend(AX)
            mem.get(z3.ZeroExt(16, ax), 16),                                     
            mem.get(ax, 16))
```

This remarks raises two points:
- Maybe the doc is not clear enough, I will try to express it better.
- Maybe it would be usefull to allow choosing the endianness of the access in the Z3Mem.get method too, to allow getting large arrays, which you want to get with the "big endian" style. Another solution would be to add a method to get a concatenation of bytes rather than a number.

Does this solution fit the needs you raised? Do you think the API should be modified?

miasm2/ir/translators/z3_ir.py
fmonjalet: Thanks for all the details! That will be done.

miasm2/ir/translators/z3_ir.py
commial: Thanks for your details !

Actually, I misunderstood the purpose of the argument `size`. But your docstring `@size: integer, size in bit of addresses in the memory to get.` is enough; I didn't see it in the first commit.

So, the current implementation sounds good for me.

serpilliere:
Some modifications have been done on the main translator code. Please, can you rebase your PR?

fmonjalet:
The PR is now rebased and the tests have been moved to Miasm's test suite.

commial:
Sounds good for me. Thanks for your code and your reactivity !

PR #46, opened 2015-01-23
arm: Add instruction umlal

laanwj:
umull, smull, smull and smlal are there, but not umlal
(unsigned multiply and accumulate long)


laanwj:
Sure, I'll add a test.

laanwj:
Re-pushed, now includes a test (verified with arm-linux-gnueabi-as)

PR #47, opened 2015-01-26
arm: Add umull,umlal,smull,smlal to sem

laanwj:
Also add tests.

For extra verification I've run the secp256k1 inner field functions ( https://github.com/laanwj/secp256k1/blob/2014_12_field10x26_arm_asm/src/asm/field_10x26_arm.s - lots of `umull`/`umlal`) using `test_jit_arm`, and it produced expected outputs for my test vectors.


serpilliere:
Nice work, thanks for your contribution!

PR #48, opened 2015-01-26
Arm: fix execflow code

serpilliere:
Do execflow tests using a white list


commial:
Do you have any plans to apply this methods to others architectures ? If so, in this PR ?

PR #52, opened 2015-02-09
Ida symbol exec

commial:
This PR add a plug-in example for IDA, based on Miasm symbolic execution engine. It also features the `VarExpander` possibility.

For instance, on `samples/sc_connect_back.bin`, one selects the targeted area.
select

Then, by calling `symbolic_exec()` or pressing `F3`, the new view `Symbolic Execution ...` is created.
symbolexec

If a line is too complicated, one can expand it by pressing `E` or `Right click > Expand`.
varexpanded

The common code between IDA examples has been moved to `samples/ida/utils.py`.

### DIFF
--- a/example/ida/graph_ir.py
+++ b/example/ida/graph_ir.py
@@ -16,44 +16,7 @@ from miasm2.analysis.machine import Machine
 from miasm2.analysis.data_analysis import intra_bloc_flow_raw, inter_bloc_flow
 from miasm2.analysis.data_analysis import intra_bloc_flow_symbexec
 
-
-
-def expr2colorstr(ir_arch, e):
-    # print "XXX", e
-    if isinstance(e, ExprId):
-        s = str(e)
-        if e in ir_arch.arch.regs.all_regs_ids:
-            s = idaapi.COLSTR(s, idaapi.SCOLOR_REG)
-    elif isinstance(e, ExprInt):
-        s = str(e)
-        s = idaapi.COLSTR(s, idaapi.SCOLOR_NUMBER)
-    elif isinstance(e, ExprMem):
-        s = '@%d[%s]' % (e.size, expr2colorstr(ir_arch, e.arg))
-    elif isinstance(e, ExprOp):
-        out = []
-        for a in e.args:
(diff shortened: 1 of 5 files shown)


p-l-:
Really nice work!

aguinet:
yep nice one :) !

commial:
Thanks you !
Actually, it's just a basic binding with already available features.

I've add the possibility to get back an equivalent Miasm2 expression in its Python form.

For instance, by pressing `P` or `Right click > Get in Python` on `nf`:
translateform

That is to say:
```Python
>>> expr = ExprAff(ExprId('nf', size=1), ExprSlice(ExprOp('+', ExprId('EDX_init', size=32), ExprMem(ExprOp('+', ExprMem(ExprId('ESP_init', size=32), size=32), ExprInt_fromsize(32, 0x24)), size=32)), 31, 32))
>>> print expr
nf = (EDX_init+@32[(@32[ESP_init]+0x24)])[31:32]
```

commial:
There is no reason to limit the translation to `Miasm`translator.

Add a ComboBox to choose the translator (`T` or `Right click > Translator`).
multi

commial:
An infinite loop was reached in a corner case of Variable Identifier (reported by @serpilliere).

One of the commit also add the prefix feature, which result in:
```Python
of_v0 = @32[@32[ESP_init] + 0x24]
of_v1 = EDX_init
of = ((of_v1 ^ of_v0 ^ 0xFFFFFFFF) & (of_v0 ^ (of_v1 + of_v0)))[31:32]
cf_v0 = EDX_init
cf_v1 = @32[@32[ESP_init] + 0x24]
cf_v2 = cf_v0 + cf_v1
cf = (cf_v0 ^ cf_v1 ^ ((cf_v0 ^ cf_v1 ^ 0xFFFFFFFF) & (cf_v1 ^ cf_v2)) ^ cf_v2)[31:32]
```
That way, the IDA highlighting of variables can be used in a clever way.

serpilliere:
The module needs to be imported twice due to a strange bug between IDA and Miasm.

PR #63, opened 2015-02-11
Bin stream no len

serpilliere:
The `__len__` cannot be used any more in `bin_stream`: Python returns an int object, which will cap values to `0x7FFFFFFF` on 32 bit systems. A binary can have a base address higher than this, making it impossible to handle such programs.

The `__len__` is replaced by a call to `getlen`.

:warning: Elfesteem has been updated to support `max_addr` API in order to retrieve the elf/pe max virtual address. See revision `772380e060e8` on [elfesteem](https://code.google.com/p/elfesteem/source/detail?r=772380e060e856c98b65ba327ab5f7edb2a2bac4).

### DIFF
--- a/miasm2/core/bin_stream.py
+++ b/miasm2/core/bin_stream.py
@@ -32,17 +32,21 @@ class bin_stream(object):
         return self.bin[start:start + l]
 
     def getbits(self, start, n):
+        """Return the bits from the bit stream
+        @start: the offset in bits
+        @n: number of bits to read
+        """
         if not n:
             return 0
         o = 0
         if n > self.getlen() * 8:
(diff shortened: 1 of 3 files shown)


miasm2/core/bin_stream.py
```diff
@@ -91,85 +90,73 @@ def __str__(self):
...
+    def getlen(self):
+        return self.l - (self.offset + self.shift)
+
 
-class bin_stream_pe(bin_stream):
 
-    def __init__(self, bin="", offset=0L):
+class bin_stream_container(bin_stream):
+
+    def __init__(self, virt_view="", offset=0L):
```
commial: The default case of `virt_view` (`""`) will crash on `self.l = virt_view.max_addr()`.

PR #70, opened 2015-02-15
MIPS32: add BNEL instruction

aguinet:
Add the BNEL instruction for MIPS32.

Not sure how to handle the fact that delayslot must be set to 0 for such
instruction. Indeed, according to the reference, the delay slot must be
executed only if the branch is taken!


serpilliere:
I have to test some code with the way you use the delayslot: At the very beginning, the attribute should have been a class attribute but the instance attributes seems clever. But the IR generation should use this. 
I will keep you in touch

aguinet:
Okay thanks ! I have some other instructions with the same "issue", so I'll wait for your final callto add them to the PR... :)

aguinet:
Actually I think my patch does not work, as it does the opposite of what it is supposed to do. Indeed, the delay slot will be executed only if the branch *is not* taken... I don't have any solution though :/

serpilliere:
I am closing this issue as It's need core modification in Miasm. I am opening issue #211 to monitor your request.

PR #74, opened 2015-02-17
Comment mips32

commial:
Add docstrings in MIPS32 semantic. The goal is to help to maintain semantic, and to offer a way to retrieve the meaning of an architecture operations.

In addition, the wildcard imports of the module have been removed.


serpilliere:
Maybe the mnemonic description could be used in a future "auto-comment" feature ?

commial:
This is exactly what I had in mind.

PR #73, opened 2015-02-16
Refactor API arguments

p-l-:
This adds the possibility to call `.func_args_*()` methods with an arguments list instead of the arguments count as the parameter, and get a `collections.namedtuple`. It is still possible to use the arguments count.

The calls in `os_dep/win_api_x86_32.py` have been updated, and some bugs have been fixed there.

Thanks to @commial (kamoul0x) for your help & time.

PR #78, opened 2015-02-20
Use named arguments in os_dep/linux

p-l-:
This tiny PR is the last step to complete the work started in #73.

### DIFF
--- a/miasm2/os_dep/linux_stdlib.py
+++ b/miasm2/os_dep/linux_stdlib.py
@@ -12,8 +12,8 @@ def xxx_isprint(jitter):
 
     checks for any printable character including space.
     '''
-    ret_addr, (c,)  = jitter.func_args_stdcall(1)
-    ret = chr(c & 0xFF) in printable and 1 or 0
+    ret_addr, args  = jitter.func_args_stdcall(['c'])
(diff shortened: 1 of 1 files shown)


miasm2/os_dep/linux_stdlib.py
```diff
@@ -36,13 +36,13 @@ def xxx_puts(jitter):
...
     writes the string s and a trailing newline to stdout.
     '''
-    ret_addr, (s,) = jitter.func_args_stdcall(1)
-    while True:
-        c = jitter.vm.get_mem(s, 1)
-        s += 1
-        if c == '\x00':
-            break
-        stdout.write(c)
+    ret_addr, args = jitter.func_args_stdcall(['s'])
```
commial: Are you sure that this is the better pattern for an equivalent of a `do... while` statement ?
It seems to introduce duplicated lines...

What about a pattern like the following ?
```Python
condition = True
while condition:
    do_stuff()
    condition = eval_cond()
```

p-l-: This introduces **one** duplicated line. I like this code better than:

    char = '\xff' # or anything else than '\x00'
    while char != '\x00':
        char = jitter.vm.get_mem(index, 1)
        stdout.write(char)
        index += 1

PR #81, opened 2015-02-20
x86 fix for PUSHFW / POPFW instructions

ggqb:
Hi,

There is a problem with PUSHFW and POPFW instructions.
PUSHF is equivalent to PUSHFW in the Intel documentation, and they work with only 16 bits of flags.
In miasm, they are differently declared and used.

For example PUSHFW is declared as:
```python
addop("pushf", [bs8(0x9c), bs_opmode16])
```
And the dictionnary which contains all instructions contains PUSHFW only [here](https://github.com/cea-sec/miasm/blob/master/miasm2/arch/x86/sem.py#L3207-L3210). It is the same
problem for POPFW. This last one has also a problem in its semantic definition where 
the undefined 'esp' variable is used [here](https://github.com/cea-sec/miasm/blob/master/miasm2/arch/x86/sem.py#L972-L987).

Patch is provided (tested/working)

I also fixed a string formatting problem during a log warning, which cause a crash instead of just the warning message.


ggqb:
Yep, fixed!

Thx.

serpilliere:
It seems you didn't fix the regression test for arch x86:
```
16 '\x9d'
dis args []
POPF
POPFW      
  File "arch.py", line 1876, in <module>
    assert(str(mn).strip() == s)
```

Can you fix it in the same PR?
For POPF:
```
    (m16, "00000000    POPF",
     "9d"),
```
and for PUSHF:
```
    (m16, "00000000    PUSHF",
     "9c"),
```

ggqb:
Indeed :) done.

serpilliere:
Thank you!

PR #83, opened 2015-02-22
Expr immutable

serpilliere:
Make expression immutable.
This may avoid bugs on the hash cache computation, and be a start point to correct hash collision.

### DIFF
--- a/miasm2/arch/arm/sem.py
+++ b/miasm2/arch/arm/sem.py
@@ -1191,11 +1191,11 @@ class ir_arml(ir):
             if args[-1].op == 'rrx':
                 args[-1] = ExprCompose(
                     [(args[-1].args[0][1:], 0, 31), (cf, 31, 32)])
-            elif args[-1].op in ['<<', '>>', '<<a', 'a>>', '<<<', '>>>']:
-                if isinstance(args[-1].args[-1], ExprId):
-                    args[-1] = ExprOp(args[-1].op,
-                                      args[-1].args[0],
-                                      args[-1].args[-1][:8].zeroExtend(32))
+            elif (args[-1].op in ['<<', '>>', '<<a', 'a>>', '<<<', '>>>'] and
+                  isinstance(args[-1].args[-1], ExprId)):
+                args[-1] = ExprOp(args[-1].op,
+                                  args[-1].args[0],
+                                  args[-1].args[-1][:8].zeroExtend(32))
         instr_ir, extra_ir = get_mnemo_expr(self, instr, *args)
         # if self.name.startswith('B'):
         #    return instr_ir, extra_ir
--- a/miasm2/arch/arm/sem.py
+++ b/miasm2/arch/arm/sem.py
@@ -1191,10 +1191,11 @@ class ir_arml(ir):
             if args[-1].op == 'rrx':
                 args[-1] = ExprCompose(
                     [(args[-1].args[0][1:], 0, 31), (cf, 31, 32)])
-            elif (args[-1].op in ['<<', '>>', '<<a', 'a>>', '<<<', '>>>'] and
-                  isinstance(args[-1].args[-1], ExprId)):
-                args[-1].args = args[-1].args[:-1] + (
-                    args[-1].args[-1][:8].zeroExtend(32),)
+            elif args[-1].op in ['<<', '>>', '<<a', 'a>>', '<<<', '>>>']:
+                if isinstance(args[-1].args[-1], ExprId):
+                    args[-1] = ExprOp(args[-1].op,
+                                      args[-1].args[0],
+                                      args[-1].args[-1][:8].zeroExtend(32))
         instr_ir, extra_ir = get_mnemo_expr(self, instr, *args)
         # if self.name.startswith('B'):
         #    return instr_ir, extra_ir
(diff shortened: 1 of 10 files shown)


miasm2/arch/arm/sem.py
```diff
@@ -1191,10 +1191,11 @@ def get_ir(self, instr):
             if args[-1].op == 'rrx':
                 args[-1] = ExprCompose(
                     [(args[-1].args[0][1:], 0, 31), (cf, 31, 32)])
-            elif (args[-1].op in ['<<', '>>', '<<a', 'a>>', '<<<', '>>>'] and
-                  isinstance(args[-1].args[-1], ExprId)):
-                args[-1].args = args[-1].args[:-1] + (
-                    args[-1].args[-1][:8].zeroExtend(32),)
+            elif args[-1].op in ['<<', '>>', '<<a', 'a>>', '<<<', '>>>']:
+                if isinstance(args[-1].args[-1], ExprId):
```
commial: Why do you have change the `and` statement for a sub-`if` ?

miasm2/expression/expression.py
```diff
@@ -331,12 +349,12 @@ def get_r(self, mem_read=False, cst_read=False):
     def get_w(self):
         return set()
 
+    def exprhash(self):
```
commial: For the same reason, can you add a docstring here ?

miasm2/expression/expression.py
```diff
@@ -399,7 +418,7 @@ def visit(self, cb, tv=None):
         return self
 
     def copy(self):
-        return ExprId(self.name, self._size)
+        return ExprId(self.name, self.size)
```
commial: In my opinion, you could use a private attribute (`_size`) inside this method and avoid an indirection. It does not break the encapsulation.

miasm2/arch/arm/sem.py
serpilliere: When I corrected the code, it took me some time to read the test, which is a bit long. So I decided to split it in 2 sub if.

miasm2/expression/expression.py
```diff
@@ -474,9 +496,9 @@ def get_modified_slice(self):
         modified_s = []
         for x in self.src.args:
             if (not isinstance(x[0], ExprSlice) or
-                x[0].arg != dst    or
+                x[0].arg != dst or
                 x[1] != x[0].start or
-                x[2] != x[0].stop):
+                    x[2] != x[0].stop):
```
commial: I think you have a bad tabulation here.

miasm2/expression/expression.py
```diff
@@ -850,23 +896,30 @@ def __init__(self, args):
         for e, a, b in args:
             assert(a >= 0 and b >= 0)
             o.append(tuple([e, a, b]))
-        self.args = tuple(o)
+        self.__args = tuple(o)
+
+        self.__size = max([x[2]
```
commial: The arguments could be ordered at the instanciation

miasm2/expression/expression.py
serpilliere: In fact *each* attribute could be manipulated using its private version.
But don't you think it's preferable for readability to use the common name?
It could be:
```python
def copy(self):
    return ExprId(self.__name, self.__size)
```

instead of:
```python
def copy(self):
    return ExprId(self.name, self.size)
```

miasm2/expression/expression.py
serpilliere: The code has been generated using autopep8, so it *may* be correct.

miasm2/expression/expression.py
serpilliere: May you clarify your thinking?

commial: Thanks to your patches, `ExprCompose` are now immutable. Therefore, the `args` attribute is immutable too, and set only during the `__init__`.

That way, `args` attribute could be ordered (non-overlapping intervals). And then, this code snippet could be re-write:
```Python
self.__size = self.args[-1][-1] - self.args[0][0]
```

In addition, we can now improve the comparison between `ExprCompose`, because we currently have:
```Python
from miasm2.expression.expression import ExprCompose, ExprInt32
e1 = ExprCompose([(ExprInt32(1), 0, 32), (ExprInt32(2), 32, 64)])
e2 = ExprCompose([(ExprInt32(1), 0, 32), (ExprInt32(2), 32, 64)][::-1])

e1 == e2
# False, but should be True
```

It is a direct consequence of your patch, so in my opinion, you can add it in this PR.

miasm2/expression/expression.py
commial: Actually, `autopep8` on:
```Python
if (not isinstance(x[0], ExprSlice) or
    x[0].arg != dst or
    x[1] != x[0].start or
        x[2] != x[0].stop):
```

Does not change anything (as you remark).
But, on this code:
```Python
if (not isinstance(x[0], ExprSlice) or
    x[0].arg != dst or
    x[1] != x[0].start or
    x[2] != x[0].stop):
```
It results in:
```Python
if (not isinstance(x[0], ExprSlice) or
        x[0].arg != dst or
        x[1] != x[0].start or
        x[2] != x[0].stop):
```

Which is, for me, easier to read and homogeneous.

Issue #29, opened 2015-01-08
Jitter (TCC) : Memory leak


ggqb:
Hi,

I recently jitted an heavy piece of code and i didn't manage to do it in one shot. It seems there is memory leak problems in the TCC jitter. 

It is possible to limit the impact by disabling the usage of caches [here](https://github.com/cea-sec/miasm/blob/master/miasm2/jitter/jitcore_tcc.py#L146-L150):
```python
self.tcc_states.append(tcc_state)
jcode = jit_tcc_code(mcode)
self.lbl2jitbloc[label.offset] = mcode
self.addr2obj[label.offset] = jcode
self.addr2objref[label.offset] = objref(jcode)
```
... and periodically clear the **self.lbl2jitbloc** list.

But the memory is still growing. By doing some analysis using python built-in tools, it seems the highest memory usage object is **py_tuples**, so probably objects created native side but never freed python side...

commial:
I've tried to reproduce and fix the issue you encountered.

My test case is `python jitter/unpack_upx.py samples/box_upx.exe`, and I use measures coming from `valgrind`.
Without any changes, I got `definitely lost: 6,904 bytes in 4 blocks` coming from 3 different sources.

After the PR #84 , I got `definitely lost: 6,624 bytes in 2 blocks` coming from only one source.
After investigations, the leaked memory come from blocks JITted with TCC (in so-called `TCCState`).

The PR #85 introduces the `BoundedDict` to set an upper bound to avoid the infinite growth of the memory footprint.
In addition, when removing blocks addresses from the dictionary, the corresponding `TCCStates` are freed thanks to `tcc_end`.

By forcing the deletion of the `BoundedDict` at the end of the script (to trigger `__del__`) with `del sb.jitter.jit.lbl2jitbloc`, I finally got `0 bytes in 0 blocks`.

But for a reason I'm still investigating, its `__del__` method is not called every time a script ends.
As this is only important while leaving Python, it is no more relevant for your case.

Can you retry on your test case ?

PR #85, opened 2015-02-23
Bounded dict

commial:
Introduce a new structure, called `BoundedDict`.

A bounded dict is a dictionnary limited in size.
To reduce combinatory cost, once an upper limit `max_size` is reached, `max_size - min_size` elements are suppressed.
The targeted elements are the less accessed.
One can define a callback called when an element is removed.

This structure is now used in `Jitcore` to keep the list of `label` -> `Jitted block addr`, with an arbitrary upper bound of `10000` (open to comments).
That way, the memory footprint should be reduced (as described in #29 ).

In addition, when the `Jitcore_TCC` has to delete a reference on a jitted block, the latter is freed thanks to `tcc_end` (through the deletion callback).

### DIFF
--- a/miasm2/core/utils.py
+++ b/miasm2/core/utils.py
@@ -1,5 +1,6 @@
 import struct
 import inspect
+import UserDict
 
 upck8 = lambda x: struct.unpack('B', x)[0]
 upck16 = lambda x: struct.unpack('H', x)[0]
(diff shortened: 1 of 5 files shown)


commial:
What about using the `options` facility ?

serpilliere:
The goal is only to give the possibility to change it for example by subclassing the object?
I don't like very must the immutable hard coded value

commial:
I agree with you, it could be very dependent of a project needs. I've updated the PR.

PR #101, opened 2015-03-06
Faster depgraph

commial:
Ridiculously increase `DepGraph` performance.

In some corner cases, the complexity go from factorial to polynomial. Indeed, irrelevant path were conserved and used during comparison. That is to say, to be equal, two depnodes must have visited the same basic blocks with same elements. It's a good thing when one of these element is relevant (ie. is a `modifier`), but not when none of them are relevant.
In addition, some pure data-structures improvement have been done.

An naive hash on `DepDict` has not been used because there are a lot of collisions. Indexing `DepDict` by `label` seems to be a more efficient patch.


serpilliere:
Indeed.

PR #104, opened 2015-03-06
IR: factor identical slices in multi dst effects

serpilliere:
`get_modified_slice` can send back multiple identical slices.
Those slices need to be factorised for next steps.

### DIFF
--- a/miasm2/ir/ir.py
+++ b/miasm2/ir/ir.py
@@ -146,7 +146,7 @@ class ir(object):
         """
         If multiple affection to a same ExprId are present in @affect_list,
         merge them (in place).
-        For instance, XCGH AX, AL semantic is
+        For instance, XCGH AH, AL semantic is
         [
             RAX = {RAX[0:8],0,8, RAX[0:8],8,16, RAX[16:64],16,64}
             RAX = {RAX[8:16],0,8, RAX[8:64],8,64}
@@ -173,21 +173,22 @@ class ir(object):
                 continue
 
             # Find collision
-            e_colision = reduce(
-                lambda x, y: x + y, [e.get_modified_slice() for e in expr_list])
-
+            e_colision = reduce(lambda x, y: x.union(y),
+                                (e.get_modified_slice() for e in expr_list),
+                                set())
             # Sort interval collision
-            known_intervals = sorted([(x[1], x[2]) for x in set(e_colision)])
+            known_intervals = sorted([(x[1], x[2]) for x in e_colision])
 
             # Fill with missing data
-            missing_i = get_missing_interval(known_intervals, 0, e.src.size)
+            missing_i = get_missing_interval(known_intervals, 0, dst.size)
 
-            rest = [(m2_expr.ExprSlice(dst, r[0], r[1]), r[0], r[1])
-                    for r in missing_i]
+            remaining = ((m2_expr.ExprSlice(dst, *interval),
+                          interval[0],
+                          interval[1])
+                         for interval in missing_i)
 
             # Build the merging expression
-            slices = e_colision + rest
-            slices.sort(key=lambda x: x[1])
+            slices = sorted(e_colision.union(remaining), key=lambda x: x[1])
             final_dst = m2_expr.ExprCompose(slices)
 
             # Remove unused expression


miasm2/ir/ir.py
```diff
@@ -176,6 +176,9 @@ def merge_multi_affect(self, affect_list):
             e_colision = reduce(
                 lambda x, y: x + y, [e.get_modified_slice() for e in expr_list])
 
+            # Factorize identic collitions
+            e_colision = list(set(e_colision))
+
             # Sort interval collision
             known_intervals = sorted([(x[1], x[2]) for x in set(e_colision)])
```
p-l-: `set()` is useless now here.

miasm2/ir/ir.py
```diff
@@ -176,6 +176,9 @@ def merge_multi_affect(self, affect_list):
             e_colision = reduce(
                 lambda x, y: x + y, [e.get_modified_slice() for e in expr_list])
 
+            # Factorize identic collitions
+            e_colision = list(set(e_colision))
```
p-l-: Can't you keep a `set()` instead of creating another list?

miasm2/ir/ir.py
```diff
@@ -176,6 +176,9 @@ def merge_multi_affect(self, affect_list):
             e_colision = reduce(
                 lambda x, y: x + y, [e.get_modified_slice() for e in expr_list])
```
p-l-: This creates two lists (one as the output of `reduce()`, one by list comprehension.

What do you think of replacing lines 176 to 180 by something like:

```python
e_colision = reduce(lambda x, y: set(x).union(y), (e.get_modified_slice() for e in expr_list))
```

If you really need a list use `e_colision = list(reduce([...]))`. And anyway drop the `set()` in `known_intervals = [...]`

serpilliere:
I have updated the code.
By the way, there was a mistake in the code:
```python
missing_i = get_missing_interval(known_intervals, 0, e.src.size)
```

`e` was referencing a variable generated in a comprehension list.

miasm2/ir/ir.py
```diff
@@ -173,21 +173,20 @@ def merge_multi_affect(self, affect_list):
                 continue
 
             # Find collision
-            e_colision = reduce(
-                lambda x, y: x + y, [e.get_modified_slice() for e in expr_list])
-
+            e_colision = reduce(lambda x, y: set(x).union(y),
+                                (e.get_modified_slice() for e in expr_list))
```
p-l-: By bad, it's better written as this:

```Python
e_colision = reduce(lambda x, y: x.union(y), (e.get_modified_slice() for e in expr_list), set())
```

This is a tiny fix but I think it's better as it avoids to call the `set()` constructor on a `set` object for each element from the generator. Sorry I didn't write it like that in the first place.

The rest of the patch looks good to me.

serpilliere:
Modification done.

miasm2/ir/ir.py
```diff
@@ -173,21 +173,21 @@ def merge_multi_affect(self, affect_list):
...
-            known_intervals = sorted([(x[1], x[2]) for x in set(e_colision)])
+            known_intervals = sorted([(x[1], x[2]) for x in e_colision])
 
             # Fill with missing data
-            missing_i = get_missing_interval(known_intervals, 0, e.src.size)
+            missing_i = get_missing_interval(known_intervals, 0, dst.size)
 
-            rest = [(m2_expr.ExprSlice(dst, r[0], r[1]), r[0], r[1])
-                    for r in missing_i]
+            remaining = [(m2_expr.ExprSlice(dst, *interval),
```
p-l-: Unless I'm wrong you do not use `remaining` after line 199 where a generator would be enough and use less memory. Just replace `[...]` by `(...)`:

```Python
remaining = ((m2_expr.ExprSlice(dst, *interval), interval[0], interval[1]) for interval in missing_i)

**Comment by serpilliere** [comment]
done.

File: miasm2/ir/ir.py [hunk 6]
```diff
@@ -173,21 +173,22 @@ def merge_multi_affect(self, affect_list):
...
-            known_intervals = sorted([(x[1], x[2]) for x in set(e_colision)])
+            known_intervals = sorted([(x[1], x[2]) for x in e_colision])
 
             # Fill with missing data
-            missing_i = get_missing_interval(known_intervals, 0, e.src.size)
+            missing_i = get_missing_interval(known_intervals, 0, dst.size)
 
-            rest = [(m2_expr.ExprSlice(dst, r[0], r[1]), r[0], r[1])
-                    for r in missing_i]
+            remaining = remaining = ((m2_expr.ExprSlice(dst, *interval),
```
p-l-: I know you're going to hate me but... you have one `remaining = ` more than needed here. Sorry mate, I owe you a good beer for that PR.

serpilliere:
done.

PR #106, opened 2015-03-10
Dominator

serpilliere:
Dominator algorithm can only work on graph with one head.
Add PostDominator algorithm

### DIFF
--- a/miasm2/core/graph.py
+++ b/miasm2/core/graph.py
@@ -216,3 +216,45 @@ shape = "box"
             for succ in self.successors_iter(node):
                 todo.add(succ)
         return dict(dominators)
+
+
+    def compute_postdominators(self, leaf):
+        """Compute postdominators of the graph"""
+        nodes = set(self.reachable_parents(leaf))
+        postdominators = defaultdict(set)
+        for node in nodes:
+            postdominators[node].update(nodes)
+
+        postdominators[leaf] = set([leaf])
+        modified = True
+        todo = set(nodes)
+
+        while todo:
+            node = todo.pop()
+
+            # Leaf state must not be changed
+            if node == leaf:
+                continue
+
+            # Compute intersection of all successors'postdominators
+            new_post_dom = None
+            for succ in self.successors_iter(node):
+                if not succ in nodes:
+                    continue
+                if new_post_dom is None:
+                    new_post_dom = set(postdominators[succ])
+                new_post_dom.intersection_update(postdominators[succ])
+
+            # We are not a leaf to we have at least one post dominator
+            assert(new_post_dom is not None)
+
+            new_post_dom.update(set([node]))
+
+            # If intersection has changed, add sons to the todo list
+            if new_post_dom == postdominators[node]:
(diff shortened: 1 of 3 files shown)


miasm2/core/graph.py
```diff
@@ -162,42 +148,113 @@ def dot(self):
...
+                continue
+            reachable.add(node)
+            yield node
+            for pred in self.predecessors_iter(node):
+                todo.add(pred)
+
+    def compute_dominators(self, head):
+        """Compute dominators of the graph"""
+        nodes = set(self.reachable_sons(head))
         dominators = defaultdict(set)
```
commial: Where did you use the fact that dominators is a `defaultdict`?
You can use `set` for copy in line 184, and avoid using again a `dict()` at the end of the function. In addition, `defaultdict` can easily be prone to error.

miasm2/core/graph.py
```diff
@@ -162,42 +148,113 @@ def dot(self):
         return out
 
 
+    def reachable_sons(self, head):
```
commial: To factorize and avoid future copy/paste error on patches, you could use a protected method with a callback (`successors_iter` for `reachable_sons` and `predecessors_iter` for `reachable_parents`)

miasm2/core/graph.py
```diff
@@ -162,42 +148,113 @@ def dot(self):
...
                 continue
+
             dominators[node] = new_dom
-            for succ in self.successors(node):
+            for succ in self.successors_iter(node):
                 todo.add(succ)
         return dict(dominators)
+
+
+    def compute_postdominators(self, leaf):
```
commial: Same remark as `reacheable_*`. As dominator and post-dominator algorithms are completely symmetric, you can use two callbacks for the only changes: `reachable_*` and `*_iter`.

serpilliere:
Correct.
The default dict is removed.
The code is factorized for dominators/postdominators.

commial:
Thanks for your modifications. Did you intend to apply code factorizing on `reacheable_*` too ?

serpilliere:
Correct.
Code factorized.

miasm2/core/graph.py
```diff
@@ -162,42 +148,89 @@ def dot(self):
...
+            if node in reachable:
+                continue
+            reachable.add(node)
+            yield node
+            for next_node in next_cb(node):
+                todo.add(next_node)
+
+    def reachable_sons(self, head):
+        """Compute every nodes reachable from node @head"""
+        for node in self._reachable_nodes(head, self.successors_iter):
```
commial: `self._reachable_nodes(head, self.successors_iter)` is already an iterator. You can directly write:
`return self._reachable_nodes(head, self.successors_iter)`.

For instance:
```Python
def test():
    for i in xrange(4):
        yield i
def foobar():
    return test()
for i in foobar():
    print i
# 0
# 1
# 2
# 3
```

Idem for `reachable_parents` :smiley:

serpilliere:
Modified.

PR #111, opened 2015-03-12
Make miasm compile on MacOS X.

jmichelp:
Endianness macros are different on Mac. Define them for compatibility.


jmichelp:
Looks good to me.
I tested it on Yosemite and it seems to work, at least for what I was trying + the examples in the wiki. But running test_all.py fails on some tests (and unpack_pack it taking forever).

Issue #27, opened 2014-12-30
Is miasm compatible with OSX ?


dummys:
Hello,

I try to build miasm on my osx. After installing all the depedencies I type python setup.py build and got this:

building
build with 'all'
running build
running build_py
running build_ext
building 'miasm2.jitter.arch.JitCore_x86' extension
clang -fno-strict-aliasing -fno-common -dynamic -I/usr/local/include -I/usr/local/opt/sqlite/include -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/include/python2.7 -c miasm2/jitter/vm_mngr.c -o build/temp.macosx-10.9-x86_64-2.7/miasm2/jitter/vm_mngr.o
miasm2/jitter/vm_mngr.c:230:8: warning: explicitly assigning a variable of type 'uint64_t'
      (aka 'unsigned long long') to itself [-Wself-assign]
                        ret = ret;
                        ~~~ ^ ~~~
miasm2/jitter/vm_mngr.c:306:8: warning: explicitly assigning a variable of type 'uint64_t'
      (aka 'unsigned long long') to itself [-Wself-assign]
                        src = src;
                        ~~~ ^ ~~~
2 warnings generated.
clang -fno-strict-aliasing -fno-common -dynamic -I/usr/local/include -I/usr/local/opt/sqlite/include -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/include/python2.7 -c miasm2/jitter/vm_mngr_py.c -o build/temp.macosx-10.9-x86_64-2.7/miasm2/jitter/vm_mngr_py.o
miasm2/jitter/vm_mngr_py.c:711:24: error: use of undeclared identifier '__BIG_ENDIAN'
        self->vm_mngr.sex   = __BIG_ENDIAN;
                              ^
miasm2/jitter/vm_mngr_py.c:719:24: error: use of undeclared identifier '__LITTLE_ENDIAN'
        self->vm_mngr.sex   = __LITTLE_ENDIAN;
                              ^
2 errors generated.
SystemExit("error: command 'clang' failed with exit status 1",)
build with 'notcc'
running build
running build_py
running build_ext
building 'miasm2.jitter.arch.JitCore_x86' extension
[...]

PR #112, opened 2015-03-12
Add couple of macros for OS X

jmichelp:
Those macros where required for miasm to compile on Yosemite using
Apple XCode LLVM 6.0.

### DIFF
--- a/miasm2/jitter/vm_mngr.h
+++ b/miasm2/jitter/vm_mngr.h
@@ -20,6 +20,8 @@
 
 #ifdef __APPLE__
 #define __BYTE_ORDER __BYTE_ORDER__
+#define __BIG_ENDIAN __DARWIN_BIG_ENDIAN
+#define __LITTLE_ENDIAN __DARWIN_LITTLE_ENDIAN
 #elif defined(__NetBSD__)
 #define __BYTE_ORDER _BYTE_ORDER
 #define __BIG_ENDIAN _BIG_ENDIAN


commial:
It seems to be related to #27. Does it fully work for you with this patch ?

jmichelp:
I spent quite some time to get tcc to compile on os x (a few CFLAGS/LDFLAGS were missing or needed slightly modification) and then link with miasm. Unfortunately the jitter fails when trying the jitter example from the README.md (compilation error, etc.)

With LLVM jitter, the example from README.md fails with KeyError: 'IRDst_new' and the test suite is not able to detect I have python llvm bindings so it simply skips the tests.

I was going to have a look at the LLVM jitter to make it work.

commial:
If you have a way to have TCC working on OS X, can you share it here or on the aforementionned issue ? It will help us and others OS X users.

Regarding LLVM support, as specified in #5 , it is currently broken cause we want to change our LLVM binding provider. That way, we have specifically deactivate LLVM tests in `test_all`:
```Python
# TODO XXX: fix llvm jitter (deactivated for the moment)
llvm = False
```

jmichelp:
I said I was able to compile both TCC and miasm (i.e. without linking errors). Three modifications were needed in Makefile:
* Replace "-Wl,-soname,$@" by "-Wl,-install_name,$@"
* After "ifeq ($(TARGETOS),Darwin)" add the following line "CFLAGS+=-arch x86_64 -arch i386"
* After "ifeq ($(TARGETOS),Darwin)" add the following line "LDFLAGS+=-arch x86_64 -arch i386"

This will build the multiarch Mach-O library that python expects.

Unfortunately tinycc still doesn't handle Mach-O files and adding a new file format looks quite a decent amount of work. That's why I was considering using LLVM Jitter instead.

serpilliere:
As the PR is linked the issue #27 and doesn't bring more clues, I'm closing it.

guedou:
I think that this patch should be merged. It is not perfect, yet it allows to compile and use the Python jitter.
I tested it with the example from the README, and it worked fine.

commial:
Good news! Did you test with test_all.py?

guedou:
After editing `test_all.py` to only activate `jitter_engines = ["python"]`, I got:
```
Result: 174/176 pass
ERROR x86/unit/mn_float.py python
[..]
  File "/tmp/miasm/build/lib.macosx-10.9-x86_64-2.7/miasm2/jitter/emulatedsymbexec.py", line 44, in _func_write
    raise RuntimeError("A simplification is missing: %s" % data)
RuntimeError: A simplification is missing: (double_to_int_32 ((int_32_to_double 0x1) fadd (int_32_to_double 0x1)))

ERROR x86/unit/mn_int.py python
[..]
  File "/private/tmp/miasm/test/arch/x86/unit/asm_test.py", line 45, in __call__
    self.check()
  File "x86/unit/mn_int.py", line 27, in check
    assert self.int_num == 0x42
AssertionError
```

I can provide the full traceback if needed.

Concerning the "gcc" jitter (gcc is aliased to clang on OS X), I got a bunch of the following errors, that I don't really understand the meaning:
```
/var/folders/2d/dykg9rb91ll53skxl7jkjq340000gn/T/tmpxcMsv3.c:50:5: error: indirect goto in function with no address-of-label expressions
    goto *local_labels[BlockDst->address];
    ^
1 error generated.
```

commial:
Thanks to #408 and #421, clang is now supported (`gcc` jitter uses now `cc`, aliased to `clang` on Mac). We're getting closer!

It seems that some issues regarding bundle / dylib remains (maybe related to the `-bundle` option used by  `distutils`).

As travis now support OS X builds, once the support work, it could be great to add a new target (test matrix) in the `.travis.yml`.

PR #123, opened 2015-03-21
Depgraph filter

serpilliere:
Support callback to filter depnode to follow.
Update IDA/Depgraph to support `NoCallNoMem` depgraph.

### DIFF
--- a/miasm2/analysis/depgraph.py
+++ b/miasm2/analysis/depgraph.py
@@ -1,5 +1,6 @@
 """Provide dependency graph"""
 import itertools
+from collections import namedtuple
 
 import miasm2.expression.expression as m2_expr
 from miasm2.core.graph import DiGraph
@@ -7,7 +8,6 @@ from miasm2.core.asmbloc import asm_label
 from miasm2.expression.simplifications import expr_simp
 from miasm2.ir.symbexec import symbexec
 from miasm2.ir.ir import irbloc
-
 
 class DependencyNode(object):
     """Node elements of a DependencyGraph
@@ -374,6 +374,8 @@ class DependencyResult(object):
                 for depnode in self.input}
 
 
+FollowExpr = namedtuple("FollowExpr", ["follow", "element"])
+
 class DependencyGraph(object):
     """Implementation of a dependency graph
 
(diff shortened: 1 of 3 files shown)


miasm2/analysis/depgraph.py
```diff
@@ -374,6 +374,37 @@ def emul(self, step=False):
...
 
+def follow_simp_expr(exprs):
+    """Avoid tracking useless elements, as XOR EAX, EAX"""
+
+    follow = set()
+    for expr in exprs:
+        follow.add(expr_simp(expr))
+    return follow, set()
+
+def follow_mem_cst(exprs):
```
commial: As for `follow_simp_expr`, you could add docstring for these functions

miasm2/analysis/depgraph.py
```diff
@@ -392,6 +423,20 @@ def __init__(self, ira):
 
         # The IRA graph must be computed
         self._ira.gen_graph()
+        self.cb_follow = [follow_simp_expr, follow_mem_cst]
```
commial: As `cb_follow` seems to be class dependent and not instance dependent, you could use a class attribute instead

miasm2/analysis/depgraph.py
```diff
@@ -374,6 +374,37 @@ def emul(self, step=False):
...
+        follow.add(expr_simp(expr))
+    return follow, set()
+
+def follow_mem_cst(exprs):
+    follow = set()
+    for expr in exprs:
+        follow.update(expr.get_r(mem_read=True, cst_read=True))
+    return follow, set()
+
+def follow_nomem_cst(exprs):
```
commial: `follow_nomem_cst` implementation is very close to `follow_mem_cst`. You could define a function which just take in argument `mem_read` and `cst_read` values and merge them.

miasm2/analysis/depgraph.py
```diff
@@ -417,27 +462,27 @@ def _resolve_depNode(self, depnode):
...
-
-                    read.update(src.get_r(mem_read=True, cst_read=True))
+                    elements = self.follow_apply_cb(affect.src)
+                    read.update(elements)
                     modifier = True
 
             ## If it's not a modifier affblock, reinject current element
             if not modifier:
-                read = set([depnode.element])
+                read = set([(True, depnode.element)])
```
commial: To avoid confusion, you could use a `NamedTuple` here, for instance with `follow` and `element` attribute.
The code snippets which are using this structure will be less prone to error and more intelligible.

miasm2/analysis/depgraph.py
```diff
@@ -605,15 +650,12 @@ class DependencyGraph_NoMemory(DependencyGraph):
...
-        # Get inital elements
-        result = super(DependencyGraph_NoMemory, self)._resolve_depNode(depnode)
-
-        # If @depnode depends on a memory element, give up
-        for node in result:
-            if isinstance(node.element, m2_expr.ExprMem):
-                return set()
-
-        return result
+class DependencyGraph_NoCall_NoMemory(DependencyGraph):
```
commial: To avoid defining `n!` class for each callbacks possibility, a metaclass could be a solution. That is to say, a class generator which take in input the callback to use, or with more abstraction, the element to handle (`follow_call`, `follow_mem`, ...).

miasm2/analysis/depgraph.py
```diff
@@ -417,27 +462,27 @@ def _resolve_depNode(self, depnode):
...
-                                                modifier=modifier))
+            for follow, element in read:
+                dependencies.add((follow,
+                                  DependencyNode(depnode.label,
+                                                 element,
+                                                 depnode.line_nb - 1,
+                                                 modifier=modifier)))
             output = dependencies
 
         return output
```
commial: Can you update the method `docstring` to specify the output particularity ?

miasm2/analysis/depgraph.py
```diff
@@ -392,6 +423,20 @@ def __init__(self, ira):
 
         # The IRA graph must be computed
         self._ira.gen_graph()
+        self.cb_follow = [follow_simp_expr, follow_mem_cst]
+
+    def follow_apply_cb(self, expr):
```
commial: If this function is public, can you add a `docstring` here ?

Otherwise, you could declare it as protected. In my mind, there is no reason to call this function from outside.

PR #126, opened 2015-03-23
Cache translators

commial:
With this PR, translators are now used with instance.

That way, a cache (here bounded thanks to `BoundedDict`) can be used on expression and shared accros translators.
In addition, some translators (ie. `z3`) need to share attributes during recursion calls (`Z3_mem`). This way is less hacky than the previous one.

Thanks to @fmonjalet for the idea.

### DIFF
--- a/miasm2/ir/ir2C.py
+++ b/miasm2/ir/ir2C.py
@@ -1,7 +1,7 @@
 import miasm2.expression.expression as m2_expr
 from miasm2.expression.simplifications import expr_simp
 from miasm2.core import asmbloc
-from miasm2.ir.translators.C import TranslatorC
+from miasm2.ir.translators import Translator
 import logging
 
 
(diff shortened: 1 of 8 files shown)


example/ida/utils.py
```diff
@@ -132,7 +132,7 @@ def __init__(self, expr):
         self.expr = expr
 
         # Initial translation
-        text = Translator.to_language(self.languages[0]).from_expr(self.expr)
+        text = Translator.to_language(self.languages[0])().from_expr(self.expr)
```
fmonjalet: I'm not quite sure of how to handle that, but to_language(...) might return an already instanciated Translator, to avoid the weird double-() syntax. One of the good things of the factory pattern is to have no idea of how the object is instanciated. But the current syntax allows to pass arguments to the translator instance easily.

Maybe something like `Translator.to_language(cls, target_lang, *args, **kwargs)` could do the job? Not quite sure.

commial: Yep, I agree with you. The use of the instance seems to be the 99% case, not the class use.

I'll update the PR ASAP accordingly (and thus keep compatibility with the old API).

serpilliere:
Can you rebase the PR?
We unfortunately fixed some bugs in this code before merging the PR :smile:

commial:
Done !

PR #131, opened 2015-03-24
added semantics for x86 pxor

mrphrazer:
pxor doesn't change any flags, so I'm not sure if this is needed:

```
e += update_flag_logic(c)
e.append(m2_expr.ExprAff(a, c))
```

### DIFF
--- a/miasm2/arch/x86/sem.py
+++ b/miasm2/arch/x86/sem.py
@@ -302,6 +302,13 @@ def xor(ir, instr, a, b):
     return e, []
 
 
+def pxor(ir, instr, a, b):
+    e = []
+    c = a ^ b
+    # e += update_flag_logic(c)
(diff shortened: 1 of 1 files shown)


mrphrazer:
Thanks for the fast response. Done

serpilliere:
Ok. I just spot you didn't add the PXOR into the mnemonic list (in `sem.py`):
```python
mnemo_func = {'mov': mov,
              'xchg': xchg,
              'movzx': movzx,
              'movsx': movsx,
...
              'fst': fst,
              'fstp': fstp,
              'fist': fist,
              'fistp': fistp,
```


Can you add an entry like:
```python
              'pxor': pxor,
```

Everything may be ok after this fix :smile:

mrphrazer:
eh, please look again, I did that in line 3148 :)

PR #134, opened 2015-03-31
Sandbox Win: add PE object to win_api_x86_32.winobjs

p-l-:
This is needed at least to "implement" `GetModuleHandle`.


miasm2/analysis/sandbox.py
```diff
@@ -183,6 +183,7 @@ def __init__(self, custom_methods, *args, **kwargs):
         # Load main pe
         with open(self.fname) as fstream:
             self.pe = vm_load_pe(self.jitter.vm, fstream.read())
+            win_api_x86_32.winobjs.current_pe = self.pe
```
commial: Can it be done out of the scope of the `with` statement ?
This statement is only required for the input read.

PR #139, opened 2015-04-02
added z3 translation for call_func_ret

mrphrazer:
This allows reasoning about return values of function calls within a function.

For instance, foo(a) will return 1 if bar(a) returns 1335.
```
int foo(int a)
{
    if (bar(a) + 2 == 1337)
        return 1;
    return 0;
}
```

### DIFF
--- a/miasm2/ir/translators/z3_ir.py
+++ b/miasm2/ir/translators/z3_ir.py
@@ -170,6 +170,8 @@ class TranslatorZ3(Translator):
                     res = res << arg
                 elif expr.op == "idiv":
                     res = res / arg
+                elif expr.op == "call_func_ret":
+                    res = arg
                 else:
                     raise NotImplementedError("Unsupported OP yet: %s" % expr.op)
         elif expr.op == 'parity':


serpilliere:
Hi

If I understand correctly your code, the expression (for example):
```
EAX = ExprOp('call_func_ret', ExprId('bar'), ESP, EBX, ECX)
```
will return in every cases:
```
EAX_z3 = ECX_z3
```
For instance, it will be wrong on
```C
int bar(int a)
{
    return 1335;
}
int foo()
{
    a = 2;
    if (bar(a) + 2 == 1337)
        return 1;
    return 0;
}
```
In this case, z3 will say it is unsat (every time `2 + 2 != 1337`) and this example is clearly always satisfiable. So the solution here is to model correctly your sub function. It can be done by subclassing `ira` and replace function `call_effects`.

Another way is to replace the `ExprOp` 'call_func_ret' by a new ExprId 'bar_ret'. That way you will obtain your equation:
```
bar_ret + 2 == 1337
```

mrphrazer:
Hi,

thanks, you are right.  Your second way is the way I wanted to go.

fmonjalet:
Hi,

Translating ``bar(a)`` to ``bar_ret`` might not be the best way to go: only constant functions will be handled correctly. For example:

```c
int minus_one(int a) {
    return a-1;
}

int foo(int a) {
    int b = minus_one(a);
    int c = minus_one(b);
    if (b + c == 4)
        return 1;
    return 0;
}
```

In this example (maybe not the simplest...) ```b + c == 4``` will be found satisfyable (```minus_one_ret == 2```) while it is not (```b + c = a - 1 + a - 2 = 2a - 3``` is necessarily odd).

One interesting solution might be to take a look at the z3.Function class (http://research.microsoft.com/en-us/um/redmond/projects/z3/z3.html#FuncDeclRef). I never used it, I think it behaves a bit like a multidimensionnal array. I don't know if it will work, though. :)
In any case, if you use an identifier, I would say it has to be unique for every call.

For an accurate satisfyability answer, the best way stays Serpi's one : to model the function. For "permissive" satifyability, z3.Function might be interesting.

Good luck!

PR #145, opened 2015-04-13
make asm_label comparable

mrphrazer:


mrphrazer:
just saw that the code influences other parts in such a way that they won't work then.

serpilliere:
Hi

In fact this was true some commit ago, but I removed it recently in order to sanitize `asmbloc.py`. We tend to force user to use only `symbol_pool.add_label` api instead of `asm_label` directly. Thus an `asm_label`is always linked to a symbol pool and is unique.

PR #147, opened 2015-04-22
Printable sem

commial:
Add a helper to make instruction semantic writing easier and more readable.

For instance:
```Python
@sb.parse
def mov(a, b):
    a = b
    mem32[a] = mem32[b]
```
Will be translated to:
```Python
def mov(ir, instr, a, b):
    return [ExprAff(a, b), ExprAff(ExprMem(a, 32), ExprMem(b, 32))], []
```

A part of the *Mips32* semantic has been rewritten accordingly.

### DIFF
--- a/miasm2/core/sembuilder.py
+++ b/miasm2/core/sembuilder.py
@@ -90,6 +90,7 @@ class SemBuilder(object):
         # Get the function AST
         parsed = ast.parse(inspect.getsource(func))
         fc_ast = parsed.body[0]
+        argument_names = [name.id for name in fc_ast.args.args]
         body = []
 
         # AffBlock of the current instruction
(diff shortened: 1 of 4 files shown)


miasm2/core/sembuilder.py
```diff
@@ -0,0 +1,155 @@
...
+    def visit_Call(self, node):
+        """iX(Y) -> ExprIntX(Y)"""
+        # Match the function name
+        if not isinstance(node.func, ast.Name):
+            # TODO: launch visitor on node
+            return node
+
+        fc_name = node.func.id
+        new_name = fc_name
+        integer = self.parse_integer.findall(fc_name)
```
p-l-: Why do you use `.findall()` here? Since you only seem to need the first result, you may use `.search()` instead (not tested ;-)):

```python
    integer = self.parse_integer.search(fc_name)
    if integer is not None:
        new_name = "ExprInt%s" % integer.group()
```

miasm2/core/sembuilder.py
```diff
@@ -0,0 +1,155 @@
...
+        return node
+
+    def visit_Subscript(self, node):
+        """memX[Y] -> ExprMem(Y, X)"""
+
+        # Detect the syntax
+        if not isinstance(node.value, ast.Name):
+            return node
+        name = node.value.id
+        mem = self.parse_mem.findall(name)
```
p-l-: Same here than for line 31.

miasm2/core/sembuilder.py
commial: Nice catch !  It seems to be a debug residue :smile: 
I use `groups` instead of `group` because I only want the integer.

The PR has been updated accordingly.

PR #148, opened 2015-04-22
Depgraph Implicit

commial:
This PR introduces the `implicit` option in depgraph.
That is to say, giving the following code snippet:
```C
if zf
  a = b + 2;
else
  a = b + 3;
```

The dependencies are view as:
* Explicit: `b`, `2`, `3`
* Implicit: `zf`

When the option is used, both dependency types are tracked.

The IDA example `depgraph.py` has been updated.
For instance, on the `simple_test.bin` sample, two solutions are founded for `0x1007`, depending on the path.
exp1
exp2

In addition, a new example `symbol_exec/depgraph.py` has been added.
This example illustrates the depgraph use. For instance:
```
$ python symbol_exec/depgraph.py samples/simple_test.bin -m x86_32 0x0 0x8B eax 
Solution 0: EAX: 0x1006 -> sol_0.dot
Solution 1: EAX: 0x1003 -> sol_1.dot
Solution 2: EAX: 0x1008 -> sol_2.dot
Solution 3: EAX: 0x1005 -> sol_3.dot
Solution 4: EAX: 0x1004 -> sol_4.dot
Solution 5: EAX: 0x1002 -> sol_5.dot
Solution 6: EAX: 0x1007 -> sol_6.dot
Solution 7: EAX: 0x1001 -> sol_7.dot
```


commial:
You're right. PR updated.

PR #150, opened 2015-04-24
Fix uint

serpilliere:
`ExprInt` uses a `modint`as argument.
This modint can take an int or a long, which change its representation so:
Exprint(0x32) != ExprInt(0x32L)
This fix ensure `modint` argument is always a long

The bug is trigger by the actual regression tests on some 32 bit systems.
Add a dedicated regression tess

### DIFF
--- a/miasm2/expression/modint.py
+++ b/miasm2/expression/modint.py
@@ -4,9 +4,7 @@
 class moduint(object):
 
     def __init__(self, arg):
-        if isinstance(arg, moduint):
-            arg = arg.arg
-        self.arg = arg % self.__class__.limit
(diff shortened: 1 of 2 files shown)


miasm2/expression/modint.py
```diff
@@ -6,7 +6,7 @@ class moduint(object):
     def __init__(self, arg):
         if isinstance(arg, moduint):
```
commial: As `long()` is always call on arg, the `if` statement in no more needed

serpilliere:
This is wrong.
Convince yourself using a breakpoint in the `if`

PR #151, opened 2015-04-24
Jitter: fix cast for 32 bits machines

serpilliere:
On 32 bit systems, the high bytes of `size` are not set to 0 to generate random argument to be send to `automod_cb`, so generate a random behaviour in regression tests.

Fix every architectures.


commial:
Why don't you change the signature instead ? (`void check_automod(JitCpu* jitcpu, uint64_t addr, uint64_t size)`
Or `"LI"` in `PyObject_CallMethod` ?

serpilliere:
fixed

PR #154, opened 2015-04-28
IR: Improve dead code elimination

carolineLe:
Dead code analysis computed using Ken Kennedy Algorithm

Ref: A survey of data flow analysis techniques, IBM Thomas
J. Watson Research Division, 1979

### DIFF
--- a/miasm2/expression/expression.py
+++ b/miasm2/expression/expression.py
@@ -152,6 +152,11 @@ class Expr(object):
     def get_w(self):
         return self.arg.get_w()
 
+    def is_function_call(self):
+        """Returns true if the considered Expr is a function call
+        """
+        return False
+
     def __repr__(self):
         if self._repr is None:
             self._repr = self._exprrepr()
@@ -511,6 +516,9 @@ class ExprAff(Expr):
             return set([self._dst])  # [memreg]
         else:
             return self._dst.get_w()
+
+    def is_function_call(self):
+        return isinstance(self.src, ExprOp) and self.src.op.startswith('call')
 
     def _exprhash(self):
         return hash((EXPRAFF, hash(self._dst), hash(self._src)))
--- a/miasm2/ir/analysis.py
+++ b/miasm2/ir/analysis.py
@@ -6,7 +6,7 @@ import logging
 from miasm2.ir.symbexec import symbexec
 from miasm2.core.graph import DiGraph
 from miasm2.expression.expression \
-    import ExprAff, ExprCond, ExprId, ExprInt, ExprMem, ExprOp
+    import ExprAff, ExprCond, ExprId, ExprInt, ExprMem
 
 log = logging.getLogger("analysis")
 console_handler = logging.StreamHandler()
@@ -15,6 +15,10 @@ log.addHandler(console_handler)
 log.setLevel(logging.WARNING)
 
 class ira:
+
+    def ira_regs_ids(self):
+        """Returns ids of all registers used in the IR"""
+        return self.arch.regs.all_regs_ids + [self.IRDst]
 
     def sort_dst(self, todo, done):
         out = set()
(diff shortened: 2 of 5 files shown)


miasm2/expression/expression.py
```diff
@@ -151,7 +151,10 @@ def get_r(self, mem_read=False, cst_read=False):
 
     def get_w(self):
         return self.arg.get_w()
-
+    
+    def is_function_call(self):
```
commial: This code snippet seems to be specific to `ExprAff` and should not be implemented in `Expr` (`.src` undefined).
If this code can be called on any `Expr`, you could rewrite it as follow:
```Python
class Expr():
    ...
    def is_function_call(self):
        "Some Docstring"
        return False
    ...
...
class ExprAff(Expr):
    ...
    def is_function_call(self):
        return isinstance(self.src, ExprOp) and self.src.op.startswith('call')
...
```

miasm2/ir/analysis.py
```diff
@@ -115,194 +119,246 @@ def graph(self):
...
+            elem = worklist.pop()
+            useful.add(elem)
+            irb, irs_ind, ins = elem
+
+            bloc = self.blocs[irb]
+            instr_defout = bloc.defout[irs_ind]
+            cur_kill = bloc.cur_kill[irs_ind]
+            cur_reach = bloc.cur_reach[irs_ind]
+
+            ir = bloc.irs[irs_ind]
```
p-l-: It seems you do not use `ir` after that.

miasm2/ir/analysis.py
```diff
@@ -115,194 +119,246 @@ def graph(self):
...
-                    pass
-                elif (isinstance(i_cur.src, ExprOp) and
-                    i_cur.src.op.startswith('call')):
-                    # /!\ never remove ir calls
-                    pass
-                elif i_cur.dst not in c_out:
-                    del(ir[j])
+                cur_instr = ir[j]
+                if (isinstance(cur_instr.dst, ExprId)
+                    and not (irb.label, k, cur_instr) in useful):
```
commial: Please use `... not in ...` instead of `not ... in ...` (more homogenous with Miasm's code)

miasm2/ir/analysis.py
```diff
@@ -115,194 +119,246 @@ def graph(self):
...
-                    continue
-                j += 1
+                else:
+                    j += 1
+        return modified
 
-        # print 'state2'
-        # self.dump_bloc_state(irb)
+    def init_useful_instr(self):
+        """Computes a set of triples (bloc, instruction number, instruction)
```
commial: Can you please use `block` instead of `bloc` (in the whole PR) ?

miasm2/ir/analysis.py
```diff
@@ -115,194 +119,246 @@ def graph(self):
...
+        for k, ir in enumerate(irb.irs):
             j = 0
             while j < len(ir):
-                i_cur = ir[j]
-                if not isinstance(i_cur.dst, ExprId):
-                    pass
-                elif i_cur.dst == self.IRDst:
-                    # never delete irdst
-                    pass
-                elif (isinstance(i_cur.src, ExprOp) and
```
p-l-: Since this removes the only use of `ExprOp`, you should no longer import it.

miasm2/ir/analysis.py
```diff
@@ -115,194 +119,246 @@ def graph(self):
...
+            # Affecting return registers
+            if not has_son:
+                for r in self.get_out_regs(bloc):
+                    useful.update(bloc.defout[-1][r]
+                                  if bloc.defout[-1][r] else
+                                  bloc.cur_reach[-1][r])
+
+        return useful
+
+    def mark_useful_code(self):
```
commial: If this method is only used in the internal dead code algorithm, please mark it as protected (`_mark_useful_code`).

miasm2/ir/analysis.py
```diff
@@ -115,194 +119,246 @@ def graph(self):
...
+                    irb.cur_reach[i - 1][c_reg].difference(
+                        irb.cur_kill[i - 1][c_reg]))
+                irb.cur_reach[i][c_reg].update(pred_through)
+                if irb.defout[i][c_reg]:
+                    # KILL(n) = DEFOUT(n) ? REACH(n)\DEFOUT(n) : EMPTY
+                    irb.cur_kill[i][c_reg].update(
+                        irb.cur_reach[i][c_reg].difference(
+                            irb.defout[i][c_reg]))
+
+    def test_kill_reach_fix(self):
```
commial: This method seems to be part of internal algorithm (-> protected)

test/ir/analysis.py
```diff
@@ -0,0 +1,680 @@
...
+    open("simp_graph_%02d.dot" % (i+1), "w").write(g_ira.graph())
+
+    # Same number of blocks
+    assert len(g_ira.blocs) == len(g_exp_ira.blocs)
+    # Check that each expr in the blocs are the same
+    for lbl , irb in g_ira.blocs.iteritems():
+        exp_irb = g_exp_ira.blocs[lbl]
+        assert len(irb.irs) == len(exp_irb.irs), "(%s)  %d / %d" %(
+            lbl, len(irb.irs), len(exp_irb.irs))
+        for i in xrange(0,len(exp_irb.irs)):
```
commial: Here, you redefined `i` (previously defined at l. 635)

test/test_all.py
```diff
@@ -117,6 +117,11 @@ def __init__(self, arch, container, address, *args, **kwargs):
                "symbexec.py",
                ]:
     testset += RegressionTest([script], base_dir="ir")
+testset += RegressionTest(["analysis.py"], base_dir="ir",
```
commial: To avoid using twice time the same generator, one can use:
```Python
sum((["simp_graph_%02d.dot" % test_nb, "graph_%02d.dot" % test_nb] for test_nb in xrange(1, 18)), [])
```

p-l-: You may want to avoid `sum()` (or `reduce()`) and prefer a double comprehension.

```Python
[fname for fnames in (["simp_graph_%02d.dot" % test_nb, "graph_%02d.dot" % test_nb]
       for test_nb in xrange(1, 18)) for fname in fnames]
```

PR #160, opened 2015-05-09
Extend digraph

mrphrazer:
graph.py: extend DiGraph with implementations for computation of immediate dominators and dominance frontier

### DIFF
--- a/miasm2/core/graph.py
+++ b/miasm2/core/graph.py
@@ -304,17 +304,12 @@ shape = "box"
                                             self.successors_iter)
 
     def compute_immediate_dominators(self, head):
-        """
-        Compute the immediate dominators of the graph
-
-        Depends on the implementation characteristics of reachable_parents.
-        Modifications could break this algorithm.
-        """
+        """Compute the immediate dominators of the graph"""
         dominators = self.compute_dominators(head)
         idoms = {}
 
         for node in dominators:
-            for predecessor in self.reachable_parents(node):
+            for predecessor in self.walk_dominators(node, dominators):
                 if predecessor in dominators[node] and node != predecessor:
                     idoms[node] = predecessor
                     break
--- a/miasm2/core/graph.py
+++ b/miasm2/core/graph.py
@@ -300,3 +300,15 @@ shape = "box"
         return self._walk_generic_dominator(node,
                                             postdominators,
                                             self.successors_iter)
+
+    def compute_immediate_dominators(self, head):
+        """Compute the immediate dominators of the graph"""
+        dominators = self.compute_dominators(head)
+        idoms = {}
+
+        for n in dominators.keys():
+            for p in self.reachable_parents(n):
+                if p in dominators[n] and n != p:
+                    idoms[n] = p
+                    break
+        return idoms
--- a/miasm2/core/graph.py
+++ b/miasm2/core/graph.py
@@ -338,9 +338,9 @@ shape = "box"
                     if runner not in idoms:
                         continue
                     while runner != idoms[node]:
-                        if node not in frontier:
-                            frontier[node] = set()
+                        if runner not in frontier:
+                            frontier[runner] = set()
 
-                        frontier[node].add(runner)
+                        frontier[runner].add(node)
                         runner = idoms[runner]
         return frontier
--- a/miasm2/core/graph.py
+++ b/miasm2/core/graph.py
@@ -306,10 +306,10 @@ shape = "box"
         dominators = self.compute_dominators(head)
         idoms = {}
 
-        for n in dominators:
-            for p in self.reachable_parents(n):
-                if p in dominators[n] and n != p:
-                    idoms[n] = p
+        for node in dominators:
+            for predecessor in self.reachable_parents(node):
+                if predecessor in dominators[node] and node != predecessor:
+                    idoms[node] = predecessor
                     break
         return idoms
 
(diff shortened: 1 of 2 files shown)


commial:
Thanks for your contribution. Can you add associated regression tests ?
You can add them in `test/core/graph.py`.

Regarding the commit message, thanks for taking in account my previous remark. You can omit the `.py`, this is more to indicated the part of Miasm, instead of the concerned file.

miasm2/core/graph.py
```diff
@@ -229,3 +229,38 @@ def compute_postdominators(self, leaf):
...
                                                 self.successors_iter,
                                                 self.predecessors_iter)
+
+    def compute_immediate_dominators(self, head):
+        """Compute the immediate dominators of the graph"""
+        dominators = self.compute_dominators(head)
+        idoms = {}
+
+        for n in dominators:
+            for p in self.reachable_parents(n):
```
fmonjalet: Since reachable_parents relies on set(), I am not sure it gives any guarantee on the order of the iteration on the reachable parents (i.e., the first yielded may not be the closest to the node). Therefore I think that in some cases, the first dominator-reachable-parent encountered will not be the closest to the node.
I may be unclear, but do you see what I mean?

mrphrazer:
Thanks for your feedback!

@commial: Sure, will do that.

@fmonjalet: Good point, I see what you mean. I'll take a look at that.

mrphrazer:
@commial: I've added the tests now

@fmonjalet: You are right about the set. However, in this case it works . reachable_parents builds all paths from the node to the root in an arbitrary order. There is no possibility to build a path to the root without including a dominator which is'nt the nearest dominator to the node.

fmonjalet:
Hi @mrphrazer, you are right: the way `reachable_parents` is implemented makes `compute_immediate_dominators` correct.
The fact that `reachable_parents` always yields a node after at least one of its children could be documented, since it is a strong (and implementation dependant) property, what do you think @commial ?

Thanks for the contribution by the way!

commial:
@fmonjalet I agree with you. It should be present in the `reachable_parents` since a modification could break this algorithm.
Regarding the time we've passed to check that it will actually work, a comment in the `compute_immediate_dominators`would be welcome.

The dominance frontier looks good for me, but please avoid using one/two characters variables in the future :smiley:.

mrphrazer:
@commial Why waiting for the future? :) Renamed the variables and added the comments.

mrphrazer:
I changed the filling order of the dict in compute_dominance_frontier. Now the mapping shows which block influences which set of blocks. That makes more sense than the other way round.

serpilliere:
Hi,

The dominator frontier is really interesting to me, and the implementation seems great!

I realized that I already needed something like this in a previous script, but a more generic approach: A function which returns an iterator of ordered dominators. What about generalizing this feature into something like:

```python
def walk_dominators(graph, node, dominators):
    """Return an iterator of the ordered list of @node's dominators
    The function doesn't return the self reference in dominators.
    @node: The start node
    @graph: the input graph
    @dominators: The dictionnary containing at least node's dominators
    """
    done = set()
    node_dominators = list(dominators[node])
    todo = set([node])
    while node_dominators:
        new_node = None
        while todo:
            node = todo.pop()
            if node in node_dominators:
                new_node = node
                break
            if node in done:
                continue
            done.add(node)
            for pred in graph.predecessors(node):
                todo.add(pred)
        assert(new_node is not None)
        yield new_node
        node_dominators.remove(new_node)
        todo = set([new_node])
```

So if we want the "direct" dominator, this correctponds to the first iterative element, and so on.

Do you think the feature is the same, or should it appear in another separated function?

mrphrazer:
Hi @serpilliere, thanks! :)

Well, your ```walk_dominators``` is also very useful, but it is somehow different than ```compute_immediate_dominators```:

1. Since every node dominates itself,  the second dominator yielded by  ```walk_dominators``` is the immediate dominator.
2.  There are cases besides the dominance_frontier, where you need a dictionary which contains all immediate dominators. For instance, if you need the dominator tree, it is enough to swap keys and values in that dict and you have the edges for the dominator tree.

So, I think both function fulfil different purposes and are both useful as they are and should be seperate funcitons.

serpilliere:
In fact, as described in the comment, the `walk_dominator` doesn't return the self reference, so the first yield element is the immediate dominator.

But your may be right, the patch may be in another commit.

mrphrazer:
I made my first statement because of the behaviour of the following code:
```python
from miasm2.core.graph import DiGraph
g = DiGraph()
g.add_edge(1, 2)
g.add_edge(1, 3)
g.add_edge(2, 4)
g.add_edge(2, 5)
g.add_edge(3, 5)
g.add_edge(4, 6)
g.add_edge(5, 6)

dominators = g.compute_dominators(1)
for dom in walk_dominators(g, 4, dominators):
    print dom
```
The output:
```
4
2
1
```

commial:
It seems that in your first iteration:

1. `node = todo.pop()` returns the input node
2. we take this branch `if node in node_dominators:` due to `compute_dominators` API
3. we finnish on `yield new_node`, yielding the input node, that is to say "self"

mrphrazer:
@commial  This is exactly what happens. 

A fix which doesn't yield self-reference would be replacing
```python
node_dominators = list(dominators[node])
```
with
```python
node_dominators = [n for n in dominators[node] if n != node]
```
This doesn't' require  changing the algorithm's logic, but isn't an elegant solution.

miasm2/core/graph.py
```diff
@@ -229,3 +231,45 @@ def compute_postdominators(self, leaf):
...
+        Compute the immediate dominators of the graph
+
+        Depends on the implementation characteristics of reachable_parents.
+        Modifications could break this algorithm.
+        """
+        dominators = self.compute_dominators(head)
+        idoms = {}
+
+        for node in dominators:
+            for predecessor in self.reachable_parents(node):
```
mrphrazer: Just an idea:

If  ```walk_dominators``` would be a function of DiGraph (mainly replacing the paratemeter ```graph``` with ```self```), then that line could be replaced with
```python
for predecessor in self.walk_dominators(node, dominators):
```
In that case, [the second property](https://github.com/cea-sec/miasm/pull/160#issuecomment-103938202)  would be fulfilled and the dependence on ```reachable_parents``` would be gone.

serpilliere:
You are right: I will push `walk_dominator` in DiGraph.

mrphrazer:
I changed the code to depend on ```walk_dominators```. Tested the code with  [that](https://github.com/cea-sec/miasm/pull/160#issuecomment-104904326) fix of ```walk_dominators``` and the committed graph tests.

commial:
Thanks to @serpilliere PR #173 , you can now rebase your pull request in order to use the proper `walk_dominators`.
Thanks again for your patience and your feedbacks.

mrphrazer:
Thank you @serpilliere  and @commial  for committing ```walk_dominators```. I rebased the pull request and extended the regression tests.

commial:
Thanks again for not giving up :smiley:

PR #168, opened 2015-05-22
Loader/Pe: fix vm2pe for 64 bit pe

serpilliere:


commial:
I know it is not fully relative to this commit, but at line 241 (after the commit), does `e_orig` need to be check ?

p-l-:
@commial this would need another PR in my opinion. The test `e_orig is not None` should be done only once:

```Python
if e_orig is not None:
    if min_addr is None:
        min_addr = min([e_orig.rva2virt(s.addr) for s in e_orig.SHList])
    if max_addr is None:
        max_addr = max([e_orig.rva2virt(s.addr + s.size) for s in e_orig.SHList])
    if img_base is None:
         img_base = e_orig.NThdr.ImageBase
```

serpilliere:
You are right, this has nothing to do with the commit: I only fixed 64 bit issues here.
Anyway, for your point, we search here to have custom variables which could be missing. If some of these values are missing, the code look into the original PE.

So we can either raise an exception or let the code spit an error if both are missing (the current case).
Maybe vm2pe needs a complete rewritting.

PR #169, opened 2015-05-22
Win_api: fix getprocaddr bad ptr tuncation

serpilliere:


miasm2/os_dep/win_api_x86_32.py
```diff
@@ -913,7 +913,7 @@ def kernel32_LoadLibraryExA(jitter):
 def kernel32_GetProcAddress(jitter):
     ret_ad, args = jitter.func_args_stdcall(["libbase", "fname"])
     # Ensure high-order WORD is 0
```
commial: It seems that you could remove the comment too.

serpilliere:
Here is the fix

PR #177, opened 2015-06-14
Extend ExprOp and Z3 translation with If-Then-Else

mrphrazer:
The idea behind this PR is to introduce a conditional expression in another context then ExprCond. The ExprOp ```ite``` intends to receive a condition and two expressions of the same size. The condition itself doesn't depend on the other expressions' size.

That allows it to make assignments as
```
x = ite(a + b = 7, y, z ) # x = (a + b == 7) ? y : z
```
or nested assignments as 
```
x  = (a == 1) ? b : ((a == 2) ? b : c)
```
and add them as contraints/implications to Z3.

### DIFF
--- a/miasm2/expression/expression.py
+++ b/miasm2/expression/expression.py
@@ -735,10 +735,19 @@ class ExprOp(Expr):
 
         if len(sizes) != 1:
             # Special cases : operande sizes can differ
-            if op not in ["segm"]:
+            if op not in ["segm", "ite"]:
                 raise ValueError(
                     "sanitycheck: ExprOp args must have same size! %s" %
                                  ([(str(arg), arg.size) for arg in args]))
+
+        if op in ["ite"]:
+            if len(args) != 3:
+                raise ValueError("sanitycheck: ExprOp 'ite' must have three args!")
+
+            elif args[1].size != args[2].size:
+                raise ValueError(
+                    "sanitycheck: ExprOp 'ite' args must have same size! %s" %
+                                 ([(str(arg), arg.size) for arg in args[1:]]))
 
         if not isinstance(op, str):
             raise ValueError("ExprOp: 'op' argument must be a string")
@@ -771,7 +780,7 @@ class ExprOp(Expr):
             sz = 64
         elif self._op in ['double_to_mem_80', 'double_to_int_80', 'double_trunc_to_int_80']:
             sz = 80
-        elif self._op in ['segm']:
+        elif self._op in ['segm', 'ite']:
             sz = self._args[1].size
         else:
             if None in sizes:
--- a/miasm2/ir/translators/z3_ir.py
+++ b/miasm2/ir/translators/z3_ir.py
@@ -169,6 +169,9 @@ class TranslatorZ3(Translator):
                     res = res << arg
                 elif expr.op == "idiv":
                     res = res / arg
+                elif expr.op == "ite":
+                    res = z3.If(args[0], args[1], args[2])
+                    break
                 else:
                     raise NotImplementedError("Unsupported OP yet: %s" % expr.op)
         elif expr.op == 'parity':


serpilliere:
Hi

I have a problem with this form: some months ago, a Miasm instruction could only generate one IR block (no extra IR blocks). So there was a problem for example in the representation of `cmovz eax, ebx`

This was represented using the following form:
```
EAX = ExprCond(zf, EBX, EAX)
```

At first glimpse, it can be a good way of representing the mnemonic, but there are some drawbacks. What will you answer if I ask you which registers are read/written? Maybe you will answer `zf`, `EBX`  and `EAX` for the read, and EAX for the write.

The fact is that is the `zf` is null, EAX is not written and EBX is not read. If zf is not null, EAX is written, not read, and EBX is read.
If you look the current Miasm IR for the cmoz, you will have:
```
IRDST = ExprCond(zf, lbl_do, lbl_skip)

lbl_do:
EAX = EBX
IRDST = lbl_skip

lbl_ksip:
```

The problem here is transferred from a data flow problem into an execution flow problem. But the fact are there: If you analyze the previous IR we can deduce that:
* ZF is always read
* EAX is written/EBX is read in the first branch
* nothing happen in the other case.

So global analysis should give that EAX is not read in both cases. This is why the ternary operator form has been dropped and replaced by the generation of multiple IR blocks.

mrphrazer:
Hi,

thanks for your answer, @serpilliere :)

I agree that this behaviour is necessary and that statements as ```ite```  yield problems, especially in the IR transformation or as part of an IRA block.

My intention was not to introduce that for IR transformation. I wasn't sure if you would like to integrate IR expressions which aren't used in assembly -> IR transformations.

Instead, it was to provide the opportunity to create more complex IR statements as part of a global program analysis.  For instance, to define additionally conditions/constraints for the SMT solver that have to be satisfied in the case of reasoning about a particular program behaviour.  One example would be reasoning about a basic block's inputs in dependence on the output of one of its predecessors.

fmonjalet:
Hi @mrphrazer,

I do not really get the difference between the `ite` operator and `ExprCond`. Why wouldn't you use `ExprCond` instead of `ExprOp('ite', ...)` in your case?

mrphrazer:
Hi @fmonjalet,

the problem is the z3 translation:
```python
return z3.If(cond != 0, src1, src2)
```
If you have a statement as:
```python
z = (x = y?(100,200))
```
or, in miasm2 expressions:
```python
ExprAff(ExprId('z', 64), ExprCond(ExprAff(ExprId('x', 1), ExprId('y', 1)), ExprId('100', 64), ExprId('200', 64)))
```
For instance, if an ```ExprId```  is part of the condition (an IRA block as ```bl_gen_00000000:None```), then  ```from_ExprId``` will return
```python
return z3.BitVec(str(expr), expr.size)
```
Then, the Z3 translation of the condition will cause an error as 
```python
Z3Exception: True, False or Z3 Boolean expression expected
```

serpilliere:
In fact we cannot have such an IR expression generated from an instruction:
```python
ExprAff(ExprId('z', 64), ExprCond(ExprAff(ExprId('x', 1), ExprId('y', 1)), ExprId('100', 64), ExprId('200', 64)))
```

In Miasm, the `ExprAff`  is not a real "word" of the language, and cannot be used as a "right" operand.

Did you find a mnemonic that generates such a code? If so, it may be a bug.

commial:
Actually, the z3 translation of `ExprAff` is used in https://github.com/cea-sec/miasm/blob/master/miasm2/analysis/depgraph.py in order to resolve path conditions of implicit dependency graph.
It may not be related, but you can take a look at https://github.com/cea-sec/miasm/blob/master/miasm2/expression/simplifications_cond.py for conditions as `A < B`.

mrphrazer:
@serpilliere 
No, I did not found an instruction creating an IR instruction as that.

I think we are disussing two different things.  That PR is not  related to asm -> IR transformation in any kind.

The intention of  ```ite``` is to create a possibility to insert additional conditional statements to the SMT solver.  For example:

"Block C shall be jump to block D if  prior control flow passed from block  A to B, otherwise to block E."
That would be s.th. as
```python
IRDst_c = ite(IRDst_a = address of B, address of D, address of E)
```

Because of that I put a ```ExprAff``` in my prior post. 

@commial 
Thanks! :) 

The problem with that is, passing non-conditional equalities to Z3 via ```ExprAff``` is no problem. But, if the ```ExprAff``` is  the condition for another assignment and part of that assignment is a ExprId, [as stated above](https://github.com/cea-sec/miasm/pull/177#issuecomment-111877268), it will not work.

I currently subclassed that PR and it works fine for me. I just think that it could also be useful for others who would like to pass extented conditions to Z3.

commial:
I may misunderstood, as I'm not very familiar with z3 API but, what is the difference between:
```Python
# your code
z = (x = y?(100,200))
# expected ?
--- ite ---> z3.If( from_expr(x = y), from_expr(100), from_expr(200))
```
And, with current Miasm:
````Python
translate(ExprCond(x - y, 100, 200))
-> z3.If(from_expr(x - y) != 0, from_expr(100), from_expr(200))
```

That is to say, using a `ExprOp('-')` instead of an `ExprAff`. It seems equivalent to me.
Please correct me if I misunderstood your words.

mrphrazer:
Oh,  you are completly right, sorry about that. I misunderstood @serpilliere's comment and  used ```ExprCond``` in the wrong way in [that](https://github.com/cea-sec/miasm/pull/177#issuecomment-111877268) statement.

To be exact, ```ite``` and ```ExprCond``` are the same, if you swap the arguments:
```python
# ite
ite(x = y, 100,200)
# ExprCond
ExprCond(x - y, 200, 100))
```

Thank you all for clarification and sorry about the noise. :)

PR #179, opened 2015-06-16
typofix and logic error

gelim:
Two trivial commits having a glance at the code. Should be self explanatory ;-)
Cheers !


serpilliere:
спасибо!

PR #189, opened 2015-07-03
sandbox: added sandbox for Linux_x86_64

mrphrazer:
Hi,

I wondered why a linux 64 sandbox doesn't exist, so I pushed it.


mrphrazer:
You are right. Than [that](https://github.com/cea-sec/miasm/blob/master/miasm2/analysis/sandbox.py#L356) should also be changed. I did that in the last commit.

serpilliere:
You are right!
Thanks

PR #199, opened 2015-07-29
Fix x86 sem

serpilliere:


miasm2/arch/x86/sem.py
```diff
@@ -2416,9 +2459,65 @@ def rdtsc(ir, instr):
     return e, []
 
 
-# XXX TODO
 def daa(ir, instr):
-    return [], None
+    e = []
```
commial: Here, you take the time to use variables with expressive names, but you keep `e` for the output.
You could, for instance, replace it with `affblock`.

miasm2/arch/x86/sem.py
```diff
@@ -2416,9 +2459,65 @@ def rdtsc(ir, instr):
...
+                             r_al)
+
+    new_al = m2_expr.ExprCond(cond3,
+                              al_c1 + m2_expr.ExprInt8(0x60),
+                              al_c1)
+    e.append(m2_expr.ExprAff(r_al, new_al))
+    return e, []
+
+def das(ir, instr):
+    e = []
```
commial: Same remark as `daa`

miasm2/expression/expression_helper.py
```diff
@@ -533,3 +533,20 @@ def get(cls, size=32, depth=1, clean=True):
             cls.generated_elements = {}
 
         return got
+
+
+def expr_cmpu(arg1, arg2):
```
commial: Actually, you could rewrite it using `expr_cmps`, avoiding equation redundancy (and potential mistakes).
Indeed, `a <u b == (a <s b) ^ a.msb ^ b.msb`.

serpilliere:
You are right for the `e`.
But I will PR a stand alone patch to modify every instructions of every architectures.

For the equation, the error argument is ok, but the resulting code is a bit more complex due to the insertion of two extra `msb`. Anyway, I will submit a modified patch.

PR #206, opened 2015-08-08
Aarch64

serpilliere:


miasm2/arch/aarch64/arch.py
```diff
@@ -0,0 +1,1968 @@
...
+            arg = e.args[1]
+            if isinstance(arg, m2_expr.ExprId):
+                arg = str(arg)
+            elif arg.op == 'LSL' and arg.args[1].arg == 0:
+                arg = str(arg.args[0])
+            else:
+                arg = "%s %s %s" % (arg.args[0], arg.op, arg.args[1])
+            return '[%s, %s]' % (e.args[0], arg)
+
+        fds
```
commial: You should raise an error instead

miasm2/arch/aarch64/arch.py
```diff
@@ -0,0 +1,1968 @@
...
+        e = self.args[index]
+        if not isinstance(e, m2_expr.ExprInt):
+            return
+        ad = e.arg + self.offset
+        l = symbol_pool.getby_offset_create(ad)
+        s = m2_expr.ExprId(l, e.size)
+        self.args[index] = s
+
+    def breakflow(self):
+        if self.name in brcond + ["BR", "BLR", "RET", "ERET", "DRPS", "B", "BL"]:
```
commial: You can replace it with `return self.name in ....`

miasm2/arch/aarch64/arch.py
```diff
@@ -0,0 +1,1968 @@
...
+            return True
+        return False
+
+    def is_subcall(self):
+        if self.name in ["BLR", "BL"]:
```
commial: You can replace it with `return self.name in ....`

miasm2/arch/aarch64/arch.py
```diff
@@ -0,0 +1,1968 @@
...
+            return True
+        return False
+
+    def getdstflow(self, symbol_pool):
+        index = self.mnemo_flow_to_dst_index(self.name)
+        return [self.args[index]]
+
+    def splitflow(self):
+        if self.name in brcond + ["BLR", "BL"]:
```
commial: You can replace it with `return self.name in ....`

miasm2/arch/aarch64/arch.py
```diff
@@ -0,0 +1,1968 @@
...
+            return True
+        return False
+
+    def get_symbol_size(self, symbol, symbol_pool):
```
commial: This comment is not related only to this PR. Some of the elements of this class look like `property`, as `symbol_size`. May for a future arch refactoring

miasm2/arch/aarch64/arch.py
```diff
@@ -0,0 +1,1968 @@
...
+        if not isinstance(e, m2_expr.ExprInt):
+            log.debug('dyn dst %r', e)
+            return
+        off = e.arg - self.offset
+        if int(off % 4):
+            raise ValueError('strange offset! %r' % off)
+        self.args[index] = m2_expr.ExprInt32(off)
+
+    def get_args_expr(self):
+        fds
```
commial: You should raise an error instead

miasm2/arch/aarch64/arch.py
```diff
@@ -0,0 +1,1968 @@
...
+        args = [a for a in self.args]
+        return args
+
+    def get_asm_offset(self, expr):
+        fds
```
commial: You should raise an error instead

miasm2/arch/aarch64/arch.py
```diff
@@ -0,0 +1,1968 @@
...
+        self.parent.shift.value = shift_expr.index(self.expr.op)
+        self.parent.imm.value = int(args[1].arg)
+        self.value = self.reg_info[size].expr.index(args[0])
+        return True
+
+    def decode(self, v):
+        size = 64 if self.parent.sf.value else 32
+        e = self.reg_info[size].expr[v]
+        amount = self.parent.imm.value
+        if amount == 0:
```
commial: You can rewrite it as `if amount != 0` and remove the `else`

miasm2/arch/aarch64/arch.py
```diff
@@ -0,0 +1,1968 @@
...
+
+
+class additional_info:
+
+    def __init__(self):
+        self.except_on_instr = False
+        self.lnk = None
+        self.cond = None
+
+conds = [
```
commial: `conds`, `conds_inv`, `brcond`, `uints`, ... are constants. You may rewrite them in upper case, to clarify the code and distinguish them with vars.

miasm2/arch/aarch64/arch.py
```diff
@@ -0,0 +1,1968 @@
...
+    def get_symbol_size(self, symbol, symbol_pool, mode):
+        return 32
+
+    def reset_class(self):
+        super(mn_aarch64, self).reset_class()
+        if hasattr(self, "sf"):
+            self.sf.value = None
+
+
+def aarch64op(name, fields, args=None, alias=False):
```
commial: As for constants, one should begins the name of classes with an upper letter, in CamlCase [PEP naming convention](https://www.python.org/dev/peps/pep-0008/#naming-conventions). I know this is not the case for a part of other classes in Miasm (as `ir`, `irbloc`, ...), but we should try to respect it in new code snippets.

miasm2/arch/aarch64/arch.py
```diff
@@ -0,0 +1,1968 @@
...
+    dct = {"fields": fields}
+    dct["alias"] = alias
```
commial: This line can be merged with `{"fields": fields}` by adding `"alias": alias`.

miasm2/arch/aarch64/disasm.py
```diff
@@ -0,0 +1,27 @@
...
+        func(mn, attrib, pool_bin, cur_bloc, offsets_to_dis, symbol_pool)
+
+
+class dis_aarch64b(disasmEngine):
+    attrib = 'b'
+
+    def __init__(self, bs=None, **kwargs):
+        super(dis_aarch64b, self).__init__(
+            mn_aarch64, self.attrib, bs, **kwargs)
+        self.dis_bloc_callback = cb_aarch64_disasm
```
commial: This line can be passed as a keyword argument to `super(dis_aarch64b, self).__init__(..., dis_bloc_callback=...`

miasm2/arch/aarch64/disasm.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class dis_aarch64l(disasmEngine):
+    attrib = 'l'
+
+    def __init__(self, bs=None, **kwargs):
```
commial: This is exactly the same code that for `dis_aarch64b` and for some architecture.
May the code of `disasmEngine`, which seems to be an abstract class, should use `self.attrib` and `self.dis_bloc_callback` instead of taking it as an `__init__` argument.
Indeed, it is related to the class, not to the instance.

Then, this code can be rewritten:
```Python
class dis_aarch64l(disasmEngine):
    attrib = 'l'
    dis_bloc_callback = cb_aarch64_disasm

class dis_aarch64b(disasmEngine):
    attrib = 'b'
    dis_bloc_callback = cb_aarch64_disasm
```
It seems clearer for me, and, in addition, it avoids one to think that there is a special treatment in the `__init__` for *this* architecture.

miasm2/arch/aarch64/ira.py
```diff
@@ -0,0 +1,135 @@
...
+        irs = [[ExprAff(self.ret_reg, ExprOp('call_func_ret', ad, self.sp)),
+                ExprAff(self.sp, ExprOp('call_func_stack', ad, self.sp)),
+                ]]
+        return irs
+
+    def post_add_bloc(self, bloc, ir_blocs):
+        ir.post_add_bloc(self, bloc, ir_blocs)
+        # flow_graph = DiGraph()
+        for irb in ir_blocs:
+            # print 'X'*40
```
commial: If these print are here for debug, you should use a logger.DEBUG instead.

miasm2/arch/aarch64/ira.py
```diff
@@ -0,0 +1,135 @@
...
+            lbl = bloc.get_next()
+            new_lbl = self.gen_label()
+            irs = self.call_effects(pc_val)
+            irs.append([ExprAff(self.IRDst, ExprId(lbl, size=self.pc.size))])
+            nbloc = irbloc(new_lbl, irs)
+            nbloc.lines = [l] * len(irs)
+            self.blocs[new_lbl] = nbloc
+            irb.dst = ExprId(new_lbl, size=self.pc.size)
+
+        """
```
commial: This code is already saved in the CVS. Then, the comment is useless.

miasm2/arch/aarch64/sem.py
```diff
@@ -0,0 +1,745 @@
...
+@sbuild.parse
+def bl(arg1):
+    PC = arg1
+    ir.IRDst = arg1
+    LR = m2_expr.ExprInt64(instr.offset + instr.l)
+
+
+def csel(ir, instr, arg1, arg2, arg3, arg4):
+    e = []
+    cond_expr = cond2expr[arg4.name]
```
commial: Is this kind of expression unsupported in `@sbuild.parse` ?
If so, please open a feature request. These functions (with `csinc`, `csinv`, ...) may be more understandable as:
```Python
arg1 = cond_expr[arg4.name] if arg2 else arg3
```

miasm2/os_dep/linux_stdlib.py
```diff
@@ -4,6 +4,19 @@
...
 from string import printable
 
+from miasm2.os_dep.common import \
+    heap, set_str_ansi, set_str_unic, get_str_ansi, get_str_unic
+
+
+class c_linobjs:
+
+    def __init__(self):
+        self.alloc_ad = 0x20000000
```
commial: As `alloc_*` are constants, you can move them to the definition of `c_linobjs`.

miasm2/os_dep/linux_stdlib.py
```diff
@@ -4,6 +4,19 @@
 from sys import stdout
 from string import printable
 
+from miasm2.os_dep.common import \
+    heap, set_str_ansi, set_str_unic, get_str_ansi, get_str_unic
+
+
+class c_linobjs:
```
commial: Every classes should inherit from `object`

test/arch/aarch64/unit/asm_test.py
```diff
@@ -0,0 +1,65 @@
...
+        blocs, symbol_pool = parse_asm.parse_txt(mn_aarch64, 'l', self.TXT,
+                                                 symbol_pool = self.myjit.ir_arch.symbol_pool)
+        # fix shellcode addr
+        symbol_pool.set_offset(symbol_pool.getby_name("main"), 0x0)
+        s = StrPatchwork()
+        patches = asmbloc.asm_resolve_final(mn_aarch64, blocs[0], symbol_pool)
+        for offset, raw in patches.items():
+            s[offset] = raw
+
+        s = str(s)
```
commial: Useless line

miasm2/arch/aarch64/disasm.py
serpilliere: This doesn't work actually. The `disasmengine` take `arch`as argument.

test/test_all.py
```diff
@@ -277,6 +282,10 @@ def __init__(self, *args, **kwargs):
                               "0"], depends=[test_armtl])
 testset += ExampleDisasmFull(["armtb", Example.get_sample("demo_armt_b.bin"),
                               "0"], depends=[test_armtb])
+testset += ExampleDisasmFull(["aarch64l", Example.get_sample("demo_aarch64_l.bin"),
+                              "0"], depends=[test_aarch64l])
+testset += ExampleDisasmFull(["aarch64b", Example.get_sample("demo_aarch64_b.bin"),
+                              "0"], depends=[test_armb])
```
commial: It seems that this test depends on `test_aarch64b`, not `test_armb`

miasm2/arch/aarch64/arch.py
serpilliere: Maybe in another pr

miasm2/arch/aarch64/disasm.py
commial: I missed it. You just have to add a `arch = mn_aarch64` in the class definition

miasm2/arch/aarch64/arch.py
```diff
@@ -0,0 +1,1944 @@
...
+        size = 64 if self.parent.sf.value else 32
+        if self.parent.shift.value == 0:
+            self.expr = m2_expr.ExprInt_fromsize(size, v)
+        elif self.parent.shift.value == 1:
+            self.expr = m2_expr.ExprInt_fromsize(size, v << 12)
+        else:
+            return False
+        return True
+
+option2size = [32, 32, 32, 64,
```
commial: Sorry for not quoting every vars to rewrote as upper case. This is a remark which is apply-able on others file of this commit, and for `option2size`, `ext2_*`

test/arch/aarch64/unit/asm_test.py
serpilliere: the s=str(s) is needed actually.

miasm2/arch/aarch64/disasm.py
serpilliere: attrib='b' as argument doesn't work either.

test/arch/aarch64/unit/asm_test.py
commial: Yep, but the intermediate `s` var is useless, the code is really understandable without and it avoid an indirection

PR #212, opened 2015-08-18
Aarch64 fix

serpilliere:


miasm2/arch/aarch64/sem.py
```diff
@@ -730,6 +793,36 @@ def irbloc_fix_regs_for_mode(self, irbloc, mode=64):
                 irs[i] = self.expr_fix_regs_for_mode(e)
         irbloc.dst = self.expr_fix_regs_for_mode(irbloc.dst)
 
+    def mod_pc(self, instr, instr_ir, extra_ir):
+        "Replace PC by the instruction's offset"
+        cur_offset = m2_expr.ExprInt64(instr.offset)
+        for i, expr in enumerate(instr_ir):
```
commial: This is exactly the same code snippet that line 807, you could move it to a protected method

miasm2/arch/aarch64/sem.py
```diff
@@ -730,6 +793,36 @@ def irbloc_fix_regs_for_mode(self, irbloc, mode=64):
...
+                    irs[i] = m2_expr.ExprAff(dst, src)
+
+
+    def del_dst_zr(self, instr, instr_ir, extra_ir):
+        "Writes to zero register are discarded"
+        instr_ir = [expr for expr in instr_ir if expr.dst not in [WZR, XZR]]
+
+        for b in extra_ir:
+            for i, irs in eunmerate(b.irs):
+                irs = [expr for expr in irs if expr.dst not in [WZR, XZR]]
```
commial: Here, you rewrite `irs`, the loop variable. You may merge this line with the next one:
`b.irs[i] = [expr for expr in irs if expr.dst not in [WZR, XZR]]`
Or with a `map`:
`b = map(lambda irs:  [expr for expr in irs if expr.dst not in [WZR, XZR]], b)`

In addition, `[WZR, XZR]` is redefined in each iteration, you should get it off the loop inside a `CONST`.

miasm2/arch/aarch64/sem.py
serpilliere: Correct: I will factorize this code among others architectures as well in another PR.

PR #213, opened 2015-08-21
Dependency graph

carolineLe:
This PR improves the DependencyGraph algorithm :  in the analysis, when two parrallel expressions depend on the same variable, only one DependencyNode is created.

### DIFF
--- a/miasm2/analysis/depgraph.py
+++ b/miasm2/analysis/depgraph.py
@@ -68,9 +68,10 @@ class DependencyNode(object):
 
     def __str__(self):
         """Returns a string representation of DependencyNode"""
-        return "<%s %s %s %s M:%s>" % (self.__class__.__name__,
-                                       self.label.name, self.element,
-                                       self.line_nb, self.modifier)
+        return "<%s %s %s %s M:%s S:%s>" % (self.__class__.__name__,
+                                            self.label.name, self.element,
+                                            self.line_nb, self.modifier,
+                                            self.step)
 
     def __repr__(self):
         """Returns a string representation of DependencyNode"""
@@ -110,9 +111,8 @@ class DependencyNode(object):
     @modifier.setter
     def modifier(self, value):
         """Evaluating the current line involves a modification of tracked
-        dependencies"""
-        if not isinstance(value, bool):
-            raise ValueError("Modifier must be a boolean")
+        dependencies if @value.
+        @value: boolean"""
         self._modifier = value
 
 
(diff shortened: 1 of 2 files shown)


miasm2/analysis/depgraph.py
```diff
@@ -110,7 +111,8 @@ def modifier(self):
     @modifier.setter
     def modifier(self, value):
         """Evaluating the current line involves a modification of tracked
-        dependencies"""
+        dependencies
+        @value: boolean"""
         if not isinstance(value, bool):
```
commial: We can consider that the docstring is enough and then the check is superfluous (in addition, this methods has speed concern).

miasm2/analysis/depgraph.py
```diff
@@ -246,7 +248,9 @@ def pending(self):
     def _get_modifiers_in_cache(self, depnode, force=False):
         """Recursively find nodes in the path of @depnode which are modifiers.
         Update the internal cache
-        If @depnode is already managed (ie. in @depnode_queued), abort"""
+        If @depnode is already managed (ie. in @depnode_queued), abort
+        @depnode: DependencyNode
+        @force (optionnal): boolean"""
```
commial: This is a protected method, so docstring can be omitted (but it's still better to add it).
But if you add it for args, please specify what are their purposes, not only their expected types.

miasm2/analysis/depgraph.py
```diff
@@ -388,12 +394,19 @@ def filter_unmodifier_loops(self, implicit, irdst):
...
-            # Remove unused elements
+            # Replace unused keys by previous ones
             for key in to_remove:
                 if depnode.nostep_repr == key.nostep_repr:
                     self._cache[depnode] = self._cache.get(key, set()).copy()
                     self.pending.discard(key)
                     self.pending.add(depnode)
+
+                    # Replace occurences of key to remove
+                    for dependencies in self._cache.values():
```
commial: You should use `itervalues` instead

test/analysis/depgraph.py
```diff
@@ -8,6 +8,13 @@
     DependencyDict
 from itertools import count
 
+EMULATION=True
+try:
+    import z3
+except ImportError:
+    EMULATION=False
+    pass
```
commial: `pass` is useless here

test/analysis/depgraph.py
```diff
@@ -133,6 +115,12 @@ def __eq__(self, graph):
         return True
 
     def node2str(self, node):
+        if isinstance(node, asm_label):
+            if not node in self.ira.blocs:
```
commial: The statement `... not in ...` is more readable than `not ... in ...`

PR #215, opened 2015-08-26
Depgraph modifier

serpilliere:
Improve `depgraph` modifier complexity

### DIFF
--- a/miasm2/analysis/depgraph.py
+++ b/miasm2/analysis/depgraph.py
@@ -25,6 +25,8 @@ class DependencyNode(object):
     line.
     """
 
+    __slots__ = ["_label", "_element", "_line_nb", "_modifier",
+                 "_step", "_nostep_repr", "_hash"]
     def __init__(self, label, element, line_nb, step, modifier=False):
         """Create a dependency node with:
         @label: asm_label instance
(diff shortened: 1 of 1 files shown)


miasm2/analysis/depgraph.py
```diff
@@ -243,47 +246,52 @@ def pending(self):
...
+                    if not depnode.modifier:
+                        cache[depnode] = []
+                    continue
+
+                dependencies = self._cache[depnode]
+                for son in dependencies:
+                    worklist_depnodes.append(son)
+                worklist.append((depnode, dependencies))
+
+            while worklist and not worklist_depnodes:
```
commial: Once you get here, `not worklist_depnodes` is always `True`, and `worklist_depnodes` is never modified in the loop.
So you can replace it with `while worklist`, more comprehensive

miasm2/analysis/depgraph.py
```diff
@@ -243,47 +246,52 @@ def pending(self):
...
-
-    def clean_modifiers_in_cache(self):
+    def _get_modifiers_in_cache(self, nodes_heads):
+        """Find modifier nodes in cache starting from @nodes_heads.
+        Returns new cache"""
+        worklist = []
+        worklist_depnodes = list(nodes_heads)
+        cache = {}
+        forced_nodes = set(nodes_heads)
+        while worklist or worklist_depnodes:
```
commial: To reach the end of the loop, we first reach the end of `while worklist_depnodes` at line 274, so `if worklist_depnodes` is `False`. During the second loop, `worklist_depnodes` is never modified, so at the end of it (line 286), we both have `worklist_depnodes` and `worklist`evaluated to `False` in condition statements. So the loop will always occurs one time max, so this `while worklist or worklist_depnodes:` can be omitted.

PR #216, opened 2015-08-28
Analysis/InfluenceGraph: Influence graph implementation

carolineLe:
InfluenceGraph implementation using dependncygraph algorithm.
TODO: Code refactoring

### DIFF
--- a/miasm2/analysis/depgraph.py
+++ b/miasm2/analysis/depgraph.py
@@ -202,7 +202,7 @@ class DependencyDict(object):
 
         # Initialize
         new_history = list(self.history)
-        depdict = DependencyDict(self.label, new_history)
+        depdict = self.__class__(self.label, new_history)
 
         # Copy values
         for key, values in self.cache.iteritems():
@@ -215,7 +215,7 @@ class DependencyDict(object):
         """Return a copy of itself, with itself in history and pending clean
         @label: asm_label instance for the new DependencyDict's label
         """
-        depdict = DependencyDict(label, list(self.history) + [self])
+        depdict = self.__class__(label, list(self.history) + [self])
         for key, values in self.cache.iteritems():
             depdict.cache[key] = set(values)
         return depdict
@@ -693,8 +693,8 @@ class DependencyGraph(object):
 
         Following arguments define filters used to generate dependencies
         @apply_simp: (optional) Apply expr_simp
-        @follow_mem: (optional) Track memory syntactically
-        @follow_call: (optional) Track through "call"
+        @follow_mem: (optional) Track memory subexpressions
+        @follow_call: (optional) Track call parameters
         """
         # Init
         self._ira = ira
--- a/miasm2/analysis/influencegraph.py
+++ b/miasm2/analysis/influencegraph.py
@@ -1,0 +1,298 @@
+"""Provide influence graph"""
+
+import itertools
+from collections import deque
+
+import miasm2.expression.expression as m2_expr
+from miasm2.analysis.depgraph import DependencyNode, DependencyDict, \
+    DependencyResult, DependencyResultImplicit, FollowExpr, DependencyGraph
+from miasm2.expression.simplifications import expr_simp
+from miasm2.core.asmbloc import expr_is_label
+from miasm2.core.graph import DiGraph
+
+class InfluencyDict(DependencyDict):
+
+    """Internal structure for the InfluenceGraph algorithm"""
+
+    def __init__(self, label, history):
+        """Create a DependencyDict
+        @label: asm_label, current IRblock label
+        @history: list of DependencyDict
+        """
+        super(self.__class__, self).__init__(label, history)
+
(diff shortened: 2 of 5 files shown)


test/test_all.py
```diff
@@ -153,6 +153,24 @@ def __init__(self, arch, container, address, *args, **kwargs):
...
+                           "graph_test_06_01.dot",
+                           "graph_test_08_01.dot",
+                           "graph_test_09_01.dot",
+                           "graph_test_10_01.dot",
+                           "graph_test_13_01.dot",
+                           "graph_test_14_01.dot",
+                           "graph_test_14_02.dot",
+                           "graph_test_15_01.dot"
+                       ])
+
```
serpilliere: I think there is a problem with generated tests:
There is an interference between tests generated by the `depgraph` and tests generated by the `influence` graph, as both have tests'name in common. Running regression tests on a multicpu machine will generate warnings:
```
Cleanning error: Unable to remove ..../test/analysis/graph_test_04_01.dot
Cleanning error: Unable to remove ..../test/analysis/graph_test_06_01.dot
Cleanning error: Unable to remove ..../test/analysis/graph_test_14_02.dot
```

Maybe you need to modify generated tests'names for example  `graph_dependence_test_13_01.dot` and `graph_influence_test_13_01.dot` to avoid such interference.

miasm2/analysis/depgraph.py
```diff
@@ -759,15 +759,13 @@ def _follow_mem_wrapper(exprs, mem_read):
             follow.update(expr.get_r(mem_read=mem_read, cst_read=True))
         return follow, set()
 
-    @staticmethod
-    def _follow_mem(exprs):
+    def _follow_mem(self, exprs):
```
commial: This method depends on the class, not on the instance itself. So it should be a `classmethod`.

miasm2/analysis/depgraph.py
```diff
@@ -759,15 +759,13 @@ def _follow_mem_wrapper(exprs, mem_read):
...
         """Follow expression from memory pointer"""
-        return DependencyGraph._follow_mem_wrapper(exprs, True)
+        return self._follow_mem_wrapper(exprs, True)
 
-    @staticmethod
-    def _follow_nomem(exprs):
+    def _follow_nomem(self, exprs):
```
commial: Same remark as `_follow_mem`

miasm2/analysis/influencegraph.py
```diff
@@ -0,0 +1,293 @@
+"""Provide influence graph"""
+
+import miasm2.expression.expression as m2_expr
+from miasm2.analysis.depgraph import DependencyNode, DependencyDict, \
+    DependencyResult, DependencyResultImplicit, FollowExpr, DependencyGraph
+from miasm2.expression.simplifications import expr_simp
+from miasm2.core.asmbloc import expr_is_label
+from miasm2.core.graph import DiGraph
+
+import itertools
```
commial: For more homogeneity, please first import Python module as `itertools` and `collections`, and then import miasm2 relative modules (take a look at [PEP08](https://www.python.org/dev/peps/pep-0008/#imports) for details).

test/analysis/influencegraph.py
```diff
@@ -0,0 +1,945 @@
+from miasm2.expression.expression import ExprId, ExprInt32, ExprAff, ExprMem, ExprCond
+from miasm2.core.asmbloc import asm_label
+from miasm2.ir.analysis import ira
+from miasm2.ir.ir import ir, irbloc
+from miasm2.core.graph import DiGraph
+from miasm2.analysis.influencegraph import DependencyNode, InfluenceGraph,\
+    DependencyDict
+import itertools
```
commial: Same remark regarding import order

test/analysis/influencegraph.py
```diff
@@ -0,0 +1,945 @@
...
+class InfluencyNode(DependencyNode):
+    counter = itertools.count()
+
+    def __init__(self, label, element, line_nb, modifier=False):
+        DependencyNode.__init__(self, label, element, line_nb, next(self.counter),
+                                modifier=modifier)
+
+
+def gen_irbloc(lbl, exprs):
+    lines = [None for _ in xrange(len(exprs))]
```
commial: You can rewrite it `lines = [None] * len(exprs)`

miasm2/analysis/depgraph.py
serpilliere: `influencegraph` subclass `depgraph` , so a call to `copy` or `extend` in incluence graph will generate a depgraph. As we need history and self, this the `copy` and `extend` cannot be a class method.

commial: My remark only concerns `_follow_mem` and `_follow_nomem`, as they both call `_follow_mem_wrapper`, which is a static method in both `depgraph` and `influencegraph`.

But your note still stands for call to `copy` and `extend` in influence graph.

PR #218, opened 2015-09-04
Extend digraph with loop detection algorithms

mrphrazer:
Part of this PR:
- search-first walks
- natural loop detection
- strongly connected components detection

### DIFF
--- a/miasm2/core/graph.py
+++ b/miasm2/core/graph.py
@@ -1,4 +1,4 @@
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 
 class DiGraph(object):
     """Implementation of directed graph"""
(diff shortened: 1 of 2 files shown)


miasm2/core/graph.py
```diff
@@ -339,3 +339,167 @@ def compute_dominance_frontier(self, head):
...
+
+        # init index with 0
+        index = {v: 0 for v in self.nodes()}
+
+        for node in self.nodes():
+            # next node if node was already visited
+            if index[node]:
+                continue
+
+            todo = [(0, node)]
```
commial: `todo` elements could come from a `namedtuple`. That way, the code will be easier to understand, and less prone to error (the tuple in this list is always composed of {1,2,3}, node).

mrphrazer:
Thanks for the feedback!

See the last commits :)

mrphrazer:
@commial: I also don't like it that way, I misunderstood `namedtuple`.  How would you use/define the `namedtuple`, I don't see the advantage so far.

serpilliere:
Ok. May that's commial might have said:

With `namedutple` the code transforms to:
```python
NodeState = namedtuple('NodeState', ['state', 'node'])
...
todo = [NodeState(VISIT, node)]
...
cur_node = todo.pop()
if  cur_node.state == VISIT:
    stack.append(cur_node.node)
...
todo.append(NodeState(MERGE, node))
```

mrphrazer:
Thanks that sounds good. Fixed.

serpilliere:
Great PR!

PR #226, opened 2015-10-10
Optimize jitter execution time

fmonjalet:
Hi,

This PR speeds up some of the python code used by the jitter:
- `BoundedDict`'s `Counter` is replaced with a simple `dict` {key -> int}. This makes the run loop twice as fast on programs where compilation time is way inferior to the jitted code execution time (typically programs with intensive loops).
- in `runiter_once`, a preliminary check is made before checking all the callbacks. This brings something like 20% speedup.

Globally, most of the time is wasted by the huge function call overhead of python. As discussed previously with @serpilliere, moving the whole `continue_run`, or at least `runiter_once` loop in C would grant a great speedup (an additional x5 or x10 seems pretty reachable).

Profiling have mainly been performed with cProfile and Runsnakerun.

Here is how it looked like before the PR:
profile_before

Here is what it looks like now:
profile_after

**Note:** `get_exception` can be inlined in `runiter_once` to save some function calls (it is the most called function for this sample), but I think the very slight speedup (non negligeable when benchmarking, but negligeable when running the program without the profiler) is not worth the lack of clarity it brings.

### DIFF
--- a/miasm2/core/utils.py
+++ b/miasm2/core/utils.py
@@ -75,7 +75,7 @@ class BoundedDict(UserDict.DictMixin):
         self._max_size = max_size
         self._size = len(self._data)
         # Do not use collections.Counter as it is quite slow
-        self._counter = dict((k, 1) for k in self._data.iterkeys())
+        self._counter = {k: 1 for k in self._data}
         self._delete_cb = delete_cb
 
     def __setitem__(self, asked_key, value):
@@ -99,7 +99,7 @@ class BoundedDict(UserDict.DictMixin):
                 self._size = self._min_size
 
                 # Reset use's counter
-                self._counter = dict((k, 1) for k in self._data.iterkeys())
+                self._counter = {k: 1 for k in self._data}
 
             # Avoid rechecking in dict: set to 1 here, add 1 otherwise
             self._counter[asked_key] = 1
--- a/miasm2/core/utils.py
+++ b/miasm2/core/utils.py
@@ -1,6 +1,7 @@
 import struct
 import inspect
 import UserDict
+from operator import itemgetter
 
 upck8 = lambda x: struct.unpack('B', x)[0]
 upck16 = lambda x: struct.unpack('H', x)[0]
@@ -73,7 +74,8 @@ class BoundedDict(UserDict.DictMixin):
         self._min_size = min_size if min_size else max_size / 3
         self._max_size = max_size
         self._size = len(self._data)
-        self._counter = collections.Counter(self._data.keys())
+        # Do not use collections.Counter as it is quite slow
+        self._counter = dict((k, 1) for k in self._data.iterkeys())
         self._delete_cb = delete_cb
 
     def __setitem__(self, asked_key, value):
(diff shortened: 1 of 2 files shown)


miasm2/core/utils.py
```diff
@@ -73,7 +74,8 @@ def __init__(self, max_size, min_size=None, initialdata=None,
         self._min_size = min_size if min_size else max_size / 3
         self._max_size = max_size
         self._size = len(self._data)
-        self._counter = collections.Counter(self._data.keys())
+        # Do not use collections.Counter as it is quite slow
+        self._counter = dict((k, 1) for k in self._data.iterkeys())
```
commial: As we don't care about Python 2.6, you can use the comprehension dict construction

miasm2/core/utils.py
```diff
@@ -83,31 +85,46 @@ def __setitem__(self, asked_key, value):
...
 
                 # Keep only the most @_min_size used
                 self._data = {key:self._data[key]
-                              for key in most_commons[:self._min_size - 1]}
+                              for key, _ in most_common[:self._min_size - 1]}
                 self._size = self._min_size
 
                 # Reset use's counter
-                self._counter = collections.Counter(self._data.keys())
+                self._counter = dict((k, 1) for k in self._data.iterkeys())
```
commial: Same remark for dict comprehension

miasm2/core/utils.py
fmonjalet: Alright!

p-l-: Also, you can iterate over the dictionary object directly, instead of calling `.iterkeys()`.

miasm2/core/utils.py
p-l-: Same remark for `.iterkeys()` ;-)

miasm2/jitter/jitload.py
```diff
@@ -134,13 +137,19 @@ class CallbackHandlerBitflag(CallbackHandler):
 
     "Handle a list of callback with conditions on bitflag"
 
+    # Overrides CallbackHandler's implem, but do not serve for optimization
+    def has_callbacks(self, bitflag):
```
p-l-: This function can be rewritten as (more readable, in my opinion):
```Python
    return any(b & bitflag != 0 for b in self.callbacks)
```
This function would return either `True` or `False` instead of either `True` or `None`, by the way.

miasm2/jitter/jitload.py
```diff
@@ -134,13 +137,19 @@ class CallbackHandlerBitflag(CallbackHandler):
...
+                return True
+
     def __call__(self, bitflag, *args):
         """Call each callbacks associated with bit set in bitflag. While
         callbacks return True, continue with next callback.
         Iterator on other results"""
 
         res = True
-        for b in self.callbacks.keys():
+        for b in self.callbacks.iterkeys():
```
p-l-: Same remark about `.iterkeys()`

miasm2/jitter/jitload.py
serpilliere: Moreover, you might use `cbs` instead of `b` (for example).

miasm2/core/utils.py
fmonjalet: Hum, yeah, well seen. :)

fmonjalet:
I think I took all the comment into account, see last commits.

miasm2/core/utils.py
```diff
@@ -83,31 +85,46 @@ def __setitem__(self, asked_key, value):
 
             # Bound can only be reached on a new element
             if (self._size >= self._max_size):
-                most_commons = [key for key, _ in self._counter.most_common()]
+                most_common = sorted(self._counter.iteritems(),
+                                     key=itemgetter(1), reverse=True)
```
p-l-: Just being curious: what are the advantages of importing and using `itemgetter` over using a `lambda` expression: `key=lambda item: item[1]` or `lambda (_, count): count`? Thanks!

fmonjalet: Actually, this is the implementation used in `collections.Counter`, I took it as is, trusting the python guys.
But this is a good question. :)
It seems to have better performance than the lambda function, even in trivial cases: http://stackoverflow.com/questions/17243620/operator-itemgetter-or-lambda.

For the import part, it is imported by `collections` anyway, and this module imports `collections`.

p-l-: Thanks! (and sorry, I could have asked Google, I have been lazy ;-) )

PR #232, opened 2015-10-21
Fixes for the tcc jitter

a-vincent:
The first two commits are required for me to use the tcc jitter on my Linux box.
The third one is more of a cosmetic nature but should prevent picking a wrong Jittcc module.

Each commit comes with a description so I'll keep the pull request message short.

### DIFF
--- a/miasm2/jitter/jitcore_tcc.py
+++ b/miasm2/jitter/jitcore_tcc.py
@@ -2,11 +2,11 @@
 #-*- coding:utf-8 -*-
 
 import os
-from miasm2.ir.ir2C import irblocs2C
-from subprocess import Popen, PIPE
-import miasm2.jitter.jitcore as jitcore
 from distutils.sysconfig import get_python_inc
-import Jittcc
+from subprocess import Popen, PIPE
+from miasm2.ir.ir2C import irblocs2C
+from miasm2.jitter import jitcore
+from miasm2.jitter import Jittcc
 
 
 def jit_tcc_compil(func_name, func_code):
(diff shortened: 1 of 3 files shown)


serpilliere:
In fact this is not the first modification of `tcc` api. That's why we only support the latest stable libtcc (as described in Miasm README).

a-vincent:
Ah OK, sorry then. Feel free not to merge that commit.

miasm2/jitter/jitcore_tcc.py
```diff
@@ -2,11 +2,11 @@
 #-*- coding:utf-8 -*-
 
 import os
-from miasm2.ir.ir2C import irblocs2C
-from subprocess import Popen, PIPE
-import miasm2.jitter.jitcore as jitcore
 from distutils.sysconfig import get_python_inc
-import Jittcc
+from subprocess import Popen, PIPE
+from miasm2.ir.ir2C import irblocs2C
```
commial: Actually, according to pep8 convention and in order to be homogeneous in Miasm, import order is divided between python common (as `subprocess`) and miasm import (as `miasm2. ...`) with an empty line. So, an empty line is missing here.
.

miasm2/jitter/jitcore_tcc.py
```diff
@@ -2,11 +2,11 @@
...
+from miasm2.jitter import jitcore
```
commial: These 2 lines can be merged in `... import jitcore, Jittccc`

a-vincent:
After investigation, using a stable tcc tarball doesn't produce the problems I was seeing which prompted this pull request. I'm closing it and will reopen a new one just for the cleanup of imports in miasm2/jitter/jitcore_tcc.py so as not to lose the fruit of this discussion.

PR #235, opened 2015-10-22
X86 fix

serpilliere:
Multiple fixes in `x86` semantic/architecture.


miasm2/arch/x86/regs.py
```diff
@@ -342,6 +342,22 @@
 float_st6 = ExprId("float_st6", 64)
 float_st7 = ExprId("float_st7", 64)
 
+float_replace = {ExprId('ST', 64): float_st0,
+                 fltregs32_expr[0]: float_st0,
```
commial: By declaring `float_list` before, you can use:
```Python
float_replace = {fltregs32_expr[i]: float_list[i] for i in xrange(8)}
float_replace[ExprId('ST', 64)] = float_st0
```

miasm2/arch/x86/regs.py
```diff
@@ -342,6 +342,22 @@
 float_st6 = ExprId("float_st6", 64)
 float_st7 = ExprId("float_st7", 64)
 
+float_replace = {ExprId('ST', 64): float_st0,
```
commial: `ExprId('ST', 64)` is already declared at line 88, may a variable can avoid future error.

miasm2/expression/expression.py
```diff
@@ -760,13 +761,20 @@ def __init__(self, op, *args):
                           'int_16_to_double', 'int_32_to_double',
                           'int_64_to_double', 'int_80_to_double']:
             sz = 64
-        elif self._op in ['double_to_mem_16', 'double_to_int_16', 'double_trunc_to_int_16']:
+        elif self._op in ['double_to_mem_16', 'double_to_int_16',
```
commial: For `double_to_mem_`, `double_to_int_`, `double_trunc_to_int_` it seems that the size is embedded in the `ExprOp` name.
Using a ```if ... startswith``` and getting the size at the end  may allow to automatically supporting every size for these instructions (if this is a wanted behavior), avoiding copying the same code for each new size to support.

serpilliere: I prefer (for the moment) to keep a white list for such sanity checks

PR #238, opened 2015-10-23
Feature cloop

commial:
This PR modifies the way `jittcc_exec_bloc` work: now, the next block to be executed is fetched during the C loop when it's possible (no exception has been raised, no breakpoint to handle, the block is already jitted).
That way, in many case (such as *inlined strlen*), there is no need to go back and forth between Python and execution, leading to a faster execution.
On my tests, it cut execution time by 3 to 5 (on 10 min examples).

### DIFF
--- a/miasm2/jitter/jitload.py
+++ b/miasm2/jitter/jitload.py
@@ -137,11 +137,7 @@ class CallbackHandlerBitflag(CallbackHandler):
 
     "Handle a list of callback with conditions on bitflag"
 
-    # Overrides CallbackHandler's implem, but do not serve for optimization
-    def has_callbacks(self, bitflag):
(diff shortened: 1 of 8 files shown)


miasm2/jitter/jitload.py
```diff
@@ -137,11 +137,7 @@ class CallbackHandlerBitflag(CallbackHandler):
 
     "Handle a list of callback with conditions on bitflag"
 
-    # Overrides CallbackHandler's implem, but do not serve for optimization
-    def has_callbacks(self, bitflag):
```
fmonjalet: A bit too late, but you left `has_callbacks` in the parent class `CallbackHandler`, but the implementation is inconsistent with the way `CallbackHandlerBitFlag` works. Maybe it should be removed too.

PR #252, opened 2015-11-01
JitTcc: fix memory leak

serpilliere:


miasm2/jitter/Jittcc.c
```diff
@@ -141,6 +141,9 @@ PyObject* tcc_exec_bloc(PyObject* self, PyObject* args)
 	if (!PyArg_ParseTuple(args, "OOOO", &retaddr, &jitcpu, &lbl2ptr, &breakpoints))
 		return NULL;
 
+	/* The loop will decref retaddr always once */
```
commial: This is only true because the caller ensures that `func_py = PyDict_GetItem(lbl2ptr, retaddr);` returns a pointer at least the first time.
The comment may be extended for a better comprehension (I need a few minute to figure it out).

PR #284, opened 2015-11-24
Feature MemStruct

fmonjalet:
This PR introduces an API to easily interact with C structures in miasm's sandbox.

The `example/jitter/memstruct.py` example file may be the best introduction to this feature. As a spoiler, here is how a linked list can be represented with this API (extracted from the aforementioned file):

```python
class ListNode(MemStruct):
    fields = [
        # The "<I" is the struct-like format of the pointer in memory, in this
        # case a Little Endian 32 bits unsigned int
        # One way to handle reference to ListNode in ListNode is to use the
        # special marker MemSelf.
        # You could also set or modify ListNode.fields after the class
        # declaration and call ListNode.gen_fields()
        ("next", Ptr("<I", MemSelf)),
        # Ptr(_, MemVoid) is analogous to void*, MemVoid is just an empty
        # MemStruct type
        ("data", Ptr("<I", MemVoid)),
    ]

class LinkedList(MemStruct):
    fields = [
        ("head", Ptr("<I", ListNode)),
        ("tail", Ptr("<I", ListNode)),
        # Num can take any one-field struct-like format, including floats and
        # doubles
        ("size", Num("<I")),
    ]

# [...]

link = LinkedList(jitter.vm, some_addr)
link.deref_head.data = other_addr
link.size += 1
# etc
```

Lots of FIXME/TODO are left there for now and lots of choices can be discussed, please tell me what you think!

-- Florent


example/jitter/memstruct.py
```diff
@@ -0,0 +1,229 @@
+#!/usr/bin/env python
+"""This script is just a short example of common usages for miasm2.analysis.mem.
```
commial: It seems that *memstruct* are more related to `miasm2.core`, as `DiGraph` than analysis module, such as `miasm2.analysis.depgraph`.

example/jitter/memstruct.py
```diff
@@ -0,0 +1,229 @@
...
+        # doubles
+        ("size", Num("<I")),
+    ]
+
+    def __init__(self, vm, *args, **kwargs):
+        super(LinkedList, self).__init__(vm, *args, **kwargs)
+        self.memset()
+
+    def get_head(self):
+        """Returns the head ListNode instance"""
```
commial: You choose to add a docstring here, but it is missing in others methods

example/jitter/memstruct.py
```diff
@@ -0,0 +1,229 @@
...
+memstr = datastr.deref_data
+# Note that memstr is MemStr(..., "utf16")
+memstr.value = 'Miams'
+
+print "Cast data.array to MemStr and set the string value:"
+print repr(memstr)
+print
+
+# If you followed, memstr and data.array point to the same object, so:
+raw_miams = '\x00'.join('Miams') + '\x00'*3
```
commial: This is what `miasm2.os_dep.commons.set_str_unic` do

example/jitter/memstruct.py
```diff
@@ -0,0 +1,229 @@
...
+raw_miams_array = [ord(c) for c in raw_miams]
+assert list(data.array)[:len(raw_miams_array)] == raw_miams_array
+assert data.array.cast(MemStr, "utf16") == memstr
+# Default is "ansi"
+assert data.array.cast(MemStr) != memstr
+assert data.array.cast(MemStr, "utf16").value == memstr.value
```
commial: The equality `data.array.cast(MemStr, "utf16") == memstr` does not check that `value` attributes are the same?

example/jitter/memstruct.py
```diff
@@ -0,0 +1,229 @@
...
+data.val2 = 0x12
+
+assert datastr.valshort == 0x1234
+datastr.valshort = 0x1122
+assert data.val1 == 0x22 and data.val2 == 0x11
+
+# Let's play with strings
+memstr = datastr.deref_data
+# Note that memstr is MemStr(..., "utf16")
+memstr.value = 'Miams'
```
commial: Is the typo intentional?

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
+"""This module provides classes to manipulate C structures backed by a VmMngr
+object (a miasm VM virtual memory).
```
commial: *VM* and *virtual memory* seem redundant to me

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+
+    my_heap = miasm2.os_dep.common.heap()
+    set_allocator(my_heap)
+
+Note that some structures (e.g. MemStr or MemArray) do not have a static size
+and cannot be allocated automatically.
+
+
+As you saw previously, to use this module, you just have to inherit from
+MemStruct and define a list of (<field_name>, <field_definition>). Availabe
```
commial: Typo: *Availabe*

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+    mstruct.num = 3
+    assert mstruct.num == 3
+    mstruct.other = addr2
+    mstruct.deref_other = OtherStruct(jitter.vm, addr)
+
+The `addr` argument can be omited if an allocator is set, in which case the
+structure will be automatically allocated in memory:
+
+    my_heap = miasm2.os_dep.common.heap()
+    set_allocator(my_heap)
```
commial: What is the expected interface of the allocator?

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+log.addHandler(console_handler)
+log.setLevel(logging.WARN)
+
+# ALLOCATOR is a function(vm, size) -> allocated_address
+ALLOCATOR = None
+
+# Cache for dynamically generated MemStructs
+DYN_MEM_STRUCT_CACHE = {}
+
+def set_allocator(alloc_func):
```
commial: It seems that you only use set_allocator externally, but this method doesn't provide any "improvement" (as check) compared to a `mem.ALLOCATOR = func`. Or you are keeping it for the associated docstring?

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+
+
+# Helpers
+
+def indent(s, size=4):
+    """Indent a string with @size spaces"""
+    return ' '*size + ('\n' + ' '*size).join(s.split('\n'))
+
+
+# FIXME: copied from miasm2.os_dep.common and fixed
```
commial: if this function is better than the one in `miasm2.os_dep.common`, do not hesitate to replace and import it

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+
+    Args:
+        vm: VmMngr instance
+        max_char: max number of characters to get in memory
+    """
+    l = 0
+    tmp = addr
+    while ((max_char is None or l < max_char) and
+           vm.get_mem(tmp, 1) != "\x00"):
+        tmp += 1
```
commial: If you plan to rewrite this function:

Actually, `tmp` is always equal to `addr + l`, so this variable is useless.
Additionally, there is a double memory access (one for *strlen*, the second one for getting the actual string), which may not reflect the reality, and could be inefficient.
You may rewrite it with a list joined at the end.

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        l += 1
+    return vm.get_mem(addr, l).decode("latin1")
+
+
+# TODO: get_raw_str_utf16 for length calculus
```
commial: This function can be useful in others cases, do not hesitate to move it to `os_dep.commons`

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+
+    Args:
+        vm: VmMngr instance
+        addr: start address to serialize the string to
+        s: the str to serialize
+    """
+    vm.set_mem(addr, s + "\x00")
+
+
+def set_str_utf16(vm, addr, s):
```
commial: As for `set_str_ansi`, you may move these functions to `os_dep.commons`. The use of `vm` seems to be the usual case (and more compliant with the API of `get_str_*`)

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+    s = (s + '\x00').encode('utf-16le')
+    vm.set_mem(addr, s)
+
+
+# MemField to MemStruct helper
+
+def mem(field):
+    """Generate a MemStruct subclass from a field. The field's value can
+    be accessed through self.value or self.deref_value if field is a Ptr.
+    """
```
commial: Can you specify the expected format of `@field`?

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+    def _unpack(self, raw_str):
+        """Deserializes a raw str to an object representing the python value
+        of this field.
+        """
+        raise NotImplementedError()
+
+    def set(self, vm, addr, val):
+        """Set a VmMngr memory from a value.
+
+        Args:
```
commial: To be more homogeneous with other docstring in Miasm, you can use the following format :
```Python
"""Set a VmMngr memory from a value
@vm: VmMngr instance
@addr: the start adress in memory to set
@val: the python value to serialize in @vm at @addr
"""
```

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+    def _get_self_type(self):
+        return self._self_type
+
+    def _set_self_type(self, self_type):
+        """If this field refers to MemSelf, replace it with @self_type (a
+        MemStruct subclass) when using it. Generally not used outside the lib.
+        """
+        self._self_type = self_type
+
+    def size(self):
```
commial: To be more homogenous in API, you can use a property for `size`

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+                not (isinstance(dst_type, type) and\
+                        issubclass(dst_type, MemStruct)) and\
+                not dst_type == MemSelf:
+            raise ValueError("dst_type of Ptr must be a MemStruct type, a "
+                             "MemField instance, the MemSelf marker or a class "
+                             "name.")
+        super(Ptr, self).__init__(fmt)
+        if isinstance(dst_type, MemField):
+            # Patch the field to propagate the MemSelf replacement
+            dst_type._get_self_type = lambda: self._get_self_type()
```
commial: Don't you prefer the form `dst_type._get_self_type = self._get_self_type`?

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+    def __init__(self, fmt):
+        self._fmt = fmt
+
+    def _pack(self, fields):
+        return struct.pack(self._fmt, *fields)
+
+    def _unpack(self, raw_str):
+        return struct.unpack(self._fmt, raw_str)
+
+    def size(self):
```
commial: Idem for property

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        return struct.calcsize(self._fmt)
+
+    def __repr__(self):
+        return "%s(%s)" % (self.__class__.__name__, self._fmt)
+
+    def __eq__(self, other):
+        return self.__class__ == other.__class__ and self._fmt == other._fmt
+
+    def __hash__(self):
+        return hash(hash(self.__class__) + hash(self._fmt))
```
commial: In Miasm, we usually use the form (not sure it is better):
`hash((self.__class_, self._fmt))`

example/jitter/memstruct.py
```diff
@@ -0,0 +1,229 @@
...
+class LinkedList(MemStruct):
+    fields = [
+        ("head", Ptr("<I", ListNode)),
+        ("tail", Ptr("<I", ListNode)),
+        # Num can take any one-field struct-like format, including floats and
+        # doubles
+        ("size", Num("<I")),
+    ]
+
+    def __init__(self, vm, *args, **kwargs):
```
commial: Please choose between the form `(*args, **kwargs)` and `(vm, addr, *args, **kwargs)`

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+
+    TODO: use u32, i16, etc. for format.
+    """
+
+    def _pack(self, number):
+        return super(Num, self)._pack([number])
+
+    def _unpack(self, raw_str):
+        upck = super(Num, self)._unpack(raw_str)
+        if len(upck) > 1:
```
commial: May a `!= 1` is more precise

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+
+
+class Ptr(Num):
+    """Special case of number of which value indicates the address of a
+    MemStruct. Provides deref_<field> as well as <field> when used, to set and
+    get the pointed MemStruct.
+    """
+
+    def __init__(self, fmt, dst_type, *type_args, **type_kwargs):
+        """Args:
```
commial: Idem for docstring args

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+            fmt: (str) Num compatible format that will be the Ptr representation
+                in memory
+            dst_type: (MemStruct or MemField) the MemStruct this Ptr points to.
+                If a MemField is given, it is transformed into a MemStruct with
+                mem(TheMemField).
+            *type_args, **type_kwargs: arguments to pass to the the pointed
+                MemStruct when instanciating it (e.g. for MemStr encoding or
+                MemArray field_type).
+        """
+        if not isinstance(dst_type, MemField) and\
```
commial: You can also use the form:
```
if (... and
    ... and
    ...):
```

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        Equivalent to a pointer dereference rvalue in C.
+        """
+        return self.dst_type(vm, addr, *self._type_args, **self._type_kwargs)
+
+    def deref_set(self, vm, addr, val):
+        """Serializes the @val MemStruct subclass instance in @vm (VmMngr) at
+        @addr. Equivalent to a pointer dereference assignment in C.
+        """
+        # Sanity check
+        if self.dst_type != val.__class__:
```
commial: So if the val class is subclassed, you will always raise a warning. Is this an expected behavior, rather than a `isinstance`?

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        return "%s(%r)" % (self.__class__.__name__, self._dst_type)
+
+    def __eq__(self, other):
+        return super(Ptr, self).__eq__(other) and \
+                self.dst_type == other.dst_type and \
+                self._type_args == other._type_args and \
+                self._type_kwargs == other._type_kwargs
+
+    def __hash__(self):
+        return hash(super(Ptr, self).__hash__() + hash(self._dst_type) +
```
commial: Idem for `hash(tuple)`

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        self._type_kwargs = type_kwargs
+
+    def set(self, vm, addr, val):
+        raw = str(val)
+        vm.set_mem(addr, raw)
+
+    def get(self, vm, addr):
+        return self._il_type(vm, addr)
+
+    def size(self):
```
commial: Idem property

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        return "%s(%r)" % (self.__class__.__name__, self._il_type)
+
+    def __eq__(self, other):
+        return self.__class__ == other.__class__ and \
+                self._il_type == other._il_type and \
+                self._type_args == other._type_args and \
+                self._type_kwargs == other._type_kwargs
+
+    def __hash__(self):
+        return hash(hash(self.__class__) + hash(self._il_type) +
```
commial: Idem hash

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        elif isinstance(val, list):
+            if len(val) != self.array_len:
+                raise ValueError("Size mismatch in MemSizedArray assignment ")
+            offset = 0
+            for elt in val:
+                self.field_type.set(vm, addr + offset, elt)
+                offset += self.field_type.size()
+
+        else:
+            raise NotImplementedError(
```
commial: Even if it is not the case in the rest of Miasm code, `NotImplementedError` should be reserved for abstract method.
When a method is not implemented, you can raise a `RuntimeError`, for instance.

[Source](https://docs.python.org/2/library/exceptions.html):
```
 exception NotImplementedError

    This exception is derived from RuntimeError. In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method.
```
We should probably create an exception in Miasm for this kind of case (`FdsException`? :smiley:)

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+                     )]
+
+        ex = Example(vm, addr)
+        ex.f2 = 0x1234
+        assert ex.f1 == 0x34
+        assert ex.uni == '\x34\x12'
+        assert ex.get_addr("f1") == ex.get_addr("f2")
+    """
+
+    def __init__(self, field_list):
```
commial: What do you think about using a `*field_list` instead, avoiding the creation of a list for the caller?

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        """field_list is a [(name, field)] list, see the class doc"""
+        self.field_list = field_list
```
commial: In the module, you could manage field_list as an `OrderedDict`, giving a direct access to `.keys, .values, ...` and a test for unicity of keys.

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        res_val = (num_val & num_mask) | val_shifted
+        self._num.set(vm, addr, res_val)
+
+    def get(self, vm, addr):
+        val_mask = (1 << self._bits) - 1
+        num_val = self._num.get(vm, addr)
+        res_val = (num_val >> self._bit_offset) & val_mask
+        return res_val
+
+    def size(self):
```
commial: Idem property

miasm2/os_dep/common.py
```diff
@@ -60,9 +60,15 @@ def alloc(self, jitter, size):
         @jitter: a jitter instance
         @size: the size to allocate
         """
+        return self.vm_alloc(jitter.vm, size)
 
+    def vm_alloc(self, vm, size):
```
commial: You may add the associated page rights as an argument, with default value `PAGE_READ | PAGE_WRITE`

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+    __metaclass__ = _MetaMemStruct
+
+    fields = []
+
+    _size = None
+
+    # Classic usage methods
+
+    def __init__(self, vm, addr=None, *args, **kwargs):
+        global ALLOCATOR
```
commial: The ALLOCATOR could be a class attribute, rewritable with an instance attribute. That way, different allocators can be used in the same run (for instance, with multiple heap in Windows).

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+                                 "%s.set_allocator has not been called."
+                                 % __name__)
+            self._addr = ALLOCATOR(vm, self.get_size())
+        else:
+            self._addr = addr
+
+    def get_addr(self, field_name=None):
+        """Return the address of this MemStruct or one of its fields.
+
+        Args:
```
commial: Idem docstring

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+
+    @classmethod
+    def sizeof(cls):
+        """Return the static size of this structure, when available (it is the
+        case by default).
+        """
+        # Child classes can set cls._size if their size is not the sum of
+        # their fields
+        if cls._size is None:
+            return sum(a["field"].size() for a in cls._attrs.itervalues())
```
commial: You may set `cls._size`, giving a lazy `sizeof` (it seems that this size is constant)

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        instance has a fixed size (because it has a fixed length and
+        field_type), but all the instance do not have the same size.
+        """
+        return self.sizeof()
+
+    def get_field_type(self, name):
+        """Return the MemField subclass instance describing field @name."""
+        return self._attrs[name]['field']
+
+    def get_field(self, name):
```
commial: Idem with OrderedDict facilities

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        Useless most of the time since fields are accessible via
+        self.deref_<name>.
+        """
+        addr = self.get_field(name)
+        field = self._attrs[name]["field"]
+        assert isinstance(field, Ptr),\
+               "Programming error: field should be a Ptr"
+        field.deref_set(self._vm, addr, val)
+
+    def memset(self, byte='\x00'):
```
commial: We could imagine to fill memory with a pattern, longer than 1 byte (for instance, think to Metasploit pattern_create.rb)

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        offset = 0
+        for name, field in cls.fields:
+            # For reflexion
+            field._set_self_type(cls)
+            cls.gen_field(name, field, offset)
+            offset += field.size()
+        cls._size = offset
+
+    @classmethod
+    def gen_field(cls, name, field, offset):
```
commial: Idem docstring

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+    The @encoding is passed to the constructor, and is currently either null
+    terminated "ansi" (latin1) or (double) null terminated "utf16". Be aware
+    that the utf16 implementation is a bit buggy...
+
+    The string value can be got or set (with python str/unicode) through the
+    self.value attribute. String encoding/decoding is handled by the class.
+
+    This type is dynamically sized only (get_size is implemented, not sizeof).
+    """
+    def __init__(self, vm, addr, encoding="ansi"):
```
commial: You can rewrite encoding elements as a class attribute like:
```Python
encodings = {"ansi": {"getter": get_str_ansi,
                                  "setter": set_str_ansi,
                                 },
                       ...
}
```

It could avoid hardcoding `"ansi"` and `"utf16"` in some function, and it may give a way to easily add a custom encoding "on the fly"

commial: Or, maybe, you can use subclasses for each encoding and a registration in the main class (`MemStr`).
This is the pattern used in `Container` and `Translators`, which seems to be cleaner than a list of conditions in each method. In other words, you are doing the split inside methods, but it can be done by the class.

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+
+    def _normalize_slice(self, slice_):
+        start = slice_.start if slice_.start is not None else 0
+        stop = slice_.stop if slice_.stop is not None else self.get_size()
+        step = slice_.step if slice_.step is not None else 1
+        return slice(start, stop, step)
+
+    def _check_bounds(self, idx):
+        idx = self._normalize_idx(idx)
+        if not isinstance(idx, int) and not isinstance(idx, long):
```
commial: In Python, you can use `isinstance(idx, (int, long))`, or, for this specific case, `isinstance(idx, numbers.Number)` which also work for `float` and `Decimal`

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+            raise ValueError("index must be an int or a long")
+        if idx < 0:
+            raise IndexError("Index %s out of bounds" % idx)
+
+    def index2addr(self, idx):
+        """Return the address corresponding to a given @index in this MemArray.
+        """
+        self._check_bounds(idx)
+        addr = self.get_addr() + idx * self._field_type.size()
```
commial: `addr` variable is useless here

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        to.
+        """
+        return self._field_type.deref_get(self._vm, self[idx])
+
+    def __setitem__(self, idx, item):
+        if isinstance(idx, slice):
+            idx = self._normalize_slice(idx)
+            if len(item) != len(xrange(idx.start, idx.stop, idx.step)):
+                raise ValueError("Mismatched lengths in slice assignment")
+            for i, val in zip(xrange(idx.start, idx.stop, idx.step), item):
```
commial: If you want to fully use iterators, you can use `itertools.izip` instead of `zip`

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        """
+        self._field_type.deref_set(self._vm, self[idx], item)
+
+    # just a shorthand
+    def as_mem_str(self, encoding="ansi"):
+        return self.cast(MemStr, encoding)
+
+    @classmethod
+    def sizeof(cls):
+        raise ValueError("%s is unsized, which makes some operations"
```
commial: The `%s` is never matched

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+                         " impossible. Use MemSizedArray instead.")
+
+    def raw(self):
```
commial: Idem `sizeof`

You can avoid a duplicate of code by calling `self.sizeof`

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+    def __iter__(self):
+        for i in xrange(self._array_len):
+            yield self[i]
+
+    def raw(self):
+        return self._vm.get_mem(self.get_addr(), self.get_size())
+
+    def __repr__(self):
+        item_reprs = [repr(item) for item in self]
+        if self.array_len > 0 and '\n' in item_reprs[0]:
```
commial: As you are in the class definition, you can use `self._array_len`

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+                self._array_len == other._array_len and \
+                str(self) == str(other)
+
+
+def mem_array_type(field_type):
+    """Generate a MemArray subclass that has a fixed @field_type. It allows to
+    instanciate this class with only vm and addr argument, as are standard
+    MemStructs.
+    """
+    cache_key = (field_type, None)
```
commial: This pattern seems to be the equivalent of `.setdefault`

miasm2/analysis/mem.py
```diff
@@ -0,0 +1,1259 @@
...
+        return DYN_MEM_STRUCT_CACHE[cache_key]
+
+    array_type = type('MemArray_%r' % (field_type,),
+                      (MemArray,),
+                      {'_field_type': field_type})
+    DYN_MEM_STRUCT_CACHE[cache_key] = array_type
+    return array_type
+
+
+def mem_sized_array_type(field_type, array_len):
```
commial: This pattern seems to be the equivalent of `.setdefault`

test/analysis/mem.py
```diff
@@ -0,0 +1,493 @@
...
+assert jitter.vm.get_mem(memarray.get_addr(),
+                         Num("I").size()) == '\x02\x00\x00\x00'
+memarray[2] = 0xbbbbbbbb
+assert memarray[2] == 0xbbbbbbbb
+assert jitter.vm.get_mem(memarray.get_addr() + 2 * Num("I").size(),
+                         Num("I").size()) == '\xbb\xbb\xbb\xbb'
+try:
+    s = str(memarray)
+    assert False, "Should raise"
+except (NotImplementedError, ValueError):
```
commial: Useless parenthesis

test/test_all.py
```diff
@@ -154,6 +154,8 @@ def __init__(self, arch, container, address, *args, **kwargs):
                                                         (14, 1), (15, 1)))
                            for fname in fnames])
 
+testset += RegressionTest(["mem.py"], base_dir="analysis")
```
commial: The example script is not tested

example/jitter/memstruct.py
fmonjalet: I agree, if it is ok with @serpilliere too, I will move it.

example/jitter/memstruct.py
fmonjalet: Will be fixed

example/jitter/memstruct.py
fmonjalet: That's right. Do you want me to add a docstring to the other methods, or just drop this one since it seems quite obvious?

example/jitter/memstruct.py
fmonjalet: Of course :). Do you want me to fix it, for trademark issues?

example/jitter/memstruct.py
fmonjalet: Yes, but the point is to show the actual bytes that are written to memory, I think it is clearer for an example.

example/jitter/memstruct.py
fmonjalet: No, these are two different strings in memory, corresponding to the same python str value, but with two different concrete encoding. That's why the value attribute is not used for equality. And even if it was, the encoding should also be used,and these two instances would not be eq.

However, the current implementation tests `str(self) == str(other)`, so two different strings with different values and different encoding could currently be eq, if they have the same `str()`. This may not be the expected behaviour.

So the question is, should two `MemStr` be equal when they have the same memory representation AND the same encoding (i.e. "lala" in utf8 != "lala" in latin1)or only when they have the same memory representation? (I would go for the first)

miasm2/analysis/mem.py
fmonjalet: VM stands for Virtual Machine here, I will change it to sandbox to avoid confusion, thanks.

miasm2/analysis/mem.py
fmonjalet: Will be fixed, thanks.

miasm2/analysis/mem.py
fmonjalet: I'm keeping it for two reasons:

- abstraction: the ability to hide the mechanism if it changes + possible checks, as you mentioned;
- Ease of import: the user will probably want to import classes from the `mem` module, but it seems that you have to import the module in itself to set a global (or did I miss something?). Importing the function is easier.

miasm2/analysis/mem.py
fmonjalet: Thanks, what would you think to remove the function from `miasm.os_dep.common` import it from `miasm.core.mem` instead?
Since this module is aimed at providing utilities to interract with the virtual memory, I think these functions may belong to it. Moreover, str encoding is not _directly_ related to the underlying OS.

miasm2/analysis/mem.py
fmonjalet: It seems to be a good time to apply this fix, I will do it in this PR.

miasm2/analysis/mem.py
fmonjalet: Yes, missing from the doc, sorry (it is a MemField instance)

miasm2/analysis/mem.py
fmonjalet: I'll fix it, thanks.

miasm2/analysis/mem.py
fmonjalet: I think it is functionally equivalent, but your version is cleaner + shorter, I'll fix that.

miasm2/analysis/mem.py
fmonjalet: Right.

miasm2/analysis/mem.py
fmonjalet: Will be fixed

miasm2/analysis/mem.py
fmonjalet: The two versions are actually not equivalent:
* `dst_type._get_self_type = lambda: self._get_self_type()` will make `dst_type` always call the current `_get_self_type` of `self` _at the time of the call_.
* `dst_type._get_self_type = self._get_self_type` will make `dst_type` call the `_get_self_type` function of `self` that was set _at the moment of the method assignment_. Note that `self._get_self_type` will later be patched by `MemStruct.gen_fields`.

Since inner types are created before outer types (e.g. `Ptr(_, MemSelf)` before `Ptr(_, Ptr(_, MemSelf))`), the `_get_self_type` of the outer is not patched at the moment it patches the inner class. This is why delayed evaluation of `_get_self_type` is necessary. This also is why this is a function and not an attribute.

To be honest, `MemSelf` implementation is a hacky mess; I may have been tired or confused the day I imagined that, and it stayed stuck to my mind. :)
I would really be interested in cleaner suggestion to implement that.

miasm2/analysis/mem.py
fmonjalet: This kind of thing, in a C or C++ compiler, would raise a warning or an error. The idea is that if you subclass you `MemStruct` to add fields, an assignment of the subclass to the parent _should_ raise a warning. If you subclass it just to add methods, it's okay, but I do not see real usecases for that.

In both cases, casting your subclass to the superclass (via the `.cast(<superclass>)`) method, as you would do in C, is the way to go if you want to get rid of the warning. This is the way to indicate that you know what you are doing. Is this explanation ok for you?

miasm2/analysis/mem.py
fmonjalet: Thanks, fixed.

miasm2/analysis/mem.py
fmonjalet: It could be handy, but I think it would make things non homogeneous, or at least less clear, for subclasses like `BitField`. What do you think?

miasm2/analysis/mem.py
fmonjalet: Actually, in an `Union`, the fields are not ordered. The doc talks about a list of field, but in fact any iterable is fine (I may update it in that sense).

Maybe it could be useful for the `MemStruct._attrs` attribute, but I am not convinced it is worth it yet.

miasm2/analysis/mem.py
fmonjalet: Good remark! I will update the code in that way.

miasm2/analysis/mem.py
fmonjalet: I thought about that, but it would break the `sizeof` function if fields are modified after the first `sizeof` call and `gen_fields` is called.

miasm2/analysis/mem.py
fmonjalet: Well... I was lazy on this one. :)
I will add the functionality.

miasm2/analysis/mem.py
fmonjalet: Thanks for the ideas, I will think about it. In any way, this has to be done in a smarter way than it is done now. I think I prefer the first idea.

miasm2/analysis/mem.py
fmonjalet: Thanks!

miasm2/analysis/mem.py
fmonjalet: True, but for clarity and debug I prefer to have short `return` statements.

miasm2/analysis/mem.py
fmonjalet: I wasn't aware of this one, thanks.

miasm2/analysis/mem.py
fmonjalet: Fixed by putting more precise error messages.

miasm2/analysis/mem.py
fmonjalet: I may be wrong, but I think the trap with using `.setdefault` here is to mask the original class name with a newly instanciated one (the `default` of `setdefault`) and returning the class that was in the cache. I think it would work, but I do not really like the namespace pollution it may produce (though I may already have name collisions in my generated classes...).

So what do you think, `setdefault` or not?

miasm2/os_dep/common.py
fmonjalet: Good idea, done.

test/analysis/mem.py
fmonjalet: Seems actually necessary here (not for `(ValueError)`, of course).

test/test_all.py
fmonjalet: Right. Before I add it, do you think `example/jitter/memstruct.py` is a correct path for it? I can't figure out where it would fit the best.

fmonjalet:
Hi @commial, thanks for the time you dedicated to this review, I'm working on the fixes!

fmonjalet:
Hi,

After a long discussion with @a-vincent and @serpilliere, I ended up with a _big_ refactoring of the module to mark a clear separation between `Type`s, which represent C (or higher level) types, and `MemType`s which represent a value of a given `Type` in memory; what would be called an LValue in a compiler. This distinction allows to shape this module so that it can serve as a base for type analysis.

In summary:
- A `Type` instance is a completely defined type (i.e. `char[3]*`)
- A `MemType` subclass instance is a kind of memory cell that allows to interact with a value of a given `Type`, in a given `vm`, at a given `addr` (i.e. `char[3]* var`)
- Instanciating a `Type` and accessing its `.lval` property returns an associated `MemType` subclass. Its constructor always has two arguments: `vm` and `addr`:
```python
argv_t = Array(Ptr("<I", Str()), 4)
my_argv = argv_t.lval(vm, some_addr)
```
- The `.lval` attribute returns a dynamically generated subclass of a given type (e.g. `Str().lval` is a `MemStr` subclass). These subclasses are singletons, if two types are equal, they will return the same subclass. This allows the client to manipulate the subclass, instanciate it as well as calling class methods. As ugly as this mechanism can seem, I still think it is the one that makes for an easier client API, but I am of course open to suggestions.
- The client should never have to import `Mem*` classes, except for `MemStruct`, that is used for user friendly structure declaration. That said, the API makes the two following declarations almost equivalent:
```python
s = Struct("Toto", [("f1", Num("I")), ("f2", Num("I"))])
Toto1 = s.lval
```
  Equivalent to:
```python
class Toto(MemStruct):
    fields = [("f1", Num("I")), ("f2", Num("I"))]
```
  If the two are present in the same file (with the same name and fields):
  - If the class is declarated first, the `Struct(...).lval`  will return the previously declared class (singleton pattern).
  - If the `Struct(...).lval` is used before the class declaration, the class declaration will raise an exception, saying that a `MemType` subclass of the same type (fields and name) has already been created. This avoids violating the `MemType` subclass singleton pattern (one class per `Type` instance). I believe this is not a blocking constraint for most codes.

What do you think?

example/jitter/types.py
```diff
@@ -0,0 +1,258 @@
...
+# a program:
+# Let's say that we have two arguments, +1 for the program name and +1 for the
+# final null ptr in argv, the array has 4 elements:
+argv_t = Array(Ptr("<I", Str()), 4)
+print "3 arguments argv type:", argv_t
+
+# alloc argv somewhere
+argv = argv_t.lval(vm)
+
+# Auto alloc with a buffer type
```
fmonjalet: Note to self: misleading comment, allocating a buffer is no longer necessary since `.from_str` has been added to the API. To be fixed.

fmonjalet:
Ready to merge for me, as soon as tests pass.

serpilliere:
Hey, Santa Claus brought another heavy gift! :santa: 
Thanks for the feature: It will be very useful to rewrite cleaner Windows structures!

PR #295, opened 2015-12-16
Update api according to Elfesteem

serpilliere:
:warning: This PR needs ELFESTEEM serpilliere/elfesteem@837b05affa09b2095877ad6801be82c2db557d45 update!!

### DIFF
--- a/example/asm/shellcode.py
+++ b/example/asm/shellcode.py
@@ -96,8 +96,13 @@ if args.encrypt:
     patches = new_patches
 
 print patches
-for offset, raw in patches.items():
-    virt[offset] = raw
+if isinstance(virt, StrPatchwork):
+    for offset, raw in patches.items():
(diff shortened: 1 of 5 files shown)


example/asm/shellcode.py
```diff
@@ -96,8 +96,13 @@
     patches = new_patches
 
 print patches
-for offset, raw in patches.items():
-    virt[offset] = raw
+if isinstance(virt, StrPatchwork):
```
commial: Do you plan to update `StrPatchwork` API too?
By adding a `.set()` method, in order to be independent of the sublayer (`StrPatchwork`, elfesteem `virt` wrapper, ...)

serpilliere: Yes, @fmonjalet  saw that as well, but I will not add these feature for the moment. I am 'testing' this new API for now

PR #298, opened 2016-01-05
SMT2 translator

mrphrazer:
This PR introduces a translator for the SMT2 standard. With this, SMT formulas can be solved independently from Z3 by any SMT solver that supports the standard.

### DIFF
--- a/miasm2/expression/smt2_helper.py
+++ b/miasm2/expression/smt2_helper.py
@@ -23,11 +23,22 @@ def smt2_implies(a, b):
     return "(=> {} {})".format(a, b)
 
 
-def smt2_and(a, b):
+def smt2_and(*args):
     """
-    Conjunction: a and b
+    Conjunction: a and b and c ...
     """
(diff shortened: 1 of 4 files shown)


serpilliere:
Hey, it seems Santa Claus did a great job during the holidays :smiley:

mrphrazer:
Thanks :)

I ran a few more tests and fixed one bug in ExprId and refactored minor things.

serpilliere:
Ok, just a little bug: @commial raised the fact that:
```python
testset += RegressionTest(["z3_ir.py", "smt2.py"], base_dir="ir/translators",
```

will not run the `smt2.py`: it will pass it as argument for `z3_ir.py`
(you can check this by insering a bad test, and test_all won't see it)

You need to inser an independant test line!

mrphrazer:
thanks, you are right! fixed.

serpilliere:
Perfect.

PR #303, opened 2016-01-15
FIXE : newer tinycc need this for a successful build

Maroc-OS:
you have to apply the fix or update the readme file with the exact version and commit for desirable tinycc

### DIFF
--- a/miasm2/jitter/Jittcc.c
+++ b/miasm2/jitter/Jittcc.c
@@ -47,7 +47,7 @@ TCCState * tcc_init_state(void)
 
 	//tcc_add_file(tcc_state, libcodenat_path);
 	for (i=0;i<lib_array_count; i++){
-		tcc_add_file(tcc_state, lib_array[i]);
+		tcc_add_file(tcc_state, lib_array[i], TCC_FILETYPE_C);
 	}
 
 	for (i=0;i<include_array_count; i++){


serpilliere:
Hi!
I think you are right: the tinycc version is not well explained. To be clear, we only support the latest stable revision of tinycc (to minimize dependencies overhead), which is actually 0.9.26 (see `http://repo.or.cz/w/tinycc.git/tags`). So I won't merge your PR because it modifies API to support the latest unstable tinycc (that's why it breaks regressions tests), but I will change the README to explicitly clone 0.9.26 tinycc.

Something like:
* Clone latest stable tinycc version:   `git clone http://repo.or.cz/tinycc.git release_0_9_26`

Does it seem legit to you?

serpilliere:
Here is the future README (with tinycc version modification):

...
Miasm uses:

* LibTCC [tinycc (ONLY version 0.9.26)](http://repo.or.cz/w/tinycc.git), to JIT code for emulation mode
* or LLVM v3.2 with python-llvm, see below
* python-pyparsing
* python-dev
* elfesteem from [Elfesteem](https://github.com/serpilliere/elfesteem.git)

Configuration
-------------

* Install elfesteem
```
git clone https://github.com/serpilliere/elfesteem.git elfesteem
cd elfesteem
python setup.py build
sudo python setup.py install
```
* To use the jitter, TCC or LLVM is recommended
* LibTCC needs to be configured with the `--disable-static` option
  * remove `libtcc-dev` from the system to avoid conflicts
  * clone [TinyCC](http://repo.or.cz/tinycc.git): `git clone http://repo.or.cz/tinycc.git`
  * set branch to version 0.9.26: `cd tinycc/` and `git checkout release_0_9_26`
  * `./configure --disable-static && make && make install`
* LLVM
  * Debian (testing/unstable): install python-llvm
  * Debian stable/Ubuntu/Kali/whatever: install from [llvmpy](http://www.llvmpy.org/)
  * Windows: python-llvm is not supported :/
* Build and install Miasm:
```
$ cd miasm_directory
$ python setup.py build
$ sudo python setup.py install
```
...

Is it clearer?

commial:
You should replace "make && make install", as " make" will likely throws an error due to documentation generation on basic Linux install (and we don't care about this specific error).

Maybe splitted on several lines ?

Maroc-OS:
hello :)

Nice to see you reacting to this, okay for the PR it is okay as i said on the description to make it clear for people who want to use miasm.

>Clone latest stable tinycc version: git clone http://repo.or.cz/tinycc.git and git checkout release_0_9_26

#304 yes this is one of the two.

>You should replace "make && make install", as " make" will likely throws an error due to documentation generation on basic Linux install (and we don't care about this specific error).

Exactly and make install should be executed as root, and also when you run it first time it does not work :) they have a problem with this version which is calling *tcc* before it gets compiled, error is : ```make[1]: ../tcc: command not found```.

So you have to call make again.

>@Maroc-OS, if it seems OK for you, I'll close this one and merge the other one.

Yep same as i said before.

-----

For the check errors that are mostly something like ```NameError: global name 'JitCore_Tcc' is not defined```

It is because linker does not work properly and you can see that when trying to import 'JitCore_Tcc'

```from miasm2.jitter.jitcore_tcc import JitCore_Tcc```

you will see the error there that say's tcc_relocate is not defined

```ImportError: /usr/local/lib/python2.7/dist-packages/miasm2/jitter/Jittcc.so: undefined symbol: tcc_relocate```

PR #309, opened 2016-01-25
Feature AsmCFG

commial:
Introduce `AsmCFG`, a class standing for an ASM Control Flow Graph (CFG). This class aims to replace the old representation of ASM blocks as a list of `asm_bloc` instance.

Indeed, the representation as a `DiGraph` seems to be closer to what one can expect as an output of a disassembler. In addition, algorithm applied on blocks list, such as block splitting, bad destination retrieving, block graphing and sanity checking for assembler are working on a graph (but were using `block.bto`s to build a kind of intermediate graph construction, ending in more complex code snippets). As these algorithms were built to work on a disassembler outputs and are strongly connected to it, they have been rewritten as `AsmCFG` method.

In order to be much easier to use, edges are binded on blocks constraints, ie. adding or deleting an edge will impact implicated `.bto`s. The reverse operation (re-synchronization between `.bto`s) and edges can be done thanks to a `rebuild_edges` helper.

A regression test for `miasm2.core.asmbloc` has also been added (it should be enough commented to illustrates some uses and associated APIs).
As an example, with the new API:
```Python
# Get a AsmCFG instance from a function
blocks = mdis.dis_multibloc(addr).
# Output it for visualization
open("cfg.dot", "w").write(blocks.dot())
# Find reachable blocks from the graph root(s)
for head in blocks.heads():
    for asm_block in blocks.reachable_sons(head):
        ...
# Get a specific assembly block from its label (lazy built)
a_block = blocks.label2block(my_label)
# Remove a block and its associated edges, updating its predecessors .bto
blocks.delete(a_block)
```

To prevent hours of debugging, a few `DeprecationWarning` with hints have been added on old APIs.

As they are related, this PR also introduces `DiGraphSimplifier`, which aims to be the counterpart of `ExpressionSimplifier` for `DiGraph`.
One can register graph simplification, which will be applied on a graph (actually, on a copy of this graph) until a fixed point has been reached.
The exposed API is close to the one of `ExpressionSimplfier`, ie using `enable_passes(list of simplifications)` to enable simplification passes, then applying the simplifier thanks to `simp.apply(graph)` or `simp(graph)` to get back a new, simplified, graph.

Its use is illustrated in `miasm2.core.asmbloc: bbl_simplifier` which ship a block merging pass (former `bloc_merge`). This implementation gets rid of some piece of code (see 388212934c for more details), and seems, in my opinion, easier to understand.


example/disasm/full.py
```diff
@@ -142,15 +142,13 @@
 
 
 # Generate dotty graph
-all_blocs = []
+all_blocs = BasicBlocks()
```
mrphrazer: I don't know if this is  intended, but I noticed that you switch to "blocks", in general. In some parts of this PR you leave it as "bloc". Especially in the code above and below, down to asmbloc.py.

commial: Actually, "bloc" is the French word for "block". This is an error which is repeated all along Miasm, even in main structure (`asm_bloc`) or module name (`miasm2.core.asmbloc`).
Instead of breaking a lot of APIs, I prefer correct them inside function when I modify code around it. But we may correct it once and for all, a day.

So, following this reasoning, I will fix this one :smile:

commial: After a quick review, there are too many lines to change regarding this correction, lines which are not correlated with the logic of the concerned commit.
Then, I prefer to make a separate commit for all of them, and, a separate PR for the same reasons.

Thanks for your review!

PR #321, opened 2016-02-07
Use the right call convention for x86-64 under Linux...

aguinet:


commial:
First, thanks for your PR.

Actually, the calling convention you've replaced was the one used on Windows.
If you look at `miasm2.os_dep.win_api_x86_32`, you'll see that Windows API is already using this calling convention, so it will break these functions.

For me, the proper way to patch it is to differentiate Linux and Windows ABI simulation, and it's may be the time to refactor the whole ABI system (in a separate module, to re-use it in [ira](https://github.com/cea-sec/miasm/blob/master/miasm2/arch/x86/ira.py), may be with a merge with the one of [Sibyl](https://github.com/cea-sec/Sibyl/tree/master/sibyl/abi)).

A quicker patch, before refactoring, could be a change in `miasm2/os_dep/linux_stdlib.py` by replacing `func_args_stdcall` by `func_args_stdcall_linux` and adding it in concerned architectures (`miasm2/arch/x86/jit.py`, ...).

serpilliere:
I am closing this PR and open an issue #355 to track the feature request

PR #347, opened 2016-03-23
Dg has loop

serpilliere:
Add loop detection algorithm to the graph class.
Update `DependencyResult` to use this algorithm.

Thanks to @carolineLe for the loop detection algorithm

### DIFF
--- a/miasm2/analysis/depgraph.py
+++ b/miasm2/analysis/depgraph.py
@@ -252,10 +252,10 @@ class DependencyResult(DependencyState):
 
     @property
     def has_loop(self):
-        """True if current dictionary has a loop"""
+        """True iff there is at least one data dependencies cycle (regarding
+        the associated depgraph)"""
         if self._has_loop is None:
-            self._has_loop = (len(self.relevant_labels) !=
-                              len(set(self.relevant_labels)))
+            self._has_loop = self.graph.has_loop()
         return self._has_loop
 
     def irblock_slice(self, irb):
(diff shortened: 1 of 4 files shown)


mrphrazer:
just a quick reminer, if necessary, I will add more details later. The DiGraph class already has to loop algorithms for loop detection: `compute_natural loops` and `compute_strongly_connected_component`. the second one finds all loops in a graph: every strongly connected component with more than one node is a loop. however, this will not detect self-loops. but, a self-loop is a natural loop. therefore, all loops in the graph can be detected by combining the two algorithms. 

If needed, I will add more details tomorrow.

commial:
The goal of `has_loop` is to stop and return if at least one cycle has been detected, without any information on that loop.

Correct me if I'm wrong, but a DFS seems to be the fastest way to obtain such a result, without involving complex algorithm such as `compute_strongly_connected_components` and natural loop detection.
By the way, I think that a quicker solution for self loops hunting is a linear traversal of `graph._nodes_succ` looking for a value which is a set of only the key.

mrphrazer:
Hi, I was not sure if that approach would also detect irreducible loops. I had some time for re-thinking;  you are right, this approach will detect both kind of loops. 

Thanks for that :)

serpilliere:
Ok.

So do you agree with us to merge the PR?
It's clear, we could design an algorithm built on top of `compute_strongly_connected_components` to detect all loops, but what do you think of keeping this current `has_loop` for this usecase, *and* keeping `compute_strongly_connected_components` for, let say, advanced future needs?  :wink:

mrphrazer:
Sure! I first wasn't sure about the soundness (irreducible loops). For this use case, it is faster than my approach mentioned above.  Thanks for asking :)

In addition, you are right. Both should stay inside since they have different use cases. For instance, strongly connected components yields outermost loops, which is useful for loop unrolling.

PR #352, opened 2016-04-17
Static single assignment (SSA)

mrphrazer:
This PR introduces static single assignment (SSA). It is joint work with [Niko Schmidt](https://github.com/itsacoderepo). It implements a
* generic class for SSA transformations
* class for transformations on block level
* class for transformations on control flow graph level

We start with

```python
# initialise IRA
ira = m.ira(mdis.symbol_pool)
# initialise IRA SSA
ira_ssa = m.ira(mdis.symbol_pool)
```

Then, we transform

```python
# a single block
ssa = SSABlock(ira)
ssa.transform(block_label)

# a path
path = ira.graph.find_path(start, end)[0]
ssa = SSAPath(ira)
ssa.transform(path)

# an entire CFG
head = ira.get_bloc(start_addr).label
ssa = SSADiGraph(ira)
ssa.transform(head)
```

The transformed expressions are located in the dict `ssa.expressions`. In addition, copies of the IRA blocks in SSA form are in `ssa.blocks`.

We can view the transformed graph with

```python
# update IRA SSA
ira_ssa.blocs.update(ssa.blocks)

print ira_ssa.graph.dot()
```

The classes `SSABlock` and `SSAPath` allow the reassembling of expressions, which is useful for static slicing:

```python
e = ExprId("IRDst.55", 64)
print ssa.reassemble_expr(e)
```

The SSA transformation is prone to memory aliasing. In future, an SSA form as [memory SSA](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.33.6974&rep=rep1&type=pdf) may be considered.

The SSA form on DiGraph level is known as [minimal SSA](http://pages.cs.wisc.edu/~fischer/cs701.f03/ssa.pdf).

Up until now, regression tests for SSA are missing. Especially in the case of `SSADiGraph`, we are open for comments as useful tests could look like.


serpilliere:
Hi!

That's a really great feature :smile: 
I have successfully tested it by modifying the `full.py` script.
We will review the code with @commial in the next days (we may have some questions)
Have you got some plans in a near future based on this feature?


Thanks again to you, @mrphrazer and @itsacoderepo  for the PR!

mrphrazer:
Hi,

great! Code review and questions are welcome :)

We actively use SSA for static backward slicing. In addition, it is an integral part for our SMT-based program analysis. In general, we hope that SSA is a good foundation to push forward the data flow analysis techniques in Miasm.

miasm2/expression/expression_dissector.py
```diff
@@ -0,0 +1,168 @@
...
+        elif isinstance(e, m2_expr.ExprInt):
+            iterator = iter([])
+        elif isinstance(e, m2_expr.ExprId):
+            iterator = iter([])
+        else:
+            raise NotImplementedError()
+
+        return iterator
+
+    def _check_itetator(self, iterator, e):
```
commial: As this method is used only once, you can in-line it in `_dissect`.
Is `iteTator` a typo?

According to [Python doc](https://docs.python.org/2/library/exceptions.html#exceptions.NotImplementedError), `NotImplementedError` is reserved to indicate abstract methods in abstract classes (a bit misleading). You can raise, for instance, a `RuntimeError` instead.

miasm2/expression/expression_dissector.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        Checks if an iterator is defined
+        :param iterator: iterator
+        :param e: expression
+        """
+        if not iterator:
+            raise NotImplementedError("no handler for {}".format((type(e))))
+
+    def _dissect(self, e, expr_type):
```
commial: When it is possible, please use at least two or three character for variable name, it is easier to read / search / avoid a mess when the function is more complex.

miasm2/expression/expression_dissector.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def op(self, e):
+        """
+        Dissects an expression into ExprOps
+        :param e: expression
+        :return: set of ExprOp in e
+        """
+        return self._dissect(e, m2_expr.ExprOp)
+
+    def slice(self, e):
```
commial: `slice`, `int` and `id` are keywords in Python. It is better to avoid using them as names.
There is a common workaround: add a `_` after them.

miasm2/expression/expression_dissector.py
```diff
@@ -0,0 +1,168 @@
+import miasm2.expression.expression as m2_expr
+
+
+class ExprDissector:
```
commial: When it is possible, please inherit from `object` (to avoid old style classes).

miasm2/expression/expression_dissector.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def __init__(self, registers=None, irdst=None):
+        """
+        Initialises the architecture variables
+        :param registers: registers of an architecture
+        :param irdst: IRA instruction pointer of an architecture
+        """
+        # set architecture variables
+        if registers and irdst:
+            self.arch_vars = set(registers + [irdst])
```
commial: I'm not a big fan of attributes which are sometimes defined, sometimes not.
You can always defined it, with `set()` when no registers are provided.

Then, instead of raising an `AssertionError`, you can raise an error more specific (or a silent pass).

miasm2/ir/ssa.py
```diff
@@ -0,0 +1,279 @@
...
+        regs = self.ira.arch.regs.all_regs_ids
+        irdst = self.ira.IRDst
+        # class for parsing expressions
+        self._expr_dissect = ExprDissector(regs, irdst)
+
+    def transform(self, *args, **kwargs):
+        """Transforms into SSA"""
+        raise NotImplementedError("")
+
+    def get_block(self, block):
```
commial: It could be misleading to use `block` for an `asm_label` instance. In others code snippets, we preferably use `label`

miasm2/ir/ssa.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def _gen_var_expr(self, v, stack):
+        """
+        Generates a variable expression in SSA form
+        :param v: variable expression which will be translated
+        :param stack: self._stack_rhs or self._stack_lhs
+        :return: variable expression in SSA form
+        """
+        index = stack[v]
+        name = v.name + "." + str(index)
```
commial: Instead of using `.` as a magic separator, you can use your own class for the `name`. Indeed, this is not required that `name` is a string. Then, this class can easily handle the number associated to the var, and a `str` displaying it as the current output.
In other architecture, we may want to represent some register using a `.`, and it may end with a confusion.

miasm2/ir/ssa.py
```diff
@@ -0,0 +1,279 @@
...
+        Then, memory instruction on the LHS will be
+        transformed firstly.
+
+        :param assignblk: assignblock
+        :return: sorted list of expressions
+        """
+        instructions = []
+        for dst in assignblk:
+            # dst = src
+            aff = m2_expr.ExprAff(dst, assignblk[dst])
```
commial: You can use `assignblk.dst2ExprAff(dst)` instead

miasm2/ir/ssa.py
```diff
@@ -0,0 +1,279 @@
...
+        for index, assignblk in enumerate(ib.irs):
+            # list of instructions
+            instructions = []
+            # insert SSA instructions
+            for dst in assignblk:
+                instructions.append(ssa_iter.next())
+            # replace instructions of assignblock in IRA block
+            ib.irs[index] = AssignBlock(instructions)
+
+    def _copy_block(self, block):
```
commial: Idem, `block` is quite misleading here. You can use `label` instead.

As this function is more related to `ir`, you can move it in the class `ir`.

miasm2/ir/ssa.py
```diff
@@ -0,0 +1,279 @@
...
+
+                # rebuild SSA expression
+                e = m2_expr.ExprAff(dst_ssa, src_ssa)
+                self.expressions[dst_ssa] = src_ssa
+
+                # append ssa expression to list
+                ssa_expressions_block.append(e)
+
+        # replace blocks IR expressions with corresponding SSA transformations
+        self._convert_block(ib, ssa_expressions_block)
```
commial: Instead of having a flat list of instructions in `ssa_expressions_block` and iterating again on each destination of each `AssignBlock`, you can move the end of `_convert_block` into the above `for`.
It also seems easier to debug / understand for me, avoiding the fake separation between instructions.

miasm2/ir/ssa.py
```diff
@@ -0,0 +1,279 @@
...
+                # save transformed RHS
+                rhs.append(src_ssa)
+
+            # transform LHS
+            for e in instructions:
+                dst = e.dst
+                dst_ssa = self._transform_expression_lhs(dst)
+
+                # retrieve corresponding RHS expression
+                src_ssa = rhs.pop(0)
```
commial: List are not very effective when [used as queue](https://docs.python.org/2/tutorial/datastructures.html#using-lists-as-queues). You can use `deque` instead, which is more effective

miasm2/ir/ssa.py
```diff
@@ -0,0 +1,279 @@
...
+
+        # set next block
+        ib_ssa.dst = ib.dst.copy()
+
+        # add to SSA blocks dict
+        self.blocks.update({ib_ssa.label: ib_ssa})
+
+        return ib_ssa
+
+    def _rename_expressions(self, block):
```
commial: Idem for `block` -> `label`

commial:
I'm not sure ir `ir.ssa` is the best place for this. May `analysis` could be considered. @serpilliere , any idea?

miasm2/ir/ssa.py
```diff
@@ -322,3 +322,20 @@ def reassemble_expr(self, e):
                 if id_rhs in self.expressions:
                     todo.add(id_rhs)
         return e
+
+
+class SSAPath(SSABlock):
```
commial: This class never seems to be used. May a `map` of `SSABlock.transform` is enough, and then this class useless for now

miasm2/ir/ssa.py
```diff
@@ -339,3 +339,267 @@ def transform(self, path):
...
+        All variable definitions in self.defs contain
+        a set of IRA blocks in which the variable gets assigned
+        """
+        # architecture's instruction pointer
+        instruction_pointer = set(self.ira.arch.pc.values() + [self.ira.IRDst])
+
+        for block in self.graph.walk_depth_first_forward(head):
+            ib = self.get_block(block)
+            # blocks IR expressions
+            ir_expressions = (m2_expr.ExprAff(dst, assignblk[dst])
```
commial: You can use `dst2ExprAff` instead

miasm2/ir/ssa.py
```diff
@@ -339,3 +339,267 @@ def transform(self, path):
...
+                              for assignblk in ib.irs for dst in assignblk)
+            for e in ir_expressions:
```
commial: You only use `e.dst` here, that is to say `assignblk` destinations, so you can directly iterate on `assignblk` (which iterate over destinations)

miasm2/ir/ssa.py
```diff
@@ -339,3 +339,267 @@ def transform(self, path):
...
+                # enforce ExprId
+                if isinstance(e.dst, m2_expr.ExprId):
+                    # exclude architecture's instruction pointer
+                    if e.dst in instruction_pointer:
+                        continue
+                    if e.dst not in self.defs:
```
commial: Instead of this code pattern, you can use `setdefault`:
```Python
self.defs.setdefault(e.dst, set()).add(ib.label)
```

miasm2/ir/ssa.py
```diff
@@ -339,3 +339,267 @@ def transform(self, path):
...
+                todo.add(block)
+                intodo.add(block)
+
+            while todo:
+                block = todo.pop()
+                if block not in frontier:
+                    continue
+
+                # walk through dominance frontier of bloc
+                for node in frontier[block]:
```
commial: To avoid checking twice if `block` is in `frontier`, you can use:
```Python
for node in frontier.get(block, [])
```

miasm2/ir/ssa.py
```diff
@@ -339,3 +339,267 @@ def transform(self, path):
...
+                    if node in done:
+                        continue
+
+                    # remember blocks which contain phi nodes
+                    if node not in self._phinodes:
```
commial: You can also use `setdefault` here (which does modify the value if the key already exists)

miasm2/ir/ssa.py
```diff
@@ -339,3 +339,267 @@ def transform(self, path):
...
+                        todo.add(node)
+
+    @staticmethod
+    def _gen_empty_phi(v):
+        """
+        Generates an empty phi function for a variable
+        :param v: variable
+        :return: ExprAff, empty phi function for v
+        """
+        phi = m2_expr.ExprId("phi", v.size)
```
commial: As this method will be called several times with the same size, you could use a cache *size -> phi*

miasm2/expression/expression_dissector.py
serpilliere: We are currently trying to limit one letter variables in new Miasm code: we recently refactored a code using many one letter variables, and we have learned not to use them the hard way :smile:

miasm2/ir/ssa.py
serpilliere: Correct: the only "property" name must have is it may be immutable.
For exemple, you can write: 
```python
a = ExprId(('EAX', 18), 32)
reg_name, reg_num = a.name
```

That's not really a hack, as it is used for storing `ExprId` of `asm_label` for example.

miasm2/ir/ssa.py
serpilliere: You may be right. Maybe the `analysis` module is more likely to receive such a great PR than `ir` itself.

miasm2/ir/ssa.py
serpilliere: Did I say this was a great feature?

mrphrazer:
Thanks for the great feedback! I adapted everything besides one or to issues, I will comment them later :)

miasm2/ir/ssa.py
mrphrazer: I will move it to `analysis`. However, I am not sure if `SSAPath` itself is useless. It is a simple way to provide backward slicing in a few lines for a whole path. Are you sure that I shall remove that?

miasm2/ir/ssa.py
mrphrazer: That is a really great way. I will adapt that, but it may take some time since I expect that this will break other functions in that class.

miasm2/ir/ssa.py
mrphrazer: I am not exactly sure what you have in mind. Somethins as that?

```python
def _gen_empty_phi(v):
    self.phi_dict = dict()
    if v.size not in phi_dict:
        self.phi_dict[v.size] = m2_expr.ExprId("phi", v.size)
    return  self.phi_dict[v.size] 
```

miasm2/ir/ssa.py
mrphrazer: Same here, really good point. I will change it, but it may last some time.

miasm2/ir/ssa.py
commial: It can be rewritten as a class method, leading `phi_dict` (or `phi_cache`, ...) to be a class attribute. So you can declare it under the `class` statement, as:
```Python
class SSADiGraph(SSA):
   ...
   # _gen_empty_phi cache
   _phi_cache = {}

   def __init__ ...
```

And then:
```Python
@classmethod
def _gen_empty_phi(cls, expr):
    if expr.size not in cls.phi_cache:
        cls.phi_cache[expr.size] = m2_expr.ExprId("phi", expr.size)
    return  cls.phi_cache[expr.size] 
```

But there is no need to spend too much time on this little optimisation.

miasm2/ir/ssa.py
commial: I'm not sure to correctly understand the purpose of `SSAPath`. Is it designed to always work on a path? If it is the case, is a path a list of SSABlock ?

In this last case, `SSAPath` should not inherits from `SSABlock`, but from `UserList` or `list` to get methods such as `append`, iteration, comparison, ... Then it can be initialized with a list of `SSABlock`, and directly works on it (so `transform` method will apply on its internal path).

Am I correct?

miasm2/ir/ssa.py
```diff
@@ -339,3 +339,267 @@ def transform(self, path):
...
+        # compute dominator tree
+        dominator_tree = self.graph.compute_dominator_tree(head)
+
+        # init SSA variable stack
+        stack = [self._stack_rhs.copy()]
+
+        # walk in DFS over the dominator tree
+        for block in dominator_tree.walk_depth_first_forward(head):
+            # restore SSA variable stack of the predecessor in the dominator tree
+            self._stack_rhs = stack.pop().copy()
```
commial: Here, you always copy the top element on the stack before using it.
So, there is no need to copy elements while adding them in the stack (line 517, 537).

miasm2/ir/ssa.py
```diff
@@ -339,3 +339,267 @@ def transform(self, path):
...
+
+    def _rename_phi_lhs(self, block):
+        """
+        Transforms phi function's expressions of an IRA block
+        on the left hand side into SSA
+        :param block: IRA block label
+        """
+        if block in self._phinodes:
+            # create temporary list of phi function assignments for inplace renaming
+            tmp = list(self._phinodes[block])
```
commial: You can avoid the `if` statement using `tmp = list(self._phinodes.get(block, []))`

miasm2/ir/ssa.py
```diff
@@ -339,3 +339,267 @@ def transform(self, path):
...
+        right hand side is of the form
+
+        phi(<var>.<index 1>, <var>.<index 2>, ..., <var>.<index n>)
+
+        :param successor: label of block's direct successor in the CFG
+        """
+        # if successor is in block's dominance frontier
+        if successor in self._phinodes:
+            # walk over all variables on LHS
+            for dst in self._phinodes[successor]:
```
commial: Idem, you can avoid the `if` using a `.get(, [])`

miasm2/ir/ssa.py
```diff
@@ -339,3 +339,267 @@ def transform(self, path):
...
+            for dst in self._phinodes[block]:
+                self.expressions[dst] = self._phinodes[block][dst]
+
+    def _convert_phi(self):
+        """Inserts corresponding phi functions inplace
+        into IRA block at the beginning"""
+        for block in self._phinodes:
+            ib = self.get_block(block)
+            # list of instructions
+            instructions = []
```
commial: Instead of using an intermediate `instructions` object, build a group of `ExprAff` and finally build an `AssignBlock` from it, you can directly declare an `AssignBlock`, and then assign values to it like a dictionnary:
```Python
assignblk = AssignBlock()
for dst in self._phinodes[block]:
    ...
    assignblk[dst] = src
```

test/expression/expression_dissector.py
```diff
@@ -0,0 +1,31 @@
...
+int1 = ExprInt(0xfffffffffffffffc, 64)
+int2 = ExprInt(0xffffffce, 32)
+int3 = ExprInt(0x32, 32)
+compose = ExprCompose([(int2, 0, 32), (int3, 32, 64)])
+op1 = rbp + int1 + compose + rdx
+op2 = int2 + int3
+mem1 = ExprMem(op1, 32)
+mem2 = ExprMem(op2, 32)
+cond = ExprCond(mem1, int2, int3)
+slice = ExprSlice(mem1 + mem2 + cond, 31, 32)
```
commial: You can use `slice1` instead of the keyword `slice`

miasm2/ir/ssa.py
```diff
@@ -1,6 +1,8 @@
 import miasm2.expression.expression as m2_expr
 from miasm2.expression.expression_helper import ExprDissector
-from miasm2.ir.ir import AssignBlock, irbloc
+from miasm2.ir.ir import AssignBlock
+
+from collections import deque
```
commial: (Not necessary to take in account) We try to keep an order in imports, [which is](https://www.python.org/dev/peps/pep-0008/#imports):

1. standard library imports (`collections`)
2. related third party imports
3. local application/library specific imports (`miasm2`)

commial:
Well, this is a great PR (my comments are mainly cosmetics).
I'm just trying to convince myself on a last point (apart the discussion on `SSAPath`), regarding the handling of parallel instructions and memory expressions.

mrphrazer:
Hi,

thanks again! In the following a few explanations
 
## Memory

Regarding the parallel instructions and memory, assume `push rbp`. This will be translated into the following parallel instructions:

```
RSP = RSP + (- 0x8)
@64[RSP + (- 0x8)] = RBP
RBP = RSP
```

In SSA form, it has to be

```
RSP.0 = RSP + (- 0x8)
@64[RSP + (- 0x8)] = RBP
RBP.0 = RSP
```

If we transform the left-hand side and RSP would be translated before the memory instruction, the memory instruction's SSA transformation is incorrect because of `RSP`:
```
RSP.0 = RSP + (- 0x8)
@64[RSP.0 + (- 0x8)] = RBP
RBP.0 = RSP
```

## SSA Path

I have quickly written a short script to illustrate the usage of SSAPath:

```python
from miasm2.analysis.ssa import SSAPath
from miasm2.expression.expression import ExprId


def analyse_path(ssa):
    # just an example
    irdst = ExprId("IRDst.1", 64)

    output = "{}: {}\n".format(irdst, ssa.reassemble_expr(irdst))

    print output


code = ""
code += "554889e5897dec8975e88b45e80145ecd165e88b55ec8b45e8"
code += "01d03d380500007514c745fc00000000c745ec000000008345"
code += "e802eb20c745fc060000008b45e80145ec8b55ec8b45fc01d0"
code += "85c07507b800000000eb05b8010000005dc3"

m = Machine("x86_64")
mdis = m.dis_engine(code.decode("hex"))

ira = m.ira(mdis.symbol_pool)
ira_ssa = m.ira(mdis.symbol_pool)

asm_blocks = mdis.dis_multibloc(0)

for block in asm_blocks:
    ira.add_bloc(block)

ssa = SSAPath(ira)

start_label = ira.get_bloc(0x0).label
end_label = ira.get_bloc(0x5b).label

for index, path in enumerate(ira.graph.find_path(start_label, end_label)):
    # transform path into SSA
    ssa.transform(path)

    # update ira_ssa
    ira_ssa.blocs.update(ssa.blocks)
    ira_ssa._gen_graph()

    # write graph
    open("/tmp/" + str(index) + ".dot", "wb").write(ira_ssa.graph.dot())

    # do some analysis
    analyse_path(ssa)

    # reset SSA
    ssa.reset()
    ira_ssa.blocs = dict()
```

Here, `analyse_path` is a placeholder for an arbitrary analysis of the transformed path. One intention of `SSAPath` is to provide the ability to locate paths that with certain characteristics between two nodes, for instance, in the case of data flow analysis.

## Current state and todo

To sum up, the following things remain on the TODO

* apply the today's feedback
* modify variable renaming
* apply changes to `_convert_block`
* apply changes to `_gen_empty_phi`
* regressions tests for SSA
*  discuss `SSAPath`
* add SSA to `example/disasm/full.py`

Regarding regression tests, we do not have an idea how useful tests could look like. Perhaps it can stay open for another PR. In addition, we would prefer if you could add SSA to the examples.

I will update the post when parts of the TODO have been applied.

miasm2/ir/ssa.py
mrphrazer: Well, we did this in the beginning. It turned out that is does not work. `stack.append(self._stack_rhs)` does not copy the elements, it stores a reference to `self._stack_rhs`. If the elements within the stack will be modified, this will be applied to the same instance that is located several times in `stack`. 

It may be the case that one of the `.copy()` is unnecessary, but debugging this is hard. Side effects may be hidden in details.

miasm2/expression/expression_helper.py
```diff
@@ -674,3 +674,166 @@ def possible_values(expr):
         raise RuntimeError("Unsupported type for expr: %s" % type(expr))
 
     return consvals
+
+
+class ExprDissector(object):
```
mrphrazer: Hi!

Neat trick, I did not know that :) I am not definitively sure, but  I see two problems:

1. It is recursive. Where possible, I try to avoid recursions in python since it cannot be optimised. ; if expressions are nested, time and memory usage will explode.

2. The main reason: Parsing standard subexpressions is a nice feature, but not really used that much within this SSA implementation. However, you can add additional filter criteria such as in `variables`  (which is often used in SSA).  

At least, I think it makes sense to have a class as `ExprDissector` in which you can define arbitrary filters (e.g., parsing memory expressions that depend on the stack pointer). Do you agree?

serpilliere:
I totally agree with you for the recursive vs worklist. But in this case, (tell me if I am wrong) the parsed expressions are limited to expressions which come from the IR of instruction, which are in most case really simple. (this is different from fat expressions resulted from a symbolic emulation for example)

The feature is indeed interesting
@commial, any thoughts?

mrphrazer:
If you operate on assembly-transformed code, you are 	absolutely right: the translated instructions are really simple. Therefore, it will not make much difference in this case. 

However, I was thinking about more generalised  use cases. We can also perform SSA-based analysis of custom IRA graphs. For instance, you might create basic block summaries (e.g., via symbolic execution of the basic block) and replace the basic blocks with their summaries. In addition, you might subclass SSA and extend it with custom expressions that constraint some characteristics. In these cases, the worklist approach might be an advantage.

commial:
Well, currently with Miasm, expression are very often handled with recursive methods.
For instance, you can take expression simplification (and then symbolic execution), translators (and then the emulation part) and, even, expression display (through `__str__`, `__repr__`).

So, as the visitor paradigm seems easier and less prone to error, my vote go for it. There is no gain to handle iteratively expression that you can't use after.

This is a known limitation of Miasm (undocumented :kissing:), expression can't exceed a depth equals to Python recursive limit. We don't need it for now, maybe in the future, but it will be associated with a pretty big patch.

(Sorry for the delay, I didn't notice my opinion was expected. Your feature is actually very cool, so I expect it will be merge in a near future)

PR #353, opened 2016-04-17
Fix pusha16

serpilliere:
Fix pusha/popa semantic for 16/32 bit


miasm2/arch/x86/sem.py
```diff
@@ -1064,51 +1064,53 @@ def popfw(ir, instr):
...
-    opmode, admode = s, instr.v_admode()
-    if not s in [16, 32, 64]:
-        raise ValueError('bad size stacker!')
-
     regs = [
-        mRAX[instr.mode][:s], mRCX[instr.mode][
-            :s], mRDX[instr.mode][:s], mRBX[instr.mode][:s],
-        mRSP[instr.mode][:s], mRBP[instr.mode][:s],
-        mRSI[instr.mode][:s], mRDI[instr.mode][:s]]
+        mRAX[size], mRCX[size],
```
commial: You can move the `[size]` inside the loop, and then extract `regs` outside the function to reuse it in `popa`, and also be able to compare with stack reg without the need to `[size]` before (in `popa`).

PR #361, opened 2016-04-26
x86 arch: extended opcode decoding for MOVUPS

mrphrazer:
I also tried to add `660f73d808` -- `PSRLDQ XMM0, 0x8`, but, up until now , it doesn't work. 

```python
addop("psdrldq", [bs8(0x0f), bs8(0x73), pref_66] +
      rmmod(xmm_reg, rm_arg_xmm) + [u08], [rm_arg_xmm, u08])
```


serpilliere:
Hi!

Ok, the assembly/disassembly part of Miasm is, in my opinion,  the worst part to understand: so I don't blame you :smile: 

First of all, to do a little code factorization, you can use the last byte of the 0x10 as a `swaparg`, which will swap arguments depending of this bit. The code becomes:
```python
addop("movups", [bs8(0x0f), bs("0001000"), swapargs, no_xmm_pref] +
      rmmod(xmm_reg, rm_arg_xmm), [xmm_reg, rm_arg_xmm])
```
But in this case, we have to say Miasm what is the default order, so we have to add the `[xmm_reg, rm_arg_xmm]`

For the `psrldq`, you were pretty close. If you read the definition of the instruction:
```
66 0F 73 /3 ib
```

The `/3` means there is no reg in the modrm, and the reg value is 3. In Miasm, you replace the `xmm_reg` by `d3`, which gives:
```
addop("psrldq", [bs8(0x0f), bs8(0x73), pref_66] +
      rmmod(d3, rm_arg_xmm) + [u08], [rm_arg_xmm, u08])
```

May you modify your PR to include those modifications?

mrphrazer:
Thanks very much for the explanation! Done :)

serpilliere:
Nice!

Last little note: Can you merge your two first commits, and push/force to your branch?
This will result in a final PR with only two *perfects* commits. (the first one is not needed in the branch history)

serpilliere:
Thanks!

PR #374, opened 2016-06-03
Fix asmbloc:label2block

mrphrazer:
With this check, a key-error will not raised. Instead, it can be handled without try/except.


commial:
Actually, this is on purpose. You shouldn't end on a case in which `label` is not in the dictionary.
I prefer to let it that way, raising an error on IR mis-building rather than silently pass.

In addition, it is often considered "more pythonic" to handle this `KeyError` in the caller rather than in the callee.
I close the PR, but I'm completely open to your arguments, do not hesitate to comment.

PR #380, opened 2016-06-19
X86: add no_rex/fix xchg

serpilliere:


miasm2/arch/x86/arch.py
```diff
@@ -826,6 +826,13 @@ def add_pre_dis_info(self, pre_dis_info=None):
         self.rex_b.value = pre_dis_info['rex_b']
         self.rex_x.value = pre_dis_info['rex_x']
         self.rex_p.value = pre_dis_info['rex_p']
+
+        if hasattr(self, 'no_rex') and\
+           (self.rex_r.value or self.rex_b.value or
```
commial: Cosmetics: a list of `or` can be rewritten  `any((a, b, ...))` if you don't care about lazy eval (like here).

PR #384, opened 2016-06-24
Win_api: info

serpilliere:


commial:
We could imagine adding a debug log on stdcall return wrapper to out the returned value, instead of manually adding this kind of logging in each function...

Issue #385, opened 2016-06-26
Bad return of ExprOp_inf_signed with negative ExprInt


jbcayrou:
Hello
I try to use ExprOp_inf_signed but the comparison is wrong with negative integer.
0 < -15 (0xffffff00) is normaly false but ExprOp_inf_signed return 1 and not 0

```python
from miasm2.expression.expression import *
from miasm2.expression.simplifications import expr_simp
from miasm2.expression.simplifications import ExpressionSimplifier
from miasm2.expression.simplifications_cond import ExprOp_inf_signed, ExprOp_inf_unsigned, ExprOp_equal

expr_simp.enable_passes(ExpressionSimplifier.PASS_COND)
print expr_simp(ExprOp_inf_signed(ExprInt32(0), ExprInt32(0xfffffff0)))
0x1
print expr_simp(ExprOp_inf_signed(ExprInt32(-3), ExprInt32(-1)))
0x0
```

PR #386, opened 2016-06-26
Fix ExprOp_inf_signed + unit tests

jbcayrou:
Fix ExprOp_inf_signed
See #385

### DIFF
--- a/miasm2/expression/simplifications_cond.py
+++ b/miasm2/expression/simplifications_cond.py
@@ -198,14 +198,25 @@ def __comp_signed(arg1, arg2):
     @arg1, @arg2: ExprInt"""
 
     val1 = arg1.arg
+    s1 = 1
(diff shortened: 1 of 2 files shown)


commial:
Nice spot!

Regarding the fix, I prefer fixing the value conversion instead of adding extra variables and handling each cases (+/+, +/-, -/+, -/-). These fix is done in #388. This is why I close this PR.

Actually, there were 2 bugs:
* `arg1.arg` is a `moduint`, not a Python `int`, so -1 is still 0xffffffff
* a ^ b + c is equivalent to a ^ (b + c), so some parenthesis are needed (here we want (a ^ b) + c)

jbcayrou:
Ok, thanks !
In fact I did not correctly see these two bugs. I also prefer your fix :)

PR #404, opened 2016-08-03
X86: fix pextrw

serpilliere:


miasm2/arch/x86/sem.py
```diff
@@ -3619,6 +3619,13 @@ def ps_rl_ll(ir, instr, a, b, op, size):
     for i in xrange(0, a.size, size):
         slices.append((m2_expr.ExprOp(op, a[i:i + size], count[:size]),
                        i, i + size))
+
```
commial: The instantiation of `e`, `e_zero` and `e_do` could be done after the `if`

PR #413, opened 2016-09-01
The R15 register is declared multiple times

guedou:
Here is a small PR that removes duplicated R15* declaration.


commial:
Nice spot!

serpilliere:
Nice spot! Thanks.

PR #420, opened 2016-09-05
Fix EmulatedSymbExec segment management

WilliamBruneau:
Fix EmulatedSymbExec way to simplify a 'segm' ExprOp. See commit's comment for more detail.

### DIFF
--- a/miasm2/jitter/emulatedsymbexec.py
+++ b/miasm2/jitter/emulatedsymbexec.py
@@ -90,11 +90,22 @@ class EmulatedSymbExec(symbexec):
         """Handle 'segm' operation"""
         if expr.op != "segm":
             return expr
+
         segm_nb = int(expr.args[0].arg)
         segmaddr = self.cpu.get_segm_base(segm_nb)
-        return e_s(m2_expr.ExprOp("+",
-                                  m2_expr.ExprInt(segmaddr, expr.size),
-                                  expr.args[1]))
+
+        arch_size = self.ir_arch.attrib
+
+        base = m2_expr.ExprInt(segmaddr, arch_size)
+
+        if expr.args[1].size < arch_size:
+            offset = expr.args[1].zeroExtend(arch_size)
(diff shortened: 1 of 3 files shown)


miasm2/jitter/emulatedsymbexec.py
```diff
@@ -92,9 +92,13 @@ def _simp_handle_segm(self, e_s, expr):
             return expr
         segm_nb = int(expr.args[0].arg)
         segmaddr = self.cpu.get_segm_base(segm_nb)
-        return e_s(m2_expr.ExprOp("+",
-                                  m2_expr.ExprInt(segmaddr, expr.size),
-                                  expr.args[1]))
+	
+	arch_size = self.ir_arch.attrib
+	expr.args[1]._size = arch_size
```
commial: We actually try to keep `Expr` immutable, and then avoiding a direct modification of `._size`.
You may want to check the size, and then use a `zeroExtend(arch_size)` or a `[:arch_size]` depending on the result.

miasm2/jitter/emulatedsymbexec.py
```diff
@@ -90,11 +90,22 @@ def _simp_handle_segm(self, e_s, expr):
...
         if expr.op != "segm":
             return expr
+
         segm_nb = int(expr.args[0].arg)
         segmaddr = self.cpu.get_segm_base(segm_nb)
-        return e_s(m2_expr.ExprOp("+",
-                                  m2_expr.ExprInt(segmaddr, expr.size),
-                                  expr.args[1]))
+
+        arch_size = self.ir_arch.attrib
```
commial: Using the `attrib` will only work for *x86*. For instance, on *arml*, the `attrib` value is `l`.
In [`llvmconvert`](https://github.com/cea-sec/miasm/blob/master/miasm2/jitter/llvmconvert.py#L738), I rather use the size of `PC`

commial:
Fix by #492

PR #422, opened 2016-09-05
Fix seh tib

serpilliere:


test/arch/x86/unit/mn_seh.py
```diff
@@ -23,8 +23,9 @@ def deal_exception_priv(jitter):
 
     def init_machine(self):
         super(Test_SEH, self).init_machine()
-        build_teb(self.myjit, FS_0_AD)
         set_win_fs_0(self.myjit)
+        tib_ad = self.myjit.jitter.cpu.get_segm_base(jitter.cpu.FS)
```
commial: It seems that `.jitter` is already obtain through `.myjit`

PR #423, opened 2016-09-07
Arch/x86: fix sldt

serpilliere:


miasm2/arch/x86/sem.py
```diff
@@ -2778,8 +2778,8 @@ def sidt(ir, instr, a):
 
 
 def sldt(ir, instr, a):
-    # XXX TOOD
-    e = [m2_expr.ExprAff(exception_flags, m2_expr.ExprInt32(EXCEPT_PRIV_INSN))]
+    print "DEFAULT SLDT ADDRESS %s!!" % str(a)
```
commial: May a warning log could be a better solution

PR #430, opened 2016-09-15
Feature/int expr int

commial:
Introduce a more intuitive (at least to me) way to get the value of a `ExprInt`.
Instead of `int(x.arg)` or `x.arg.arg`, `int(x)` is now available (and `long(x)` too)


miasm2/expression/modint.py
```diff
@@ -53,7 +53,7 @@ def __int__(self):
 
     def __long__(self):
         return long(self.arg)
-
+
```
serpilliere: Bad white space

PR #438, opened 2016-10-05
Fix build mac os x

alvarofe:
Was trying miasm on mac os x but I wasn't able to build it. That seems to fix the issue.


commial:
This is actually the patch introduced in #27 . It seems OK for me to land it.
Did you get the emulation working?

alvarofe:
Did you have a test for that to try it?

commial:
Yes, you can launch Miasm regression tests and example through:
```
cd tests;
python test_all.py
```

dnutiu:
Got the same error

PR #450, opened 2016-11-10
Jitter: Add max exec per jit call option

serpilliere:


Romain6:
Hi,
Does this feature should work ?

I tried to set it to 5 and emulate a loop, but the counter does not decrement on each iteration.
This is the code : 
```
0:  b9 00 00 00 f0          mov    ecx, 0x10
5:  31 c0                   xor    eax ,eax
7:  e2 fc                   loop   0x5
```

I used a script similar to [x86_32.py](https://github.com/cea-sec/miasm/blob/master/example/jitter/x86_32.py) and displayed the counter in the 'for' loop ([Jittc.c#163](https://github.com/cea-sec/miasm/blob/5f728a282520b5b98de64053d98b935b206b61c1/miasm2/jitter/Jittcc.c#L163))

```
loc_0000000040000000:0x40000000
MOV        ECX, 0x10
XOR        EAX, EAX
LOOP       loc_0000000040000005:0x40000005
->	c_next:loc_0000000040000009:0x40000009 	c_to:loc_0000000040000005:0x40000005 
cpt : 4
40000000 MOV        ECX, 0x10
40000005 XOR        EAX, EAX
40000007 LOOP       loc_0000000040000005:0x40000005
40000005 XOR        EAX, EAX
40000007 LOOP       loc_0000000040000005:0x40000005
[...]
40000005 XOR        EAX, EAX
40000007 LOOP       loc_0000000040000005:0x40000005
cpt : 3
```

serpilliere:
There is a case not handled by the patch (discovered after the merge):
If a block loop to its own code, the callback is not called. I added the issue #465.

PR #451, opened 2016-11-10
Add simplification: ({a, 0x0} + 0x1)[0:32] => a + 0x1

commial:


commial:
This simplification is not always correct, and must be more precise. Hence, it is less convenient to having it in the main pool of simplification (corner case)

PR #452, opened 2016-11-13
arm: Add sem support for smul and smulw instructions

laanwj:
Add support to convert all the `smul[tb][tb]` and `smulw[tb]` variants into expressions.


serpilliere:
Thank you!

PR #455, opened 2016-11-21
add dont_dis_retcall_funcs set to disasmEngine

a-vincent:
With this patch, we can tell miasm which functions are non-returning so that disassembly stops after a call to such a function.
Joint work with @carolineLe


serpilliere:
Thank you!

PR #457, opened 2016-12-15
Regression test for apply_expr/apply_change/eval_ir

commial:
Add a way to insert "a knowledge" in a symbol execution context.

If the current context is:
```
a = 0
```
Then, on `b = a`:
* `apply_expr`, `eval_ir` will end with (applying the side effect of an instruction):
```
a = 0
b = 0
```
* `apply_change` will end with (injecting a statement):
```
a = 0
b = a
```

In the last example, the value of `b` is the initial value of `a` (`_init` are no longer needed since #445)

### DIFF
--- a/miasm2/ir/symbexec.py
+++ b/miasm2/ir/symbexec.py
@@ -392,6 +392,28 @@ class symbexec(object):
 
         return pool_out.iteritems()
 
+    def apply_change(self, dst, src):
+        """
+        Apply @dst = @src on the current state WITHOUT evaluating both side
+        @dst: Expr, destination
+        @src: Expr, source
+        """
+        if isinstance(dst, m2_expr.ExprMem):
+            mem_overlap = self.get_mem_overlapping(dst)
+            for _, base in mem_overlap:
+                diff_mem = self.substract_mems(base, dst)
+                del self.symbols[base]
+                for new_mem, new_val in diff_mem:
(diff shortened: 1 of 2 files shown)


serpilliere:
Does the name is not a bit confusing?
After reviewing the `symbexec` code, here are some of the exposed *api*:
```
eval_expr
eval_ir_expr
eval_ir
emulbloc
emul_ir_bloc
apply_expr
apply_change
```

I agree that this new function is useful, and *must* be integrated to the code. But a voice in my head is telling me that it's hard to understand the current API (one of the causes is the function's names) and this new one will add a bit more to the *strangeness* of the API.

The fact is that I don't have another name more explicit to propose. I just wanted to raise this fact.

So:
* Do you agree with me?
* Do you have a more explicit name in mind?
* Do you prefer I merge the PR and see later this trouble ? (and maybe we will change the name again later)

I will create a PR to add some documentation to the *API*: I will push it after or before this one as you like (I may move some functions as private as well)

commial:
I completely agree with your points.

The way I see it, it would be great to refine all these APIs, that is to say avoid exposing some of them and renaming the others to expose a clear & clean API.
But, this involves a modification of corresponding calling sites (and an API break). So, I prefer to have it in a separated and dedicated PR.

Does this sound OK for you?

serpilliere:
Perfect!

PR #458, opened 2016-12-16
Feature llvm jitter

commial:
As stated in #5, the current support of LLVM is broken.
In this work, the Numba [llvmlite](https://github.com/numba/llvmlite) backend is used instead of pyllvm.
One of its main advantage is that the IR module is entirely built in Python, and the LLVM backend is only called at the end (avoiding wasting time in the wrapper, as before). In addition, it is more up to date.

As a lot of internal API have changed since the first use of LLVM (`IRDst`, `vm_MEM*`, breakpoints, code generation optimisations as `codegen`, JiT loop while code is already Jitted with an optional maximum number of iteration, `VmMngr` move to `JitCpu` attribute, specific `dump_gpregs` for certain architecture, ...), the current code base needs a big update (perhaps a merge with `codegen`).

Here is a TODO list of what is needed, and what is done for now:

- [x] Adapt to `IRDst`
- [x] Modify API to call `vm_MEM*` wrappers
- [x] Modify API to handle architecture specific `dump_gpregs` wrappers
- [x] Expose specific operation code (as `parity`)
- [x] Merge modules between them (may with `ModuleRef.link_in`) to avoid cumulative construction :arrow_forward: finally not done, because it does not avoid cumulative construction. Then, functions are independent modules, handled within a common `ExecEngine`
- [x] Update doc
- [x] Pass `md5_arm` test with a correct MD5
- [x] Pass `unpack_upx` test with the same output
- [x] Pass all regression tests (77 / 77 LLVM only tests)
- [x] Update Travis CI
- [x] Handle compilation optimization
- [x] Clean up unused APIs

Others points could be considered as optimization, and will be handled in others PRs. With this in mind, it is not possible to conclude on performance for now.
For instance:

- Loop in the wrapper code while JiT blocks are already available
- Cache to disk already generated block (as string or LLVM bitcode)
- Separate code relative to JIT and code relative to IR translation
- `jit_maxline` option handling
- Support of floating point operations

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,19 @@ python:
 - "2.7"
 addons:
     apt:
+        sources: ['llvm-toolchain-precise-3.8', 'ubuntu-toolchain-r-test']
         packages:
             - make
             - gcc
             - python-virtualenv
             - unzip
+            - llvm-3.8
+            - llvm-3.8-dev
+            - g++-5
 before_script:
 - "cd .."
+- "export LLVM_CONFIG=$(which llvm-config-3.8)"
+- "export CXX=$(which g++-5)"
 # make virtual env
 - "python /usr/lib/python2.7/dist-packages/virtualenv.py virtualenv;"
 - "cd virtualenv;"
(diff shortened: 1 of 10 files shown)


commial:
This update should takes in account and fix @serpilliere's off-the-records remarks.

PR #459, opened 2016-12-22
Add a only-tags option for test_all

commial:
Add the possibility to launch tests regarding only a subset of tag (whitelist version of `omit-tags`).

For instance:
```
$ python test_all.py -o z3
...
Result: 27/27 pass
```

This is quite useful while, for instance, refactoring a given Miasm segment...

### DIFF
--- a/test/test_all.py
+++ b/test/test_all.py
@@ -57,7 +57,7 @@ class ArchUnitTest(RegressionTest):
 
 # script -> blacklisted jitter
 blacklist = {
-    "x86/unit/mn_float.py": ["python"],
+    "x86/unit/mn_float.py": ["python", "llvm"],
 }
 for script in ["x86/sem.py",
                "x86/unit/mn_strings.py",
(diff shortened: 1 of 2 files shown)


test/test_all.py
```diff
@@ -703,15 +714,15 @@ class ExampleJitterNoPython(ExampleJitter):
             "'py-llvm 3.2' module is required for llvm tests"
 
         # Remove llvm tests
-        if TAGS["llvm"] not in exclude_tags:
+        if (not include_tags and TAGS["llvm"] not in exclude_tags):
```
serpilliere: Maybe there is an unmanaged case where you only include a tag which will pull a dependency on tcc for example, on a machine which does not support tcc.

commial: How should it be handled? Drop the test case? Silently? Abort?

serpilliere: Oh! I don't know: I was not sure about me on the review of this case.
I may prefer the abort case, but it's up to you.
If you are aware of the case and prefer to let the code like this, tell me.

commial:
So now, there is 2 distincts part:

1. Creation of the testset (full, or tags mark as included with their dependencies)
2. Pruning of the testset according to the exclusion list

To avoid unwanted behavior, the impossibility of using both inclusion and exclusion is conserved in the CLI handling.
But now, if I ask for, say, `llvm,tcc`, and `llvm` is not supported on my install, I end with `tcc` tests and `llvm` tests dependencies.

It seems fine for me, do you agree?

commial:
Finally, I removed this weird comportement, which seems unexpected.

Now, there are 2 options:
* `-t OMIT_TAGS`: blacklist mode
* `-o ONLY_TAGS`: whitelist mode

They can be used together.
For instance, `-o tcc -t example,qemu` will only launch regression tests (without QEMU) on TCC and their corresponding dependencies.

PR #460, opened 2016-12-22
Remove useless code

commial:
Remove a few redundancy in `codegen`.

In addition, a corner case is "watched" through an assert, the details are in the commit comment.
If this case is trigger one day, one will hopefully just have to blame this assert to find the concern about it.
I prefer this way instead of fixing now the bug, because no example raising it comes to my mind with current M2 implemented architectures.

### DIFF
--- a/miasm2/jitter/codegen.py
+++ b/miasm2/jitter/codegen.py
@@ -44,25 +44,8 @@ class CGen(object):
     }
     """
 
-    CODE_EXCEPTION_MEM_POST_INSTR = r"""
-    // except fetch mem post instr
-    if (VM_exception_flag) {
-        %s = %s;
-        BlockDst->address = %s;
-        return JIT_RET_EXCEPTION;
-    }
-    """
-
     CODE_EXCEPTION_AT_INSTR = r"""
     if (CPU_exception_flag_at_instr) {
-        %s = %s;
-        BlockDst->address = %s;
-        return JIT_RET_EXCEPTION;
-    }
-    """
-
-    CODE_EXCEPTION_POST_INSTR = r"""
-    if (CPU_exception_flag) {
         %s = %s;
         BlockDst->address = %s;
         return JIT_RET_EXCEPTION;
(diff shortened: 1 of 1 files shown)


miasm2/jitter/codegen.py
```diff
@@ -283,6 +283,8 @@ def traverse_expr_dst(self, expr, dst2index):
                         "0")
 
         else:
+            # Spot case like Cond?X:Y
+            assert -1 not in dst2index.values()
```
serpilliere: I don't agree with this patch.

If you have a destination operation which looks like:
```
IRDst = ExprCond(cond, EAX, EBX)
```
The Generated code will be:
```
DST_case = cond?-1:-1;
DST_value = cond?EAX:EBX;
```
`DST_case` will be -1 in every cases (which is correct)
But the `DST_value` will be corerctly computed. So no problem in both cases.

Maybe I didn't get your point?

commial: Well, the value is erased in `dst2index`. But the keys of `dst2index` are further used to generate the goto (in some case). I'm not completely sure that there is no way that this impact the resulting `switch`. There are several conditions to consider (ExprInt in or out of `instr_offsets`, label, ExprId, ...) in `gen_dst_goto`, `gen_goto_code` and in this method.
If you are sure about it, I will remove this commit.

serpilliere:
```
Well, the value is erased in ...
```
The value of what?

I insist in order to see through your eyes.
In the code, only the default case will set an entry in `dst2index`.
The important part of the code is to keep the **value** and the **type** synchronized.

And I didn't see (for now :smile:) any case which case do this de-synchronization.
So for me two solutions here:
* either you remove the assert
* or you spend time to get an example which break the code

commial: I finally convince myself that the case could not happen. Thanks for the clarification.

I've removed the commit.

PR #463, opened 2016-12-23
Add expr type access

serpilliere:
Replace `isintance(expr, ExprIn)` by `expr.is_int()`
Code cleanup

### DIFF
--- a/example/expression/simplification_add.py
+++ b/example/expression/simplification_add.py
@@ -30,7 +30,7 @@ def simp_add_mul(expr_simp, expr):
 
         # Effective simplification
         return m2_expr.ExprOp("*", expr.args[0],
-                              m2_expr.ExprInt_from(expr.args[0], 3))
+                              m2_expr.ExprInt(3, expr.args[0].size))
     else:
         # Do not simplify
         return expr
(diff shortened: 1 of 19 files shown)


commial:
OK, so Is the goal of this PR a better code readability? (Or am I missing something?)

Going this way, why not applying your change to every Miasm code? Such as in `asmbloc.py:expr_is_*` for instance.

Final point: if homogenization to this new API is decided, I'll prefer to wait current big PR to land to avoid useless conflict. What do you think?

serpilliere:
The `expr.is_XXX` adds feature to `Expr` class **only**. The `expr_is*` if defined in `asmbloc`because this code uses class defined in this module.

The `asmbloc` module in miasm depends on `expression`, and the `expression` module does **not** rely on asmbloc (there is no cross references)

So for me, no, this code cannot be generalized for `expr_is*`. By the way, for my point of view, this is not really a code readability, it's more a convenience to write code :smile:

commial:
Ok, this is not I wanted to say. My point is that you can apply this replacement in others miasm code parts, as in the body of expr_is_* function

serpilliere:
Yes: If you don't have any arguments against the use of such a code, I may effectively replace other parts.

serpilliere:
Note this can be done in future patch as well. There is no need to fix **every** code in the same code.

commial:
Fair point. Work for me!

PR #467, opened 2017-01-05
Fix dg emul

serpilliere:
Fix depgraph result emulation in some corner cases.
Add regression tests.

### DIFF
--- a/miasm2/analysis/depgraph.py
+++ b/miasm2/analysis/depgraph.py
@@ -53,8 +53,8 @@ class DependencyNode(object):
     def __cmp__(self, node):
         """Compares @self with @node."""
         if not isinstance(node, self.__class__):
-            raise ValueError("Compare error between %s, %s" % (self.__class__,
-                                                               node.__class__))
+            return cmp(self.__class__, node.__class__)
+
         return cmp((self.label, self.element, self.line_nb),
                    (node.label, node.element, node.line_nb))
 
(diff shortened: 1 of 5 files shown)


miasm2/analysis/depgraph.py
```diff
@@ -285,8 +289,18 @@ def emul(self, ctx=None, step=False):
         assignblks = []
 
         # Build a single affectation block according to history
-        for label in self.relevant_labels[::-1]:
-            assignblks += self.irblock_slice(self._ira.blocs[label]).irs
+        if self.relevant_labels:
```
commial: The whole loop may be simplified by something like:
```Python
for index, label in enumerate(self.relevant_labels[::-1], 1):
    stop_line = self.input_line_nb if (index == len(self.relevant_labels) and
                                       label == self.input_label) else None
    assignblks += self.irblock_slice(self._ira.blocs[label], stop_line).irs
```

miasm2/analysis/depgraph.py
```diff
@@ -601,7 +620,7 @@ def get(self, label, elements, line_nb, heads):
             if (not state.pending or
                     state.label in heads or
                     not self._ira.graph.predecessors(state.label)):
-                yield dpResultcls(state, self._ira)
+                yield dpResultcls(state, self._ira, label, line_nb)
```
commial: To my opinion, it seems nicer to have the asking input rather than only a few information on it in argument.
By the way, it seems that the .inputs property is just used to propagate the initial state elements, and is only used in DependencyResult. So it can be removed (for simpler and lighter dependency state), and the `initial_state.elements` can be used instead.

PR #481, opened 2017-01-26
Duplicated WZR definition

guedou:
The WZR register is defined three times.

Before the patch:
```
>>> from miasm2.arch.aarch64.regs import all_regs_ids
>>> len(all_regs_ids), len(set(all_regs_ids))
(234, 232)
```
After the patch:
```
>>> from miasm2.arch.aarch64.regs import all_regs_ids
>>> len(all_regs_ids), len(set(all_regs_ids))
>>> (232, 232)
```


commial:
Nice spot!

PR #482, opened 2017-01-29
Add reference to R2M2

commial:
Add a reference to @guedou's project [R2M2](https://github.com/guedou/r2m2)


guedou:
Thanks guys !

PR #486, opened 2017-02-02
Add regression test for CPUID support

commial:
As pointed by @serpilliere, there were no test for #485.
This test check `CPUID` support for every jitter.


serpilliere:
Thanks!

PR #489, opened 2017-02-06
Example/trace

commial:
Add a "tracer" example showing basic memory instrumentation (in Python) and callback before each instruction.

### DIFF
--- a/example/jitter/trace.py
+++ b/example/jitter/trace.py
@@ -1,0 +1,58 @@
+"""
+This example demonstrates two instrumentation possibility:
+ - instrumentation executed at each instruction
+ - instrumentation on jitter behavior (here, memory tracking)
+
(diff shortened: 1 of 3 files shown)


example/jitter/trace.py
```diff
@@ -0,0 +1,58 @@
...
+        print "Read %s: %s" % (expr_mem, value)
+        return value
+
+    def _func_write(self, symb_exec, dest, data):
+        print "Write %s: %s" % (dest, data)
+        return super(ESETrackMemory, self)._func_write(symb_exec, dest, data)
+
+# Parse arguments
+parser = Sandbox_Linux_arml.parser(description="Tracer")
+parser.add_argument("filename", help="PE Filename")
```
serpilliere: ELF Filename

PR #495, opened 2017-02-13
Objc feature

serpilliere:
This module implements C basic manipulators.

From C types descriptions, we can now give back C code representing a Miasm expression.
For exemple on the struct:
```C
struct human {
	unsigned short age;
	unsigned int height;
	char name[50];
};

struct ll_human {
	struct ll_human* next;
	struct human human;
};
```

If we know that RSI is `ptr` and `ptr` is a `struct ll_human*`
we can say that:
`@32[(@64[@64[ptr]]+0xC)]`
has type `uint`
and can be acessed in C with:
```C
((((ptr)->next)->next)->human).height
```

See `example/expression/access_c.py` executed on  `human.bin`

### DIFF
--- a/example/expression/access_c.py
+++ b/example/expression/access_c.py
@@ -1,0 +1,188 @@
+import sys
+
+from miasm2.analysis.machine import Machine
+from miasm2.analysis.binary import Container
+from miasm2.expression.expression import ExprOp, ExprCompose, ExprId
+from miasm2.analysis.depgraph import DependencyGraph
+
+from miasm2.arch.x86.ctype import CTypeAMD64_unk
+
+from miasm2.core.objc import CTypeAnalyzer, ExprToAccessC, CHandler
+from miasm2.core.objc import ObjCPtr
+from miasm2.core.ctypesmngr import CTypesManagerNotPacked
+
+
+"""
+
+This example demonstrates the recovering of possible C types for an arbitrary
+variable in an assembly code (the types are inferred from the function
+argument types). It also displays the C code used to access this variable.
+
+Input:
+* definitions of the C types that can be used by the code
+* layout of structures (packed/not packed)
+* prototype of the analyzed function
+
+Algorithm:
+The DepGraph of the target variable is computed, which gives possible
+expressions for this variable. For each DepGraph solution, if the expression
+depends on typed arguments, the code infers the variable type and displays the C
+code to access this variable.
+
+
+Here be dragons:
+For the moment, Miasm can infer C types (and generate C) for simple expressions.
+To summarize, Miasm only supports accesses that do not involve arithmetic or
+conditional expressions such as the following:
+* var1.field
+* var1[12][4]
+* *(var1.field->tab[4])
+
+Unsupported forms:
+* var1 + var2
+* var1[var2+4]
+* var1?var2->field:6
+
+In the following example, we have an explicit cast for "age", from uint16_t to
+uint64_t, and for "height", from uint32_t to uint64_t. We are adding a naive
+reduction rule to support such a cast.
+
+First, in the type inference engine:
+ExprCompose(int, 0) => int
+Then, in the C generator:
+ExprCompose(var1, 0) => var1
+
+
+"""
+
+def find_call(ira):
+    """Returns (irb, index) which call"""
+
+    for irb in ira.blocs.values():
+        out = set()
+        if len(irb.irs) < 2:
+            continue
+        assignblk = irb.irs[-2]
+        for src in assignblk.itervalues():
+            if not isinstance(src, ExprOp):
+                continue
+            if not src.op.startswith('call_func'):
+                continue
+            out.add((irb, len(irb.irs) - 2))
+        if len(out) != 1:
+            continue
+        irb, index = out.pop()
+        yield irb, index
+
+
+class MyCTypeAnalyzer(CTypeAnalyzer):
+
+    def reduce_compose(self, node, _):
+        """Custom reduction rule: {XXX, 0} -> typeof(XXX)"""
+        if not (isinstance(node.expr, ExprCompose) and
+                len(node.expr.args) == 2 and
+                node.expr.args[1].is_int(0)):
+            return None
+        return node.args[0].info
+
+    reduction_rules = CTypeAnalyzer.reduction_rules + [reduce_compose]
+
+
+class MyExprToAccessC(ExprToAccessC):
+
+    def reduce_compose(self, node, _):
+        """Custom reduction rule: {XXX, 0} -> XXX"""
+        if not (isinstance(node.expr, ExprCompose) and
+                len(node.expr.args) == 2 and
+                node.expr.args[1].is_int(0)):
+            return None
+        found = []
+        for subcgenobj in node.args[0].info:
+            found.append(subcgenobj)
+        return found
+
+    reduction_rules = ExprToAccessC.reduction_rules + [reduce_compose]
+
+
+def get_funcs_arg0(ctx, ira, lbl_head):
+    g_dep = DependencyGraph(ira, follow_call=False)
+    element = ira.arch.regs.RSI
+
+    for irb, index in find_call(ira):
+        line = irb.lines[index]
+        print 'Analysing references from:', hex(line.offset), line
+        g_list = g_dep.get(irb.label, set([element]), index, set([lbl_head]))
+        for dep in g_list:
+            emul_result = dep.emul(ctx)
+            value = emul_result[element]
+            yield value
+
+
+class MyCHandler(CHandler):
+    cTypeAnalyzer_cls = MyCTypeAnalyzer
+    exprToAccessC_cls = MyExprToAccessC
+
+
+def test(data):
+    # Digest C informations
+    text = """
+    struct human {
+            unsigned short age;
+            unsigned int height;
+            char name[50];
+    };
+
+    struct ll_human {
+            struct ll_human* next;
+            struct human human;
+    };
+    """
+
+    my_types = CTypeAMD64_unk()
+    types_mngr = CTypesManagerNotPacked(my_types.types)
+
+    types_mngr.add_c_decl(text)
+
+    # Analyze binary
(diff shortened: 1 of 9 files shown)


commial:
Not sure about the place of `ctypes_templates`. Maybe the parent class `CTypeTemplate` could be moved into `objc`, and `CTypeAMD64_unk` to a `ctype` file in the `arch/x86` directory (to avoid having arch dependent code in the `core` module).

test/test_all.py
```diff
@@ -521,6 +521,10 @@ class ExampleExpression(Example):
                               Example.get_sample("simple_test.bin")],
                              products=["graph_instr.dot", "out.dot"])
 
+testset += ExampleExpression(["access_c.py", Example.get_sample("human.bin")],
```
commial: In the next commit, `pycparser` is noted as optional (which seems a good idea). To reflect this, you could add a tag, as for `z3`, which is automatically disabled if `pycparser` is not found on the machine. Otherwise, one cannot launch tests without this dependency...

miasm2/core/objc.py
```diff
@@ -0,0 +1,1712 @@
...
+
+    def __repr__(self):
+        return '<%s %s>' % (self.__class__.__name__, self.name)
+
+
+class ObjCInt(ObjC):
+    """C integer"""
+
+    def __init__(self):
+        self.size = None
```
commial: `super(ObjCXXX, self).__init__()` is done in other classes, not in this one

miasm2/core/objc.py
```diff
@@ -0,0 +1,1712 @@
...
+    """Int C object"""
+
+    def __init__(self, integer):
+        self.integer = integer
+        self.ctype = ObjCInt()
+
+    def to_c(self):
+        """Generate corresponding C"""
+
+        return "%s" % hex(self.integer)
```
commial: On a big integer, this will return a 0x...L. Is this wanted ?
Not related, but the format string is not needed here

miasm2/core/objc.py
```diff
@@ -0,0 +1,1712 @@
...
+
+    def __repr__(self):
+        return "<%s %s>" % (self.__class__.__name__,
+                            self.integer)
+
+    def to_expr(self):
+        """Generate Miasm expression representing the C access"""
+
+        return ExprInt(self.integer, 64)
```
commial: 64 seems to be an arbitrary value here. I understand it will probably always work, but a shared (set in the parent class?) attribute could be a way to easily update the size

miasm2/core/objc.py
```diff
@@ -0,0 +1,1712 @@
...
+        """Generate Miasm expression representing the C access"""
+
+        raise NotImplementedError("Virtual")
+
+
+class CGenInt(CGen):
+    """Int C object"""
+
+    def __init__(self, integer):
+        self.integer = integer
```
commial: (homogeneity) The type of the argument is ensured in others similar classes

miasm2/core/objc.py
```diff
@@ -0,0 +1,1712 @@
...
+    def __init__(self, name, element, void_p_align, void_p_size):
+        ctype = name.ctype
+        if isinstance(ctype, ObjCPtr):
+            pass
+        elif isinstance(ctype, ObjCArray) and isinstance(ctype.objtype, ObjCArray):
+            ctype = ctype.objtype
+        elif isinstance(ctype, ObjCArray):
+            ctype = ObjCPtr('noname', ctype.objtype, void_p_align, void_p_size)
+        else:
+            raise RuntimeError("Strange case")
```
commial: It looks like a `TypeError` instead of a Runtime one

miasm2/core/objc.py
```diff
@@ -0,0 +1,1712 @@
...
+    elif isinstance(ast, c_ast.UnaryOp):
+        if ast.op == 'sizeof' and isinstance(ast.expr, c_ast.Typename):
+            result = ast_eval_size(ast.expr.type, types_mngr)
+        else:
+            raise NotImplementedError("Not implemented!")
+
+    elif isinstance(ast, c_ast.Constant):
+        result = int(ast.value, 0)
+    elif isinstance(ast, c_ast.Cast):
+        # TODO: Can trunc intergers?
```
commial: Here is a TODO, is it wanted to stay?

serpilliere: For the TODO: wontfix:
I will see in future if we often face declaration with casting in integers of sizes for arrays.

miasm2/core/objc.py
```diff
@@ -0,0 +1,1712 @@
...
+                          offset, obj.size)
+        offset += obj.size
+
+    new_obj.set_align_size(1, offset)
+
+    types_mngr.add_type((name, ), new_obj)
+    return new_obj
+
+
+def ast_gen_struct_align_size(ast, types_mngr):
```
commial: Note for future: to avoid having a lot of module method, and for a better "management" and readability of layout, it could be great to have a `Layout` class, describing how a given layout act on union, align size, ...

miasm2/core/objc.py
```diff
@@ -0,0 +1,1712 @@
...
+    elif isinstance(ast, c_ast.Union):
+        obj = ast_gen_union_align_size(ast, types_mngr)
+    else:
+        raise NotImplementedError("Not implemented!")
+
+    fix_recursive_objects(types_mngr, obj)
+    return obj
+
+
+def ast_parse_declarations(ast, types_mngr):
```
commial: There is a lot of function working on `ast`, beginning the same way and having it as a 1st argument, and passing them to their subcall... it does look like the pattern of a class with its methods and an attribute `ast`. In addition, as this file is quite big, it helps regrouping AST relative work from the others elements

miasm2/core/objc.py
```diff
@@ -0,0 +1,1712 @@
...
+        """Update expr_types
+        @expr_types: Dictionnary associating name to type
+        """
+
+        self.expr_types = expr_types
+
+    CST = ObjCInt()
+
+    def get_typeof(self, base_type, offset, deref, lvl=0):
+        """Return a pointer on the element at @offset of an object of type
```
commial: It seems to returns a  possible list of pointer instead of just one

miasm2/core/objc.py
```diff
@@ -0,0 +1,1712 @@
...
+
+        void_type = self.types_mngr.void_ptr
+
+        if isinstance(src_type, ObjCArray):
+            out = (src.arg, ObjCPtr('noname', src_type.objtype,
+                                    void_type.align, void_type.size))
+        elif isinstance(src, ExprMem):
+            out = (src.arg, ObjCPtr('noname', src_type,
+                                    void_type.align, void_type.size))
+        elif isinstance(src_type, ObjCStruct):
```
commial: isinstance(src_type, (ObjCStruct, ObjCUnion))

example/expression/access_c.py
```diff
@@ -0,0 +1,143 @@
+import sys
+
+from miasm2.analysis.machine import Machine
+from miasm2.analysis.binary import Container
+from miasm2.expression.expression import ExprOp, ExprCompose, ExprId
+from miasm2.analysis.depgraph import DependencyGraph
+
+from miasm2.core.ctypes_template import CTypeAMD64_unk
+
+from miasm2.core.objc import CTypesManager
```
commial: Why not importing all at once?

example/expression/access_c.py
```diff
@@ -0,0 +1,143 @@
...
+    my_types = CTypeAMD64_unk()
+    types_mngr = CTypesManager(CTypesManager.LAYOUT_COMPIL, my_types.types)
+
+    types_mngr.add_c_decl(text)
+
+    # Analyze binary
+    cont = Container.fallback_container(data, None, addr=0)
+
+    machine = Machine("x86_64")
+    _, dis_engine = machine.mn, machine.dis_engine
```
commial: If `machine.mn` is not used, why not "dis_engine = machine.dis_engine"?
Idem next line with `machine.ir`

test/test_all.py
```diff
@@ -521,6 +521,10 @@ class ExampleExpression(Example):
...
+                             depends=[test_human],
+                             products=[])
```
commial: The script produces a file `graph_irflow.dot`

example/expression/access_c.py
```diff
@@ -0,0 +1,143 @@
+import sys
```
commial: This is a really nice example, but it is hard to see the purpose, then why specific reduction are used and finally how the whole thing is use.
As this will be the referring example of use of ObjC features, it will probably be more accessible with a few more comment, and a summary / warm-up at the beginning.
In addition, an example of using `c_to_expr` API could also be great, and take away the sacred aura from using this feature.

miasm2/core/objc.py
```diff
@@ -0,0 +1,1721 @@
+"""
+C helper for Miasm:
+* raw C to Miasm expression
+* Miasm expression to raw C
+* Miasm expression to C type
+"""
```
commial: There is so many object and type involved that a kind of summary on "who does what" would be a great thing, especially when we will come back on the code later.

miasm2/core/objc.py
serpilliere: The code is modified, but not as you suggested:
Those function became members of `type_mngr`. So layout disappears, and is handled in two subclasses.

miasm2/core/objc.py
serpilliere: Correct

example/expression/access_c.py
serpilliere: Correct

test/test_all.py
serpilliere: Done

example/expression/access_c.py
serpilliere: Example added

miasm2/core/objc.py
serpilliere: The code is split: typemanager is now out of objc

test/test_all.py
serpilliere: done

PR #498, opened 2017-02-20
Cleanup travis and use requirements.txt for pip

stephengroat:
follows the python standard, able to use `pip install -r requirements.txt` for python dependencies


commial:
Ho

PR #499, opened 2017-02-20
Cleanup travis, use requirements.txt for pip, add os x

stephengroat:
sorry about the last one, rebased into single commit

#438

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,0 +1,3 @@
+brew 'gcc'
+brew 'tcc'
+brew 'llvm'


serpilliere:
Big thanks for the clean-up!

commial:
Hi, thanks for the rebase and the merge in one PR.

I see that you put the full requirements into `requirements.txt`, including optional ones. I don't know enough `pip`, but is there a common way to differentiate *needed* and *optional* requirements?

In the current state, if one wants to run Miasm on its system using pip, it will end by installing `z3`, `llvm`, ... which are actually not required.
May a second requirement file could be a solution?

stephengroat:
i can change it up by using `setuptools`, which packages are optional?

stephengroat:
please hold on this, following https://github.com/numba/llvmlite/pull/250 for more cleanup

stephengroat:
@alvarofe i got `osx` travis builds going. are you see [this error](https://travis-ci.org/cea-sec/miasm/jobs/208488818#L1794)?

commial:
Hi @stephengroat,

I may have a small part of the answer, as I did some tries a long time ago (see the last commits of https://github.com/commial/miasm/commits/fix/osx).

Regarding optional packages (sorry for the delay...), I got:
Optionals:
* llvmlite
* z3
* pycparser

Mandatory:
* elfesteem
* pyparsing

In an ideal world, it would be great to have a new test dimension (travis matrix if I correctly understand) with and without optionals packages. Indeed, `test_all.py` already avoid launching some tests if optionals dependencies are not met.

Nice catch for llvmlite PR, I hope the patch will be land, it definitively simplifies the setup...

serpilliere:
Erf, the pypi elfesteem package is not compatible with current Miasm: I think it will fail

commial:
For a nicer way to debug, here is a trick:
* add `socat` in `apt` packages
* add `socat tcp-connect:$HOST:$PORT system:bash,pty,stderr || echo "Debug shell not listening."` to scripts, before any bugs
* (replace $HOST and $PORT by a server you own)
* On your server, listen on $PORT (for instance, `nc -v -l -p $PORT`) and you'll get a shell in the context of the server
* Restart the build if you loose your shell

stephengroat:
I rebased everything and updated llvmlite, cleaned up travis, and added a start of os x builds (allowing for failures).

any comments?

serpilliere:
That's very cool as the travis file is now a lot cleaner. I however have some remarks:
I noticed that you now depends on `sudo`, which hurts me a bit. Is this only needed for the `pip` stuff, or is there anything else which depends on it?
Another point:
I heard that OSX has now some container based feature, which can simplify Miasm installation.
@commial, have you got some references on this?

stephengroat:
@serpilliere No problem
The `sudo` is required for the upgrade to llvm 4.0. The new toolchain hasn't been approved for `travis` in it's `precise` no-sudo system. Since `precise` is about to be deprecated, the `apt-sources` travis approval list for `precise` has been deprecated and won't be updated

commial:
@stephengroat thanks for your work!
Regarding `sudo`, I'm not sure that having llvm 4.0 is a strong requirement for our tests. LLVM is needed for `llvmlite` to work, but the last stable release of `llvmlite` (0.17.0) needs only LLVM 3.9.

In addition, if I correctly understand how Travis is working, not having `sudo` is a benefit in term of slot availability. So for now, I'm voting for the last stable release of `llvmlite`, and then LLVM 3.9, and then no-sudo (but I'm completely open to discussion).

stephengroat:
on the [llvmlite compatibility section](https://github.com/numba/llvmlite#compatibility), it says that LLVM 4.0 is required

> As of version 0.17.0, llvmlite requires LLVM 4.0. It does not support earlier or later versions of LLVM.

serpilliere:
Hi @stephengroat !
This PR is a real plus for the clarity of a lot of things in the project.
I think It must not be closed: we was just waiting for the MaxOS fixes.

But You did a lot of very interesting job here, so I don't ant to lost this.

@commial :  Or we may merge like this even if the MacOS stuff is broken?

commial:
We can likely split it in few PRs, easier to track.
Some of the first MacOS issues, also identified in others PRs / issues could be merged.

I propose to also open a issue dedicated to Mac support tracking, for clarity.

stephengroat:
@serpilliere this just got so far behind that i didn't want to deal with the rebases

@commial i like the splitting idea, let me see what i can come up with.

serpilliere:
No problem @stephengroat : the split sounds good to me as well!

PR #522, opened 2017-04-15
Ir ro

serpilliere:
`AssignBlock` and `IRBlock` can be modified by external sources, but rely on some computed mecanism, which lead to hard problems to debug.

This patch make `IRBlock` them mutable through api.
`AssignBlock` is immutable, but has api to simply manipulate it.

### DIFF
--- a/example/ida/graph_ir.py
+++ b/example/ida/graph_ir.py
@@ -11,7 +11,7 @@ from miasm2.expression.expression import *
 from miasm2.analysis.data_analysis import inter_bloc_flow, \
     intra_bloc_flow_symbexec
 from miasm2.analysis.data_flow import dead_simp
-from miasm2.ir.ir import AssignBlock
+from miasm2.ir.ir import AssignBlock, IRBlock
 
 from utils import guess_machine, expr2colorstr
 
@@ -136,12 +136,14 @@ for block in ab:
 print "IR ok... %x" % ad
 
 for irb in ir_arch.blocks.itervalues():
-    for i, assignblk in enumerate(irb.irs):
+    irs = []
+    for assignblk in irb.irs:
         new_assignblk = {
             expr_simp(dst): expr_simp(src)
             for dst, src in assignblk.iteritems()
         }
-        irb.irs[i] = AssignBlock(new_assignblk, instr=assignblk.instr)
+        irs.append(AssignBlock(new_assignblk, instr=assignblk.instr))
+    ir_arch.blocks[irb.label] = IRBlock(irb.label, irs)
 
 out = ir_arch.graph.dot()
 open(os.path.join(tempfile.gettempdir(), 'graph.dot'), 'wb').write(out)
--- a/example/ida/graph_ir.py
+++ b/example/ida/graph_ir.py
@@ -1,40 +1,55 @@
-import sys
 import os
 import tempfile
 
 from idaapi import GraphViewer
 
 from miasm2.core.bin_stream_ida import bin_stream_ida
-from miasm2.core.asmblock import *
+from miasm2.core.asmblock import expr_is_label, AsmLabel, is_int
 from miasm2.expression.simplifications import expr_simp
-from miasm2.expression.expression import *
-from miasm2.analysis.data_analysis import inter_bloc_flow, \
-    intra_bloc_flow_symbexec
 from miasm2.analysis.data_flow import dead_simp
 from miasm2.ir.ir import AssignBlock, IRBlock
 
 from utils import guess_machine, expr2colorstr
 
 
-def color_irbloc(irbloc):
-    o = []
-    lbl = '%s' % irbloc.label
-    lbl = idaapi.COLSTR(lbl, idaapi.SCOLOR_INSN)
-    o.append(lbl)
-    for assignblk in irbloc.irs:
+# Override Miasm asmblock default label naming convention to shrink block size
+# in IDA
+
+def label_init(self, name="", offset=None):
+    self.fixedblocs = False
+    if is_int(name):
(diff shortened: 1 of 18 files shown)


serpilliere:
Don't merge.

serpilliere:
dont merge

miasm2/ir/ir.py
```diff
@@ -256,15 +258,35 @@ def __init__(self, label, irs):
...
 
-    def _get_dst(self):
-        """Find the IRDst affectation and update dst, dst_linenb accordingly"""
-        if self._dst is not None:
-            return self._dst
+    @property
+    def irs(self):
+        return self._assignments
+
+    def reset_dst(self):
```
commial: I'm not sure to see the point of this API, if IRBlock are immutable...

miasm2/ir/ir.py
```diff
@@ -256,15 +258,35 @@ def __init__(self, label, irs):
...
+        self._dst = None
+        self._dst_linenb = None
+
+    def is_dst_set(self):
+        return self._dst is not None
+
+    def add_assignblock(self, assignblock):
```
commial: A "del" or "remove" is missing to complete the API.
Why not having just an API to create a new IRBlock with a list of assignments?

`AssignBlock` doesn't have these kind of API (add, update, del, ...)

miasm2/ir/ir.py
```diff
@@ -278,25 +300,34 @@ def _get_dst(self):
         self._dst_linenb = final_linenb
         return final_dst
 
-    def _set_dst(self, value):
-        """Find and replace the IRDst affectation's source by @value"""
-        if self._dst_linenb is None:
-            self._get_dst()
+    @property
+    def dst(self):
+        """Find the IRDst affectation and update dst, dst_linenb accordingly"""
```
commial: For me, the docstring is misleading here: this is more a dev comment than the description of what `.dst` is, that is to say the destination of the block as an `Expr` instance (external user does not have to care about cache, especially in the case of immutable object).

miasm2/ir/ir.py
```diff
@@ -278,25 +300,34 @@ def _get_dst(self):
...
-            if dst.is_id("IRDst"):
-                new_assignblk[dst] = value
-                # Sanity check is already done in _get_dst
-                break
-        self._dst = value
-        instr = self.irs[self._dst_linenb].instr
-        self.irs[self._dst_linenb] = AssignBlock(new_assignblk, instr)
-    dst = property(_get_dst, _set_dst)
+    def set_dst(self, value):
+        """Generate a new IRBlock with a dst fixed to @value"""
```
commial: May a warning (in the docstring) saying the previous value of `IRDst`, if any, will be overwritten.

miasm2/ir/ir.py
```diff
@@ -256,15 +256,36 @@ def __init__(self, label, irs):
 
         assert isinstance(label, AsmLabel)
         self.label = label
-        self.irs = irs
+        self._assignments = tuple(irs)
```
commial: May the argument can be renamed to `assignments` too, or `assignblks`

example/ida/graph_ir.py
```diff
@@ -1,40 +1,55 @@
...
+# in IDA
+
+def label_init(self, name="", offset=None):
+    self.fixedblocs = False
+    if is_int(name):
+        name = "loc_%X" % (int(name) & 0xFFFFFFFFFFFFFFFF)
+    self.name = name
+    self.attrib = None
+    if offset is None:
+        self.offset = offset
```
commial: This line is not very clear; `self.offset = None` seems simpler

miasm2/ir/ir.py
serpilliere: Correct: I forgot to remove it

miasm2/ir/ir.py
serpilliere: The only API used in Miasm was this one. After a double check, it is only used in one file. So I am for removing it. What about you?
By the way the `__init__` is exactly the API you describe...

miasm2/ir/ir.py
serpilliere: Correct

miasm2/ir/ir.py
serpilliere: Updated.

miasm2/ir/analysis.py
```diff
@@ -29,44 +28,43 @@ class ira_x86_16(ir_x86_16, ira)
...
-                if n_w in symbols_init and symbols_init[n_w] == v:
+            for n_w in symbexec.symbols:
+                value = symbexec.symbols[n_w]
+                if n_w in symbols_init and symbols_init[n_w] == value:
                     continue
-                eqs.append(ExprAff(n_w, v))
+                eqs.append(ExprAff(n_w, value))
             print '*' * 40
             print irb
             irb.irs = [AssignBlock(eqs)]
```
commial: Here, the `irb.irs` is directly modified (now, it is r-only, right?). But this method seems to never be called... (and could actually have its main part replaced by `symbexec.to_assignblock()`)

miasm2/ir/ir.py
```diff
@@ -256,15 +258,27 @@ def __init__(self, label, irs):
...
-        if self._dst is not None:
-            return self._dst
+    @property
+    def irs(self):
+        return self._assignments
+
+    def is_dst_set(self):
+        return self._dst is not None
+
+    def updt_assignblock(self, index, assignblock):
```
commial: Same remark than before on `add`, `del`, ...
In addition, this API seems to be never used

serpilliere:
Fixed!

PR #534, opened 2017-04-21
DSE: support absence of z3

commial:


serpilliere:
The DSE regression test needs a z3 tag, doesn't it?

commial:
There are 2 tests using the DSE:
* the crackme example, which is based on `DSEPathConstraint` and then needs z3 (and is tagged)
* the regression test, which only use `DSEEngine`, which does not require z3 (and is not tagged)

PR #537, opened 2017-04-24
Fix/ida examples

commial:
Fix IDA examples for #522 + handle a corner case in generated labels in IDA depgraph

### DIFF
--- a/example/ida/depgraph.py
+++ b/example/ida/depgraph.py
@@ -7,6 +7,7 @@ from miasm2.expression import expression as m2_expr
 
 from miasm2.expression.simplifications import expr_simp
 from miasm2.analysis.depgraph import DependencyGraph
+from miasm2.ir.ir import AssignBlock
 
 from utils import guess_machine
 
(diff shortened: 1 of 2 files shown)


example/ida/depgraph.py
```diff
@@ -24,7 +25,11 @@ def __init__(self, ira):
         cur_block = list(ira.getby_offset(self.address))[0]
         for line_nb, l in enumerate(cur_block.lines):
             if l.offset == self.address:
```
serpilliere: duplicate line ?!

commial: This is actually a bad split between commits. I will rebase

example/ida/graph_ir.py
```diff
@@ -133,12 +135,14 @@ def Show(self):
...
-        for dst, src in assignblk.items():
-            del(assignblk[dst])
+for irb in ir_arch.blocks.itervalues():
+    for i, assignblk in enumerate(irb.irs):
+        new_assignblk = dict(assignblk)
+        for dst, src in assignblk.iteritems():
+            del new_assignblk[dst]
             dst, src = expr_simp(dst), expr_simp(src)
-            assignblk[dst] = src
+            new_assignblk[dst] = src
```
serpilliere: Ok, here is remark:
maybe you ripped the pattern from elsewhere, but `new_assignblk` is initialized from the dictionary from the old `assignblk`, but the first action we do in the loop is erasing its keys one by one.
What about creating an empty dictionary, and then adding fresh new keys?

PR #538, opened 2017-04-24
Jitter: Error on reg set not int

serpilliere:
Raise error if setting a jitter register with something else than an integer (or long)

### DIFF
--- a/miasm2/jitter/JitCore.h
+++ b/miasm2/jitter/JitCore.h
@@ -17,7 +17,7 @@
 	}								\
 
 
-#define PyGetInt_ret0(item, value)					\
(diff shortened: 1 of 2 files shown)


commial:
If possible, a non-regression test could be great here

serpilliere:
nope, this is just a sanity  check.

commial:
In others architecture, there is just an "error" printed. For homogeneity, it could be nice to have the same explicit message you have for `x86` for these architectures.
```Python
from miasm2.expression.expression import *

j = Machine("msp430").jitter()
j.cpu.R3 = ExprInt(4, 16)
```
Which outs:
```
error
```

PR #542, opened 2017-04-24
Fix/use seh

commial:
Rename `Sandbox` misleading option `--use-seh` to ` --mimic-windows-structures`, because this is what the option do.
For the same reason, rename `miasm2.os_dep.win_api_x86_32_seh` to  `miasm2.os_dep.win_api_x86_32_structs`, and its main function `init_seh` to `init_win_structs`, because this module implements and this function inits SEH, PEB, TEB, loader, process env, ...


serpilliere:
I agree for the name change, but i don't agree with the new name.
Maybe `win_32_struct_helper`  as we have the `win_32_struct` ?

PR #544, opened 2017-04-24
Fix gcc trace

serpilliere:


miasm2/jitter/codegen.py
```diff
@@ -331,12 +345,18 @@ def gen_pre_code(self, instr_attrib):
...
         out = []
         if attrib.log_regs:
             out.append('dump_gpregs(jitcpu->cpu);')
         return out
 
     def gen_goto_code(self, attrib, instr_offsets, dst):
+        """Generate C code for a potential destination @dst
+        @attrib: instruction Attributes
+        @instr_offsets: instructions offsets list
+        @dst: potention instruction destination"""
```
commial: You probably mean `potential`

PR #545, opened 2017-04-25
Expression: Forbid mix between Expr and int/long

serpilliere:
A classic error in Miasm is to confuse ExprInt with int, for example:
ExprId('a') + 3

We could return ExprOp('+', ExprId('a') + ExprInt(3, 32)), but this
may (and often) mask a problem in the source, so we forbid such a
construction (and assert this to clarify the error)

### DIFF
--- a/miasm2/expression/expression.py
+++ b/miasm2/expression/expression.py
@@ -560,7 +560,9 @@ class ExprAff(Expr):
         @dst: Expr, affectation destination
         @src: Expr, affectation source
         """
-
+        # dst & src must be Expr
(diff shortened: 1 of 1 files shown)


commial:
This could be completed to avoid another common mistake:
```Python
ExprMem(Expr, ExprInt)
```
Or
```Python
ExprSlice(Expr, ExprInt, ExprInt)
```

And enforce that non-Expr arguments are not.

serpilliere:
You are right! Fixed.

PR #547, opened 2017-04-27
arch/x86/jit.py: Update pc as well as EIP in ret_cdecl

carolineLe:


serpilliere:
Correct.
Tanks a lot!

PR #551, opened 2017-05-08
IR: explicit exception for div

serpilliere:


miasm2/arch/aarch64/sem.py
```diff
@@ -526,9 +527,27 @@ def msub(arg1, arg2, arg3, arg4):
     arg1 = arg4 - (arg2 * arg3)
 
 
-@sbuild.parse
-def udiv(arg1, arg2, arg3):
-    arg1 = m2_expr.ExprOp('udiv', arg2, arg3)
+def udiv(ir, instr, arg1, arg2, arg3):
```
commial: Here, you can still use `sbuild.parse`:
```Python
@sbuild.parse
def udiv(arg1, arg2, arg3):
    if arg3:
        arg1 = m2_expr.ExprOp('udiv', arg2, arg3)
    else:
        exception_flags = m2_expr.ExprInt(EXCEPT_INT_DIV_BY_ZERO, exception_flags.size)
```

(which I personally found more readable)
Do not forget to add `exception_flags` to the SemBuilder namespace (L140):
```Python
+        "exception_flags": exception_flags,
```

Other possible form:
```Python
arg1 = 'udiv'(arg2, arg3)
```

miasm2/arch/aarch64/sem.py
```diff
@@ -4,6 +4,7 @@
 from miasm2.arch.aarch64.regs import *
 from miasm2.core.sembuilder import SemBuilder
 
+EXCEPT_INT_DIV_BY_ZERO = (1 << 16)
```
commial: As these constants are relative to Miasm and not to the architecture, we should have them in a single place (which seems to be, for now, `csts.py`)

commial:
To complete this commit, we may add:
* modulo operations (`urem` and others)
* remove `UDIV`, `UMOD`, `IDIV` and `IMOD` C wrapper in jitters, and replace them by corresponding C operation `%`, `/`; idem in LLVM (`builder.udiv`, `builder.urem`, `builder.sdiv`, `builder.srem`)

serpilliere:
I don't agree. We have to keep udiv, umod, ... as we need unsigned _AND_ signed operations in the semantic.
Can you be more precise on this point?

commial:
Actually, I'm talking about the *implementation* of these IR operation in the jitter.
For instance in LLVM, for now, `udiv` is implemented as a call to `udiv64(a, b)`. But with this patch, it can be translated to the corresponding LLVM IR, `builder.udiv`. These way, a call is avoided at runtime, LLVM can do better optimisation, and operations on size not in [16, 32, 64] are automatically supported.

serpilliere:
This commit may resolve remarks. Tell me if everything is ok!

miasm2/arch/aarch64/sem.py
```diff
@@ -691,6 +697,8 @@ def extr(arg1, arg2, arg3, arg4):
     'cset': cset,
     'csetm': csetm,
 
+    'udiv': udiv,
```
commial: These lines are no more needed (already in `build.functions`)

miasm2/jitter/vm_mngr.h
```diff
@@ -224,55 +224,55 @@ unsigned int rcr_rez_op(unsigned int size, unsigned int a, unsigned int b, unsig
...
-		    return 0;						\
-	    }								\
-	    r = a/b;							\
-	    return r;							\
-	    }
+	int ## sizeA ## _t idiv ## sizeA (vm_cpu_t* vmcpu, int ## sizeA ## _t a, int ## sizeA ## _t b) \
+	{								\
+		int ## sizeA ## _t r;					\
+		if (b == 0) {						\
+			printf("Should not append");			\
```
commial: Don't you prefer raise an error and quit? (as a "sanity check")

PR #556, opened 2017-05-09
Convert more operators to C

guedou:
This PR proposes to convert 2 more operators to C. Is it something than could be merged ?

### DIFF
--- a/miasm2/ir/translators/C.py
+++ b/miasm2/ir/translators/C.py
@@ -292,7 +292,7 @@ class TranslatorC(Translator):
                     )
                 return out
 
-            elif expr.op in ['-']:
+            elif expr.op in ['-', '<', '<=']:
                 return '(((%s&%s) %s (%s&%s))&%s)' % (
                     self.from_expr(expr.args[0]),
                     self._size2mask(expr.args[0].size),


commial:
First, thanks for the PR. I'm OK on the principle, but, actually, it raises a few points (not your fault, that's not very clear in Miasm):
* the `-` operator should always be unary. This is not enforce in the `ExprOp` initialization to keep the expressiveness of the language, but the rest of Miasm creates expects in input an unary `-`. This is also the behavior of `Expr.__sub__`, so `a - b -> a + (- b)`. Long story short, this is kind of a typo, and should not be here (unary `-` operator is handled above). So if you want, you can add a commit to delete this support (or I will).
* '<' and '<=' are kind of special. As you probably know, the choice has been made in Miasm to describe side effects on flags with mixed boolean-arithmetic operation. For instance, `x < y (unsigned)` will be obtain through `((x - y) ^ ((x ^ y) & ((x - y) ^ x))).msb()`. But there is moment where it is preferable to have '<' instead of this kind of expression, especially for human readibility. For this usage, there are simplifications that can be activated through `ExpressionSimplifier.PASS_COND` which do exactly this conversion. As a result, one obtain an `ExprOp` with fixed operation names, that are defined in `miasm2.expression.expression`. And here is my point, `<` and `<=` are not part of these operations' names, because they lack sign information. The "supported" corresponding operations are:
```Python
# from miasm2.expression.expression
TOK_INF_SIGNED = "<s"
TOK_INF_UNSIGNED = "<u"
TOK_INF_EQUAL_SIGNED = "<=s"
TOK_INF_EQUAL_UNSIGNED = "<=u"
TOK_EQUAL = "=="
```
As a result, if you want to add a translator for these operations, you should use these names, and then force the C result to act as signed or unsigned comparison.

Do not hesitate if you have any remark, or if I'm not clear enough.

guedou:
I do not need this PR anymore for MeP. 

PR #889 fixes the remaining "issue".

PR #557, opened 2017-05-14
Fix obj

serpilliere:
This PR fixes a design error in C objects representation
It can now parse various structures (EFI, winddk, winbase, ...)

Note: The function ABI declaration is supported, using `__cdecl__` or `__stdcall__`. To have this feature, you need at least pycparser version `516af9479b9bf23569024bf38e8f7cf2681f768c`.

### DIFF
--- a/example/expression/access_c.py
+++ b/example/expression/access_c.py
@@ -127,62 +127,60 @@ class MyCHandler(CHandler):
     exprToAccessC_cls = MyExprToAccessC
 
 
-def test(data):
-    # Digest C informations
-    text = """
-    struct human {
-            unsigned short age;
-            unsigned int height;
-            char name[50];
-    };
 
-    struct ll_human {
-            struct ll_human* next;
-            struct human human;
-    };
-    """
+data = open(sys.argv[1]).read()
+# Digest C informations
+text = """
+struct human {
+        unsigned short age;
+        unsigned int height;
+        char name[50];
+};
 
-    base_types = CTypeAMD64_unk()
-    types_ast = CAstTypes()
-    types_ast.add_c_decl(text)
+struct ll_human {
+        struct ll_human* next;
+        struct human human;
+};
+"""
 
-    types_mngr = CTypesManagerNotPacked(types_ast, base_types)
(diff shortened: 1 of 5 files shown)


example/expression/access_c.py
```diff
@@ -184,5 +185,4 @@ def test(data):
 
 
 if __name__ == '__main__':
```
commial: The "__main__" with only one call snippet is not homogeneous with others examples, and only brings an unnecessary complexity (and indent) to the code

miasm2/core/ctypesmngr.py
```diff
@@ -69,6 +391,64 @@ def add_c_decl_from_ast(self, ast):
...
+        lineno = 1
+
+        # Index to lineno, column
+        for id_start, id_stop, decl in offsets:
+            nbr = c_str.count('\n', index, id_start)
+            lineno += nbr
+            last_cr = c_str.rfind('\n', 0, id_start)
+            # column starts at 1
+            column = id_start - last_cr
+            #print "%-4d %-4d %s" %(lineno, column, c_str[id_start:id_stop])
```
commial: You may want to put these line in a debug logging

miasm2/core/ctypesmngr.py
```diff
@@ -140,236 +511,186 @@ def ast_eval_int(self, ast):
...
-            obj = self.ast_gen_struct_align_size(ast)
-        elif isinstance(ast, c_ast.PtrDecl):
-            void_type = self.void_ptr
-            subobj = self.ast_get_align_size(ast.type)
-            obj = ObjCPtr('noname', subobj, void_type.align, void_type.size)
+    def ast_to_typeid_struct(self, ast):
+        """Return the CTypeBase of an Struct ast"""
+        name = self.gen_uniq_name() if ast.name is None else ast.name
+        args = []
+        if ast.decls:
```
commial: What if the expected value of `ast.decls` if it is not set?
If it is `None`, it would be better to check against `None`. If it is an empty iterable, it is not necessary.

PR #558, opened 2017-05-17
Fix x86 segm

serpilliere:
Sanitize/Fix some x86 semantic opcodes

### DIFF
--- a/test/arch/x86/unit/mn_pushpop.py
+++ b/test/arch/x86/unit/mn_pushpop.py
@@ -39,6 +39,7 @@ class Test_PUSHAD_32(Asm_Test_32):
     '''
 
     def check(self):
+        assert self.myjit.cpu.ESP == self.stk_origin - 0x4 * 8
         buf = self.myjit.vm.get_mem(self.myjit.cpu.ESP, 0x4 * 8)
         assert(buf == self.buf)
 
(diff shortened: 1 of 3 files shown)


test/arch/x86/unit/mn_pushpop.py
```diff
@@ -117,12 +120,202 @@ def test_init(self):
...
+       JMP lbl_ret
+    '''
+
+    def check(self):
+        assert self.myjit.cpu.ESP == self.stk_origin + 0x4
+        assert self.myjit.cpu.EAX == self.value
+
+
+class Test_POP_mode32_16(Asm_Test_32):
+    MYSTRING = "test pop mode32 32"
```
commial: It looks like it is "mode 32 16"

test/arch/x86/unit/mn_pushpop.py
```diff
@@ -117,12 +120,202 @@ def test_init(self):
...
+       JMP lbl_ret
+    '''
+
+    def check(self):
+        assert self.myjit.cpu.ESP == self.stk_origin + 0x2
+        assert self.myjit.cpu.AX == self.value
+
+
+class Test_POP_mode16_32(Asm_Test_16):
+    MYSTRING = "test pop mode16 16"
```
commial: And here, mode16 32

commial:
Nice spot!

test/arch/x86/unit/mn_pushpop.py
serpilliere: Ho yes. Thanks

test/arch/x86/unit/mn_pushpop.py
serpilliere: Correct

PR #567, opened 2017-06-07
Exec engine

serpilliere:
Add C type propagation engine
Add corresponding IDA plugin

### DIFF
--- a/miasm2/core/ctypesmngr.py
+++ b/miasm2/core/ctypesmngr.py
@@ -8,14 +8,16 @@ RE_HASH_CMT = re.compile(r'^#\s*\d+.*$', flags=re.MULTILINE)
 # http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf
 
 
-def c_to_ast(c_str):
+def c_to_ast(parser, c_str):
     """Transform a @c_str into a C ast
     Note: will ignore lines containing code refs ie:
     # 23 "miasm.h"
+
+    @parser: pycparser instance
+    @c_str: c string
     """
 
     new_str = re.sub(RE_HASH_CMT, "", c_str)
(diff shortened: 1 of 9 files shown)


miasm2/core/ctypesmngr.py
```diff
@@ -8,14 +8,16 @@
 # http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf
 
 
-def c_to_ast(c_str):
+def c_to_ast(parser, c_str):
     """Transform a @c_str into a C ast
     Note: will ignore lines containing code refs ie:
     # 23 "miasm.h"
+
+    @parer: pycparser instance
```
commial: Typo

miasm2/ir/symbexec.py
```diff
@@ -442,6 +442,7 @@ def emulbloc(self, irb, step=False):
         """
         for assignblk in irb.irs:
             if step:
+                print 'Instr', assignblk.instr
```
commial: `.instr` is not necessarily defined. Do you prefer to display a `Instr: None`, or nothing?

serpilliere: I don't agree with you.
It is *always* defined. But It can be *None* actually. See the init code:
```python

class AssignBlock(object):
   ...
    __slots__ = ["_assigns", "_instr"]

    def __init__(self, irs=None, instr=None):
        ...
        if irs is None:
            irs = []
        self._instr = instr
        self._assigns = {} # ExprAff.dst -> ExprAff.src

```

So it will print `Instr None` which is acceptable in my opinion.

miasm2/core/ctypesmngr.py
serpilliere: Correct!

PR #577, opened 2017-07-06
Display expr

serpilliere:
```
@64[(RSP+0x8)] = RBX

@64[(RSP+0x18)] = RSI

RSP = (RSP+(- 0x8))
@64[(RSP+(- 0x8))] = RDI

af = ((RSP^0x30)^(RSP+(- 0x30)))[4:5]
pf = parity(((RSP+(- 0x30))&0xFF))

```

is replaced by:

```
@64[RSP+0x8] = RBX

@64[RSP+0x18] = RSI

RSP = RSP+(-0x8)
@64[RSP+(-0x8)] = RDI

af = ((RSP^0x30)^(RSP+(-0x30)))[4:5]
pf = parity((RSP+(-0x30))&0xFF)
```

### DIFF
--- a/example/ida/utils.py
+++ b/example/ida/utils.py
@@ -3,7 +3,8 @@ from idc import *
 
 from miasm2.analysis.machine import Machine
 from miasm2.ir.translators import Translator
-import miasm2.expression.expression as m2_expr
+from miasm2.expression.expression import ExprInt, ExprId, ExprSlice, ExprMem, \
+    ExprCond, ExprCompose, ExprOp, ExprAff
 
 def max_size_to_size(max_size):
     for size in [16, 32, 64]:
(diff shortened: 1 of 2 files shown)


example/ida/utils.py
```diff
@@ -66,61 +67,85 @@ def guess_machine():
...
+    return new_expr
 
-    return s
+
+def expr2colorstr(regs_ids, expr):
+    """Colorize an Expr instance for IDA
+    @regs_ids: list of ExprId corresponding to available registers
+    @expr: Expr instance to colorize
+    """
+    return expr2exprida(regs_ids, expr)
```
commial: What is the interest of separating these two function?

miasm2/expression/expression.py
```diff
@@ -927,20 +952,33 @@ def __new__(cls, op, *args):
         return Expr.get_object(cls, (op, args))
 
     def __str__(self):
-        if self.is_associative():
-            return '(' + self.__op.join([str(arg) for arg in self.__args]) + ')'
-        if (self.__op.startswith('call_func_') or
-            self.__op == 'cpuid' or
-            len(self.__args) > 2 or
-                self.__op in ['parity', 'segm']):
+        #if self.is_associative():
```
commial: git will conserve this code for you :)

miasm2/expression/expression.py
```diff
@@ -511,6 +522,9 @@ def __new__(cls, name, size=32):
     def __str__(self):
         return str(self.__name)
 
+    def str_sub_expr(self):
```
commial: This method can be declared in Expr, and then there is no need to repeat it if there is no change

commial: Also, I'm not sure if this is more a Miasm internal, and we probably don't want to expose it in the public API. So maybe a '_str_sub_expr' could be fine.

miasm2/expression/expression.py
```diff
@@ -1136,7 +1177,11 @@ def __new__(cls, *args):
         return Expr.get_object(cls, args)
 
     def __str__(self):
-        return '{' + ', '.join(["%s %s %s" % (arg, idx, idx + arg.size) for idx, arg in self.iter_args()]) + '}'
+        return '{' + ', '.join(["%s %s %s" % (arg, idx, idx + arg.size) for
```
commial: The `[` in `([` is useless here; it wil create an array and then iter over it, instead of an iterator

miasm2/expression/expression.py
```diff
@@ -69,6 +69,14 @@ def wrapped(expr, callback, test_visit=lambda x: True):
     return wrapped
 
 
+DISPLAYED_CHARS = set('abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0X123456789_')
```
commial: `X` is present two times

miasm2/expression/expression.py
```diff
@@ -69,6 +69,14 @@ def wrapped(expr, callback, test_visit=lambda x: True):
...
+
+def is_prefixed_opname(name):
```
commial: From what the function seems to do, the name is quite misleading. It is more about the charset used than a notion of prefix/suffix, no?

example/ida/utils.py
serpilliere: I will keep the ExprIda generation and the color generation in two different function (I have updated the code)

miasm2/expression/expression.py
serpilliere: Erf. I removed the dummy code

miasm2/expression/expression.py
serpilliere: Correct

miasm2/expression/expression.py
serpilliere: Done

miasm2/expression/expression.py
serpilliere: Modified

miasm2/expression/expression.py
serpilliere: done

PR #579, opened 2017-07-07
Fix get bloc

serpilliere:


miasm2/core/asmblock.py
```diff
@@ -1519,44 +1519,61 @@ def _dis_bloc(self, offset):
...
                     target_offset in self.job_done):
                 continue
-            cur_block, nexts = self._dis_bloc(target_offset)
+            cur_block, nexts = self._dis_block(target_offset)
             todo += nexts
-            blocs.add_node(cur_block)
+            blocks.add_node(cur_block)
+
+        blocks.apply_splitting(self.symbol_pool,
+                               dis_block_callback=self.dis_bloc_callback,
```
fmonjalet: There is still a `dis_bloc_callback` attribute here instead of `dis_block_callback`. For backward compatibility, you can make  `dis_bloc_callback` a property the wraps the new one with a warning.

serpilliere: Yes, a lot of them!
But It's a work which is done bloc by bloc.
I mean block by block

PR #582, opened 2017-07-21
Add null pointer check after realloc

WilliamBruneau:


serpilliere:
Thanks!
Can you dilute merge the coding style compliance patch with its corresponding patches?

PR #587, opened 2017-07-21
x86: fix MOVSD semantic

commial:


miasm2/arch/x86/sem.py
```diff
@@ -1777,9 +1777,11 @@ def movs(ir, instr, size):
 def movsd(_, instr, dst, src):
     e = []
     if isinstance(dst, m2_expr.ExprId) and isinstance(src, m2_expr.ExprMem):
-        src = m2_expr.ExprMem(src.arg, dst.size)
+        src = m2_expr.ExprMem(src.arg, 64)
+        dst = dst[:64]
```
serpilliere: I have a doubt on the behavior of the mnemonic:
when you have a mov from a *memory* to a xmm register, the high part seems to be erased:
```
MOVSD (128-bit Legacy SSE version: MOVSD XMM1, m64)
DEST[63:0]  SRC[63:0]
DEST[127:64]  0
DEST[VLMAX-1:128] (Unmodified)
```

(from the Intel® 64 and IA-32 Architectures Software Developer’s Manual, page 3-588)

commial: You're completely right, I've just tested on a real CPU

I'll update the PR accordingly

PR #591, opened 2017-07-27
Asmblock: remove disasm engine job_done attribute

serpilliere:
WARNING: disasmEngine behaviour modification

Before patch: job_done containted the already disassembled addresses. If the
user disassembled twice the same addresse, the engine will return empty
object on the second call.

After patch: If the user disassemble twice the same addresse, the engine will
return result of the disassembling in both cases.

### DIFF
--- a/miasm2/core/asmblock.py
+++ b/miasm2/core/asmblock.py
@@ -1351,12 +1351,6 @@ class disasmEngine(object):
     + callback(arch, attrib, pool_bin, cur_bloc, offsets_to_dis,
                symbol_pool)
      - dis_bloc_callback: callback after each new disassembled block
-
-    The engine also tracks already handled block, for performance and to avoid
-    infinite cycling.
-    Addresses of disassembled block is in the attribute `job_done`.
-    To force a new disassembly, the targeted offset must first be removed from
-    this structure.
     """
 
     def __init__(self, arch, attrib, bin_stream, **kwargs):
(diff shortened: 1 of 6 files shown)


miasm2/core/asmblock.py
```diff
@@ -1386,11 +1400,16 @@ def __init__(self, arch, attrib, bin_stream, **kwargs):
         # Override options if needed
         self.__dict__.update(kwargs)
 
-    def _dis_block(self, offset):
+        # Deprecated
+        self.job_done = DeprecatedJobDone()
+
+    def _dis_block(self, offset, job_done=None):
```
commial: Can you add the doc  for the new argument?

miasm2/core/asmblock.py
```diff
@@ -1325,6 +1325,27 @@ def asm_resolve_final(mnemo, blocks, symbol_pool, dst_interval=None):
     return patches
 
 
+class DeprecatedJobDone(object):
```
commial: Why don't you use a property, with a getter and a setter? (returning an actual set) 
If a user is doing a `mis.job_done = set() `, he will not see the warning

serpilliere: Correct!
thanks

miasm2/core/asmblock.py
serpilliere: Done!

PR #593, opened 2017-08-02
fix atomic_mode leak

eqv:
Fix missing relase on _atomic_mode that leads to a crash if getbits fails because no content is available


serpilliere:
Nice catch!
Thanks

PR #594, opened 2017-08-03
Refactor/expression protected attribs

commial:
Move `__attribute` uses in `Expr`s to `_attribute` ones

### DIFF
--- a/miasm2/expression/expression.py
+++ b/miasm2/expression/expression.py
@@ -125,9 +125,12 @@ class Expr(object):
     def set_size(self, _):
         raise ValueError('size is not mutable')
 
-    def __init__(self):
+    def __init__(self, size):
+        """Instanciate an Expr with size @size
+        @size: int
+        """
         # Common attribute
-        self._size = None
+        self._size = size
 
         # Lazy cache needs
         self._hash = None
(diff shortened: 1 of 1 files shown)


miasm2/expression/expression.py
```diff
@@ -390,7 +393,8 @@ def __init__(self, arg, size):
         @arg: 'intable' number
         @size: int size"""
         super(ExprInt, self).__init__()
-        # Work is done in __new__
+        # Work for ._arg is done in __new__
```
serpilliere: I think it's a bit confusion to set the `._arg` in the `__new__`, and the `._size` `__init__`.
Why such a code split?

serpilliere: Ok Got it

miasm2/expression/expression.py
```diff
@@ -1116,10 +1109,9 @@ def __init__(self, *args):
         # args must be Expr
         assert all(isinstance(arg, Expr) for arg in args)
 
-        super(ExprCompose, self).__init__()
         assert isinstance(args, tuple)
         self._args = args
-        self._size = sum([arg.size for arg in args])
+        super(ExprCompose, self).__init__(sum(arg.size for arg in args))
 
     args = property(lambda self: self._args)
```
serpilliere: I agree that you call `super __init__` for expr, but I don't see the gain to move the size `__init__` in it (erf).
What is the goal of making Expr instanciable?

PR #605, opened 2017-08-15
Clean some win api

serpilliere:


miasm2/os_dep/win_api_x86_32.py
```diff
@@ -73,7 +76,7 @@
                0x100: 0
                }
 
-access_dict_inv = dict([(x[1], x[0]) for x in access_dict.items()])
+ACCESS_DICT_INV = dict([(x[1], x[0]) for x in ACCESS_DICT.items()])
```
p-l-: You can save two list creations here:
```Python
ACCESS_DICT_INV = dict((x[1], x[0]) for x in ACCESS_DICT.iteritems())
```

serpilliere: Done. Thanks!

PR #607, opened 2017-08-15
Mono breakpoint

serpilliere:


miasm2/jitter/jitload.py
```diff
@@ -257,30 +258,47 @@ def exception_memory_breakpoint(jitter):
...
+    def set_breakpoint(self, addr, callback):
+        """Add a berakpont at addr to @callback.
+        WARNING: an address can have only one callback
         @addr: breakpoint address
         @callback: function with definition (jitter instance)
         """
-        self.breakpoints_handler.add_callback(addr, callback)
+
+        if addr in self.breakpoints_handlers:
+            raise ValueError("A breakpoint is already set at %x" % addr)
```
commial: It could be great for the user to have a pattern to actually set a breakpoint even if it already exists, or at least easily detect it.
Maybe:
* A custom error `BreakpointAlreadyExist` instead of the generic `ValueError`, which may be raised from other code part
* A `force` option, by default to False, which force the breakpoint update. It will avoid the user to directly manipulate `.breakpoints_handlers` internal structure by forcing its changes through a dedicated API (BTW, this internal structure could become protected)

miasm2/jitter/jitload.py
```diff
@@ -310,14 +328,16 @@ def runiter_once(self, pc):
 
         # Check breakpoints
         old_pc = self.pc
-        for res in self.breakpoints_handler.call_callbacks(self.pc, self):
-            if res is not True:
-                if isinstance(res, collections.Iterator):
+        callback = self.breakpoints_handlers.get(self.pc, None)
+        if callback:
```
commial: An explicit pointer test against `None` is faster to detect the dict miss (avoid testing if a function is equal to 0)

miasm2/jitter/jitload.py
```diff
@@ -449,7 +469,8 @@ def handle_lib(jitter):
 
     def handle_function(self, f_addr):
         """Add a brakpoint which will trigger the function handler"""
```
commial: Typo "brEakpoint" (not added by this commit, but related)

miasm2/jitter/jitload.py
```diff
@@ -449,7 +469,8 @@ def handle_lib(jitter):
...
-        self.add_breakpoint(f_addr, self.handle_lib)
+        if self.get_breakpoint(f_addr) is None:
```
commial: Here, if a breakpoint is already set, this method will silently failed, which is probably misleading for the user and hardly debuggable after

miasm2/analysis/sandbox.py
```diff
@@ -100,13 +100,17 @@ def parser(cls, *args, **kwargs):
...
-    def run(self, addr=None):
+    def get_start_addr(self, addr=None):
+        """If addr is not set, use arguments addr"""
+        if addr is None and self.options.address is not None:
+            addr = int(self.options.address, 0)
+        return addr
+
+    def prepare_exec(self, addr):
         """
         Launch emulation (gdbserver, debugging, basic JIT).
```
commial: This comment is no longer correct ("Launch")

miasm2/analysis/sandbox.py
```diff
@@ -118,16 +122,35 @@ def run(self, addr=None):
...
             else:
                 cmd = debugging.DebugCmd(dbg)
                 self.cmd = cmd
-                cmd.cmdloop()
 
         else:
             self.jitter.init_run(addr)
+
+    def run_loop(self):
+        if any([self.options.debugging, self.options.gdbserver]):
```
commial: What about:
```Python
if self.options.gdbserver:
    gdb.run()
elif self.options.debugging:
    cmd.cmdloop()
else:
    self.jitter.continue_run()
```

miasm2/analysis/sandbox.py
```diff
@@ -118,16 +122,35 @@ def run(self, addr=None):
...
+            if self.options.gdbserver:
+                gdb.run()
```
commial: It's probably `self.gdb` and `self.cmd`

miasm2/jitter/jitload.py
```diff
@@ -320,60 +330,41 @@ def runiter_once(self, pc):
...
+            return_value &= result
+
+        # Return if an exception handler stopped
+        if not return_value:
+            return return_value
 
         # If a callback changed pc, re call every callback
         if old_pc != self.pc:
-            return
+            return return_value
```
commial: Here, `return_value` is always `True`

PR #610, opened 2017-09-01
X86 add instr

serpilliere:


test/expression/expr_cmp.py
```diff
@@ -0,0 +1,68 @@
...
+    expr_is_signed_lower_or_equal, expr_is_equal, expr_is_not_equal
+
+
+int_0 = ExprInt(0, 32)
+int_1 = ExprInt(1, 32)
+int_m1 = ExprInt(-1, 32)
+int_m2 = ExprInt(-2, 32)
+
+# above below
+assert expr_is_unsigned_greater(int_1, int_0)
```
commial: Actually, you're asserting an `Expr` instance, which will never fail
You probably want to apply a `expr_simp` and check against `int_0`?

serpilliere: Fixed!

PR #613, opened 2017-09-06
Fix z3 random seed for test reproductibility

commial:


commial:
Wrong cause, wrong fix

PR #616, opened 2017-09-09
Various Win32 API additions/fixes

aguinet:
This PR contains various Win32 API additions/fixes:
* add a get_size method to Miasm heap object, which allows the
implementation of mscvrt_realloc
* add the concept of "current directory", with the default value being
arbitrary set to "c:\tmp", which allows the implementation of
{Get,Set}CurrentDirecrtory

Various other methods implemented:

 - advapi32_RegCloseKey
 - advapi32_RegCreateKeyW
 - advapi32_RegSetValueExA
 - advapi32_RegSetValueExW
 - kernel32_GetProcessHeap
 - msvcrt_delete
 - msvcrt_fprintf
 - msvcrt_fwrite
 - msvcrt__mbscpy
 - msvcrt_new
 - msvcrt_realloc
 - msvcrt_sprintf
 - msvcrt_srand
 - msvcrt_strrchr
 - msvcrt_swprintf
 - msvcrt_wcscat
 - msvcrt_wcscmp
 - msvcrt_wcscpy
 - msvcrt__wcsicmp
 - msvcrt_wcslen
 - msvcrt_wcsncpy
 - msvcrt__wcsnicmp
 - msvcrt_wcsrchr

### DIFF
--- a/miasm2/os_dep/common.py
+++ b/miasm2/os_dep/common.py
@@ -2,6 +2,7 @@ import os
 
 from miasm2.jitter.csts import PAGE_READ, PAGE_WRITE
 from miasm2.core.utils import get_caller_name
+from miasm2.core.utils import pck64, upck64
 
 BASE_SB_PATH = "file_sb"
 
@@ -73,9 +74,12 @@ class heap(object):
             combination of them); default is PAGE_READ|PAGE_WRITE
         """
         addr = self.next_addr(size)
-        vm.add_memory_page(addr, perm, "\x00" * size,
+        vm.add_memory_page(addr, perm, "\x00" * (size),
                            "Heap alloc by %s" % get_caller_name(2))
         return addr
+
+    def get_size(self, vm, ptr):
+        return vm.get_all_memory()[ptr]["size"]
 
 
 def windows_to_sbpath(path):
@@ -94,3 +98,33 @@ def unix_to_sbpath(path):
     """
     path = [elt for elt in path.split('/') if elt]
     return os.path.join(BASE_SB_PATH, *path)
+
+def get_fmt_args(fmt, cur_arg, get_str, get_arg_n):
+    output = ""
+    idx = 0
+    fmt = get_str(fmt)
+    while True:
+        if idx == len(fmt):
+            break
+        char = fmt[idx]
+        idx += 1
+        if char == '%':
+            token = '%'
+            while True:
+                char = fmt[idx]
+                idx += 1
+                token += char
+                if char.lower() in '%cdfsux':
+                    break
+            if char == '%':
+                output += char
+                continue
+            if token.endswith('s'):
+                addr = get_arg_n(cur_arg)
+                arg = get_str(addr)
+            else:
+                arg = get_arg_n(cur_arg)
+            char = token % arg
+            cur_arg += 1
+        output += char
+    return output
--- a/miasm2/os_dep/common.py
+++ b/miasm2/os_dep/common.py
@@ -79,7 +79,25 @@ class heap(object):
         return addr
 
     def get_size(self, vm, ptr):
-        return vm.get_all_memory()[ptr]["size"]
+        """
+        @vm: a VmMngr instance
+        @size: ptr to get the size of the associated allocation.
+        
+        `ptr` can be the base address of a previous allocation, or an address
+        within the allocated range. The size of the whole allocation is always
+        returned, regardless ptr is the base address or not.
+        """
+	assert vm.is_mapped(ptr, 1)
+	data = vm.get_all_memory()
+	ptr_page = data.get(ptr, None)
+	if ptr_page is None:
+	    for address, page_info in data.iteritems():
+		if address <= ptr < address + page_info["size"]:
+		    ptr_page = page_info
+		    break
+	    else:
+		raise RuntimeError("Must never happen (unmapped but mark as mapped by API)")
+	return ptr_page["size"]
 
 
 def windows_to_sbpath(path):
--- a/miasm2/os_dep/win_api_x86_32.py
+++ b/miasm2/os_dep/win_api_x86_32.py
@@ -32,9 +32,10 @@ except ImportError:
     print "cannot find crypto, skipping"
 
 from miasm2.jitter.csts import PAGE_READ, PAGE_WRITE, PAGE_EXEC
-from miasm2.core.utils import pck16, pck32, upck32, hexdump, whoami
+from miasm2.core.utils import pck16, pck32, upck16, upck32, hexdump, whoami
 from miasm2.os_dep.common import heap, windows_to_sbpath
 from miasm2.os_dep.common import set_str_unic, set_str_ansi
+from miasm2.os_dep.common import get_fmt_args as _get_fmt_args
 from miasm2.os_dep.win_api_x86_32_seh import tib_address
 
 log = logging.getLogger("win_api_x86_32")
@@ -160,7 +161,8 @@ class c_winobjs:
         self.tls_values = {}
         self.handle_pool = handle_generator()
         self.handle_mapped = {}
-        self.hkey_handles = {0x80000001: "hkey_current_user"}
+        self.hkey_handles = {0x80000001: "hkey_current_user", 0x80000002: "hkey_local_machine"}
+        self.cur_dir = "c:\\tmp"
 
         self.nt_mdl = {}
         self.nt_mdl_ad = None
@@ -278,6 +280,18 @@ def kernel32_LocalAlloc(jitter):
     alloc_addr = winobjs.heap.alloc(jitter, args.msize)
     jitter.func_ret_stdcall(ret_ad, alloc_addr)
 
+def msvcrt_new(jitter):
+    ret_ad, args = jitter.func_args_cdecl(["size"])
+    alloc_addr = winobjs.heap.alloc(jitter, args.size)
+    jitter.func_ret_cdecl(ret_ad, alloc_addr)
+
+globals()['msvcrt_??2@YAPAXI@Z'] = msvcrt_new
+
+def msvcrt_delete(jitter):
+    ret_ad, args = jitter.func_args_cdecl(["ptr"])
+    jitter.func_ret_cdecl(ret_ad, 0)
+
+globals()['msvcrt_??3@YAXPAX@Z'] = msvcrt_delete
 
 def kernel32_GlobalFree(jitter):
     ret_ad, _ = jitter.func_args_stdcall(["addr"])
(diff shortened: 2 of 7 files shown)


miasm2/os_dep/common.py
```diff
@@ -72,10 +73,14 @@ def vm_alloc(self, vm, size, perm=PAGE_READ | PAGE_WRITE):
         @perm: permission flags (PAGE_READ, PAGE_WRITE, PAGE_EXEC or any `|`
             combination of them); default is PAGE_READ|PAGE_WRITE
         """
-        addr = self.next_addr(size)
-        vm.add_memory_page(addr, perm, "\x00" * size,
+        addr = self.next_addr(size+8)
+        vm.add_memory_page(addr, perm, "\x00" * (size+8),
```
commial: Mixing a meta-info such as the page size with the data is probably not a good idea.
This is more a "structural" information, such as memory access rights, and it is already save in the implementation of the memory pages.
For now, there is no direct API to get it back, but you can obtain it by using `vm.get_all_memory()[start_addr]["size"]`.
An API like the one with mem access is maybe missing.

miasm2/os_dep/win_api_x86_32.py
```diff
@@ -1411,6 +1456,16 @@ def kernel32_lstrcpyA(jitter):
 def kernel32_lstrcpy(jitter):
     my_strcpy(jitter, whoami(), jitter.get_str_ansi, jitter.set_str_ansi)
 
+def msvcrt__mbscpy(jitter):
+    ret_ad, args = jitter.func_args_cdecl(["ptr_str1", "ptr_str2"])
+    s2 = jitter.get_str_unic(args.ptr_str2)
+    print("mbscpy '%s'" % s2)
```
commial: You have made an effort to use logging facilities everywhere else, too bad to use a `print` here

miasm2/os_dep/win_api_x86_32.py
```diff
@@ -1963,6 +2044,72 @@ def user32_IsCharAlphaNumericA(jitter):
     jitter.func_ret_stdcall(ret_ad, ret)
 
 
+def get_fmt_args(jitter, fmt, cur_arg, get_str, char_size=1):
```
commial: Instead of copying `get_fmt_args` from `linux_stdlib`, it would be probably better to have them in common in `os_dep/common.py`; this way, they will both benefit from future update and fix...

miasm2/os_dep/win_api_x86_32.py
aguinet: It's because I missed this one :)

miasm2/os_dep/win_api_x86_32.py
aguinet: Right, this is an extract from a previous "work session" with miasm. I forgot that I copy-pasted it from `linux_stdlib`

miasm2/os_dep/win_api_x86_32.py
```diff
@@ -1963,6 +2044,72 @@ def user32_IsCharAlphaNumericA(jitter):
...
+    return jitter.func_ret_cdecl(ret_ad, ret)
+
+def msvcrt_fprintf(jitter):
+    ret_addr, args = jitter.func_args_cdecl(['file', 'fmt'])
+    cur_arg, fmt = 2, args.fmt
+    output = get_fmt_args(jitter, fmt, cur_arg)
+    ret = len(output)
+    log.info("fprintf(%x, '%s') = '%s'" % (args.file, jitter.get_str_ansi(args.fmt), output))
+
+    fd = upck32(jitter.vm.get_mem(args.file + 0x10, 4))
```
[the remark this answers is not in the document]
aguinet: There is `x86_32` in the filename :) ?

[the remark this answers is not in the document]
commial: Yeah, it is misleading. But for now, API implementation are shared between x86_32 and x86_64, thanks to "abstraction" like calling conventions wrapper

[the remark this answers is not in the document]
aguinet: It's actually done in many other places like that :)

miasm2/os_dep/common.py
aguinet: The API you give is linked to memory pages, and not the heap that is used here no? This would mean that the size returned by ``vm.get_all_memory()[start_addr]["size"]`` would be the size of the page where the heap resides, but not the size of the allocated heap chunk.

aguinet:
@commial I fixed almost everything except the 32-bit FD "issue" here https://github.com/cea-sec/miasm/pull/616/commits/789a4a937280b07fdbd6855cc7208c2b24c4f56c , and added some related tests :)

aguinet:
If that's fine for you I'll squash this!

miasm2/os_dep/win_api_x86_32.py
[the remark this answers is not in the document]
commial: Ok for me

miasm2/os_dep/common.py
```diff
@@ -73,14 +73,13 @@ def vm_alloc(self, vm, size, perm=PAGE_READ | PAGE_WRITE):
         @perm: permission flags (PAGE_READ, PAGE_WRITE, PAGE_EXEC or any `|`
             combination of them); default is PAGE_READ|PAGE_WRITE
         """
-        addr = self.next_addr(size+8)
-        vm.add_memory_page(addr, perm, "\x00" * (size+8),
+        addr = self.next_addr(size)
+        vm.add_memory_page(addr, perm, "\x00" * (size),
```
commial: Tiny remark: you left your parenthesis around `(size)`, which lead to a diff on this function for the whole PR, where there is no need to

aguinet:
@commial I removed the "bad" diff, squashed everything into one commit and rebased on master. I also added a test for ``msvcrt_realloc`` (so that ``heap.get_size`` is tested as a side effect), and also removed ``kernel32_GetProcessHeap`` which was a "return 0xdeadbeef"-only implementation..

miasm2/os_dep/common.py
```diff
@@ -73,14 +73,13 @@ def vm_alloc(self, vm, size, perm=PAGE_READ | PAGE_WRITE):
...
                            "Heap alloc by %s" % get_caller_name(2))
-        vm.set_mem(addr, pck64(size))
-        return addr+8
+        return addr
 
     def get_size(self, vm, ptr):
-        return upck64(vm.get_mem(ptr-8, 8))
+        return vm.get_all_memory()[ptr]["size"]
```
commial: As you add a new public function, can you add a docstring describing it?
I know this is not the case for every function in this module, but we (at least try to) add docstring on new implemented function. Thus, Miasm readibility gets better on average :smiley: 

For this function, there is a particularity: `ptr` must be the base address of the mapped function. If you want to have a more permissive function, you can do something like (untested but should work):
```Python
assert vm.is_mapped(ptr, 1)
data = vm.get_all_memory()
ptr_page = data.get(ptr, None)
if ptr_page is None:
    for address, page_info in data.iteritems():
        if address <= ptr < address + page_info["size"]:
            ptr_page = page_info
            break
    else:
        raise RuntimeError("Must never happen (unmapped but mark as mapped by API)")
return ptr_page["size"]
```
Note for a future PR: this function should probably reside in VmMngr implementation

miasm2/os_dep/win_api_x86_32.py
```diff
@@ -1793,13 +1848,39 @@ def msvcrt_memset(jitter):
...
+    s = jitter.get_str_unic(args.pstr)
+    c = chr(args.c)
+    ret = args.pstr + (s.rfind(c)*2)
+    log.info("wcsrchr(%x '%s',%s) = %x" % (args.pstr,s,c,ret))
+    jitter.func_ret_cdecl(ret_ad, ret)
 
 def msvcrt_memcpy(jitter):
     ret_ad, args = jitter.func_args_cdecl(['dst', 'src', 'size'])
     s = jitter.vm.get_mem(args.src, args.size)
+    #log.info("memcpy buf %s" % s.encode("hex"))
```
commial: You still have a commented log here, you can uncomment it

miasm2/os_dep/common.py
```diff
@@ -73,10 +74,13 @@ def vm_alloc(self, vm, size, perm=PAGE_READ | PAGE_WRITE):
             combination of them); default is PAGE_READ|PAGE_WRITE
         """
         addr = self.next_addr(size)
-        vm.add_memory_page(addr, perm, "\x00" * size,
+        vm.add_memory_page(addr, perm, "\x00" * (size),
                            "Heap alloc by %s" % get_caller_name(2))
         return addr
 
+    def get_size(self, vm, ptr):
+        return vm.get_all_memory()[ptr]["size"]
```
commial: As you add a new public function, can you add a docstring describing it?
I know this is not the case for every function in this module, but we (at least try to) add docstring on new implemented function. Thus, Miasm readibility gets better on average :smiley: 

For this function, there is a particularity: `ptr` must be the base address of the mapped function. If you want to have a more permissive function, you can do something like (untested but should work):
```Python
assert vm.is_mapped(ptr, 1)
data = vm.get_all_memory()
ptr_page = data.get(ptr, None)
if ptr_page is None:
    for address, page_info in data.iteritems():
        if address <= ptr < address + page_info["size"]:
            ptr_page = page_info
            break
    else:
        raise RuntimeError("Must never happen (unmapped but mark as mapped by API)")
return ptr_page["size"]
```
Note for a future PR: this function should probably reside in VmMngr implementation

commial:
Good job! I only have a few remarks left and it will be OK for me.

One last comment: adding test is a very good thing for the project's health, but when you add new test file, they must be tracked (to be launched) by `test/test_all.py`.
So you can add [here](https://github.com/cea-sec/miasm/blob/master/test/test_all.py#L269): `os.path.join("linux", "stdlib.py"),` and travis will launch it :+1:

miasm2/os_dep/win_api_x86_32.py
aguinet: Or just remove the line :) ?

miasm2/os_dep/common.py
aguinet: I think the semantic of the function should be indeed more clear. About what you propose, I think we should actually make 2 functions : ``get_page_size``, which asserts that the ``ptr`` is the start address of a page, and a ``get_alloc_size``, which does what you do and returns the number of available bytes starting from ``ptr`` (so that's actually a bit different from your implementation, but as long as it's properly documented).  What do you think?

aguinet:
Okay I'll do that :)

miasm2/os_dep/win_api_x86_32.py
commial: Your call

miasm2/os_dep/common.py
commial: It could be confusing (between `get_page_size` / `get_alloc_size`). In addition, we do not have any use of a `get_alloc_size` as you describe for now.
So I'll prefer a unique `get_page_size` API. If you find it confusing to be able to get the allocated page size for a address in the middle of the page, drop my comment but make it clear in the docstring (something like "@ptr: starting address of the target page")

aguinet: Hmm indeed the usage of ``get_alloc_size`` is really limited. Your proposal is fine to me as long as it's also stated in the documentation. I'll do it this way thus :)

aguinet:
@commial I made a commit with the latest remarks. Let me know if that's okay! (and I'll squash again :))

commial:
Thanks again for this contrib!

(I've squashed the 2 commits ;))

aguinet:
Thanks :) !

Issue #619, opened 2017-09-26
[Question] Dynamic architecture registration


guedou:
I would like to dynamically add a new architecture into `Machine`, and found [this comment](https://github.com/cea-sec/miasm/blob/master/miasm2/analysis/machine.py#L31) while reading the source code.

Do you think that dynamic registration can be implemented, or that's a dead end due to performance issues ?

serpilliere:
Yes! That's interesting. I will see with @commial some details on the way to implement this: In some Miasm parts, we used an "AbstractMethod" raising in order to enforce the sub class implementations, something like:
```python
raise NotImplementedError("Abstract")
```

guedou:
I can implement it if you tell what to do =)

commial:
Hi @guedou,

Two things here:
* the comment you link to is related to the fact that, indeed, importing big architecture can result in quite long delay. Indeed, Miasm will create new classes for each instruction, build some internal structures (the parsing dictionary), parse a lot of function (semantics, IR, jit, ...). On my computer, importing `x86*` or `arm*` takes more than 1 sec. So, this code has been done to avoid a 3/4 seconds delay on Miasm launch, even if the architecture aren't all needed.
* the `Machine` object is not yet well designed to be extended, and this is probably a fault of mine. A dynamic registration could be done with an association `"architecture name" -> callback setting IR, jit, ... attributes`. We could then imagine having either a `Machine` class with a dictionary as class attribute representing this aforementioned association, or an initial empty object and further registration. Something like:
```Python

def set_x86_16(machine):
    from ... import ...
    machine._ir = ...

class Machine(object):
    architectures = {"x86_16": set_x86_16,
                              "x86_32": set_x86_32,
                              ...
    }

    def __init__(self, arch_name):
        cb = self.architectures.get(arch_name, None)
        if cb is None:
            raise ValueError("...")
        cb(self)

# Further add of a new arch
Machine.architectures["my_awesome_arch"] = set_my_awesome_arch
```
or
```
class Machine(object):
    
    @classmethod
    def register(arch, callback):
        ...

Machine.register("x86_16", ...)
Machine.register("x86_32", ...)

# Further add of a new arch
Machine.register("my_awesome_arch", set_my_awesome_arch)
```

Is that what you requested?

guedou:
This is exactly what I meant =)

[...]

PR #620, opened 2017-09-26
[proposal] Simplified API architecture

guedou:
This PR attempts to ease implementing a new miasm architecture. The goals are to:
- define "default" values directly into miasm2/core/cpu.py
- add a generic implementation of `addop()`

Implementing a new architecture, could be done as follow:
```python
class instruction_pips(instruction):
    pass

class mn_pips(cls_mn):
    name = "pips"
    instruction = instruction_pips
    regs = pips_regs_module

class pips_reg(reg_noarg, m_arg):
    reg_info = gpr_infos
    parser = reg_info.parser

reg02 = bs(l=2,  cls=(pips_reg, ))
imm02 = bs(l=2, cls=(imm_noarg, m_arg))

addop(mn_pips, "NOP", [bs("00000000")])
addop(mn_pips, "MOV", [bs("0001"), reg02, reg02])
```

Let me know if you think that it is a good idea, and if you want me to modify the code.

This is related to issue #619 =)


commial:
I think the global idea is good: adding an architecture in Miasm shouldn't be a series of copy/paste from others architectures.
Regarding the implementation, I have some remarks (not pinpointed in the diff because kind of global):
* you've got the Miasm flue: it's basic blocK
* `addop` could become a classmethod of the `mn` class, and then the code:
```Python
mn_pips.addop("NOP", ...)
mn_pips.addop("MOV", ...)
```
This change is also valid for already existing architectures (maybe with a `armtop = mn_armt.addop` to avoid to many changes in the diff)
* Some default values are great, others risk to end in weird and hard to debug behavior. I think it is preferable to pinpoint to the user where some implementation is still needed instead of providing a arbitrary default one. I think about `*flow`, `is_subcall`.
A `raise RuntimeError("Abstract method")` could help to have Python raising an error on the right function, in the right call context with still having your comments to help

guedou:
Thanks for your comments. Does it look better now ?

I update the example included in my first post to reflect the changes to the cls_mn API.

serpilliere:
It seems good to me!

guedou:
Rebased. Let me know if you want me to change things.

PR #628, opened 2017-10-18
DiGraph: fixed node order in natural loop backedges

mrphrazer:
just noticed that the returned back edge wasn't a back edge :D


commial:
Nice spot!

PR #634, opened 2017-11-05
Use ExprCond for {f,}cmov and do not generate a new block

aguinet:
This makes symbolic emulation really much faster on codes where a lot of
cmov happen (like some code from some grehack challenge ;))! I think the performance gain comes from the fact that we do not need to "clone" the symbolic state and execute two branches just for a cmov operation.

One other way to do this I think if we want to keep the branch associated with a ``cmov`` operation is to have some kind of pre-optimisation of the CFG where small if-then-else subgraphs could be "pre-computed" and replaced by ``ExprCond`` expressions! This would have the advantage to be generic and potentially optimize more code for full symbolic executions.

### DIFF
--- a/miasm2/arch/x86/sem.py
+++ b/miasm2/arch/x86/sem.py
@@ -261,17 +261,12 @@ def gen_fcmov(ir, instr, cond, arg1, arg2, mov_if):
     @cond: condition
     @mov_if: invert condition if False"""
 
-    lbl_do = m2_expr.ExprId(ir.gen_label(), ir.IRDst.size)
-    lbl_skip = m2_expr.ExprId(ir.get_next_label(instr), ir.IRDst.size)
     if mov_if:
-        dstA, dstB = lbl_do, lbl_skip
+        dstA, dstB = arg2, arg1
     else:
-        dstA, dstB = lbl_skip, lbl_do
-    e = []
-    e_do, extra_irs = [m2_expr.ExprAff(arg1, arg2)], []
-    e_do.append(m2_expr.ExprAff(ir.IRDst, lbl_skip))
-    e.append(m2_expr.ExprAff(ir.IRDst, m2_expr.ExprCond(cond, dstA, dstB)))
-    return e, [IRBlock(lbl_do.name, [AssignBlock(e_do, instr)])]
+        dstA, dstB = arg1, arg2
+    e = [m2_expr.ExprAff(arg1, m2_expr.ExprCond(cond, dstA, dstB))]
+    return e, []
 
 
 def gen_cmov(ir, instr, cond, dst, src, mov_if):
@@ -281,17 +276,16 @@ def gen_cmov(ir, instr, cond, dst, src, mov_if):
     @cond: condition
     @mov_if: invert condition if False"""
 
-    lbl_do = m2_expr.ExprId(ir.gen_label(), ir.IRDst.size)
-    lbl_skip = m2_expr.ExprId(ir.get_next_label(instr), ir.IRDst.size)
+    if dst in [ES, CS, SS, DS, FS, GS]:
+        src = src[:dst.size]
+    if src in [ES, CS, SS, DS, FS, GS]:
+        src = src.zeroExtend(dst.size)
     if mov_if:
-        dstA, dstB = lbl_do, lbl_skip
+        dstA, dstB = src,dst 
     else:
-        dstA, dstB = lbl_skip, lbl_do
-    e = []
-    e_do, extra_irs = mov(ir, instr, dst, src)
-    e_do.append(m2_expr.ExprAff(ir.IRDst, lbl_skip))
-    e.append(m2_expr.ExprAff(ir.IRDst, m2_expr.ExprCond(cond, dstA, dstB)))
-    return e, [IRBlock(lbl_do.name, [AssignBlock(e_do, instr)])]
+        dstA, dstB = dst,src 
+    e = [m2_expr.ExprAff(dst, m2_expr.ExprCond(cond, dstA, dstB))]
+    return e, []
 
 
 def mov(_, instr, dst, src):


serpilliere:
Hi

I don't agree with your current implementation: it changes the semantic of the cmov and adds an explicit READ of the source register. This is not the case if the instruction condition is not valid.

But your second proposition seems much more attractive, as it relies on the user hypothesis: if *the user* decides to do such an optimization, he can add such transformer!

In this case, the use of the `DiGraphSimplifier` or something will be helpfull!
 Do you agree with this?


By the way, you can argue that the `ExprCond` has the same "unwanted" parasitic READs. This is true. That's why we try to limit the sources of the `ExprCond` to be only integers or label (ie const)
This results in those *interesting* side effects on cmov, shl/shr/ror/rol...

commial:
Hi,
I agree with @serpilliere. In addition, if the compiler doesn't optimize, and outs a real "if-then" pattern, you're back with the same problem.
So it's probably better to manage it at the IR-level instead of at the ASM -> IR conversion.

Some of the probably useful ingredients are already in Miasm: `DiGraphSimplifier` for applying simplification at the graph level, post-dominators computing in the graph (to obtain the "influence" of a given conditional jump, which will probably ends as the condition in the resulting `ExprCond`) and `MatchGraph` if you prefer to try with pattern detection.

aguinet:
Thanks for your answers @commial  and @serpilliere !

If I understand it correctly what you say, the transformation from ``if (cond) v = A`` to ``v = ExprCond(cond, v, A)`` isn't valid *in the regard of* (let's call it) the "register read analysis".

The way I see it, there are two ways to fix this first problem, which is basically surapproximating the register read analysis:
1) (somehow what you propose) we don't care about it as long as the user can choose whether to apply this transformation  (regardless if this is done at ASM->IR time or into an IR transformation pass, which is IMHO another separate problem).
2) we modify a bit the Miasm IR to make this transformation valid for the register read analysis. One idea is to add attributes to the ExprCond object (maybe even the Expr object itself) which would say "these operands are never read/write", or, said differently, attributes that describe the read/write effects of the operands of an Expr(Cond) object. Then, the various Miasm analyses can rely on this to be more precise.
FTR, I discussed with @JonathanSalwan about this and he told me that Triton is doing this surapproximation (and using the Triton equivalent of ``ExprCond`` for ``cmov``), and is fine with it so far.
For information/further reference, in the LLVM IR, there is something similar with attributes that can be set on function arguments, that can describe (among other) memory aliases between pointers (see https://llvm.org/docs/LangRef.html#parameter-attributes).

The solution 2) seems "cleaner" to me, but I don't know how difficult that would be to do with the actual Miasm codebase.

Regarding the second problem (where this transformation should happen), I think it's indeed better/safer to do this on with an IR transformation pass, as we know that in some cases it makes symbolic execution faster, but it might not be true for other use cases!

What do you think :) ?

serpilliere:
Ok:
First of all, the example of the non-optimized code which Camille gave will break your second solutions.

Moreover, (tell me if I am wrong), if you use the `ExprCond` trick, you only "move" your problem: I am ok if the symbolic emulation is faster, but the complexity will be translated into the resulting expressions themselves. 8 consecutive cmovz will generate a really big mess.

Another point about the *context*: it may also break the jitter. Imagine the instruction (which doesn't exist):
```
cmovz EAX, DWORD PTR [0]
```

At execution time, Miasm preloads every read memory in order to deal with parallel instructions so it will try to load `[0]` even if the cmovz is not executed.

By the way, I look at nearly every `ExprCond` in x86 semantic: very few of them (about 5 on a total of 80 ExprCond) use something else than an integer or a label in their sources. (and you gave me a bit of motivation to fix those remaining :smiley: )

aguinet:
> First of all, the example of the non-optimized code which Camille gave will break your second solutions.

Hmm, what example are we talking about here :) ?

> Moreover, (tell me if I am wrong), if you use the ExprCond trick, you only "move" your problem: I am ok if  the symbolic emulation is faster, but the complexity will be translated into the resulting expressions themselves. 8 consecutive cmovz will generate a really big mess.

Yes, it clearly depends on what is your final goal. If for instance you want to generate a Z3 expression out of a loop-free and pure function, you want to end up with these bunch of ``ExprCond`` in the end, and the full symbolic execution gets faster if you remove "branches" as early as possible. If you are doing something else, it might be indeed a bad idea to have these ``ExprCond`` (don't have any example in mind but that should be easy to find for you guyz :)). In the end, I think we all agree that the user must have to choice!

My first point was to say that it seems ""broken by design"" to allow a transformation that breaks an analysis, so the idea of adding read/write effects attributes to ``Expr`` objects was to somehow fix this issue, but by reading your last example it might be not be such a great idea (see below).

> Another point about the context: it may also break the jitter. Imagine the instruction (which doesn't exist):
> 
> cmovz EAX, DWORD PTR [0]
> 
> At execution time, Miasm preloads every read memory in order to deal with parallel instructions so it will try to load [0] even if the cmovz is not executed.

That is indeed a problem, and it actually looks like a generalization of the first "dead register read" problem. It looks to me that the whole problem is how should ``ExprCond`` be interpreted by various analyses/evaluators (sym exec, jitter, etc...). The easy way seems to stay on what I called the first solution ("we don't care" and let the user do this transformation if he knows what he is doing), but I still can be interesting to see if there is a way to somehow "extend" ``ExprCond`` to make its behavior equivalent to the one of an if-then-else sequence w.r.t every analyses/evaluators of Miasm! (that can be a discussion to be continued at grehack next week :)).

serpilliere:
Here is @commial  example:
You *currently* have the following code:
```
cmp ecx, 3
cmovz eax, ebx
```

But if the compilator didn't optimized your program, it would not have generated a cmovz instruction. It would have generated something like:
```
cmp ecx, 3
jnz skip
mov eax, ebx
skip:
```

And in this case, you are back to the same IR miasm generates currently with a cmovz: a condition and a split in the execution control flow graph.

That's why @commial said that generating an ExprCond for a cmovz doesn't resolve fully the problem.

commial:
> Yes, it clearly depends on what your final goal is. If for instance you want to generate a Z3 expression out of a loop-free and pure function, you want to end up with these bunch of ExprCond in the end anyway, and the full symbolic execution tends to get faster if you remove as much "branches" as early as possible. If you are doing something else, it might be indeed a bad idea to have these ExprCond (don't have any example in mind but that should be easy to find for you guyz :)). In the end, I think we all agree that the user must have to choice!

For me, the main advantage of merging if-then-else blocks in an ExprCond while doing symbolic + z3 analysis is, beside the potential speed-up, that you're giving to z3 the full information in one shot.
With several blocks, you'll have to manage a constraint on the path, basically saying to z3 "with the constraint 'x != 0', eax = 4". But z3 will not now what happen when "x != 0" doesn't stand.
After `ExprCond` transforms, you'll have the info on both possibilities. When asking for satisfiability / counter-example generation, it will be able to give a more educated answer.

> That is indeed a problem, and it actually looks like a generalization of the first "dead register read" problem. It looks to me that the whole problem is how should ExprCond be interpreted by various analyses/evaluators (sym exec, jitter, etc...). The easy way seems to stay on what I called the first solution (the current definition is kept and we let the user do this transformation if he knows what he is doing), but I think it can still be interesting to see if there is a way to somehow "extend" ExprCond to make its behavior equivalent to the one of an if-then-else sequence w.r.t every analyses/evaluators of Miasm! (that can be a discussion to be continued at grehack next week :)).

From what I understand, you're saying the `ExprCond` should be considered as a if-then-else pattern in the CFG.
With this in mind, it should means that a `get_r()` on `ExprCond` should not return a set of `Expr`, but includes the condition, ie. `src1` is read if `cond`, `src2` is read if `!cond`.
There is an additional question: how the if-then pattern should be represented? A new word?

So, even if I totally agree with this view of `ExprCond`, it is probably not a short-term change.
As evoked by @a-vincent, another possibility is to change the representation of condition, to move them to edge annotation. This is the model BinCOA project is using, and you can have additional information [here](https://bincoa.labri.fr/trac/raw-attachment/wiki/public/publications/BINCOA-formal-model.pdf). But it likely involves huge changes in the whole framework.

However, a transformation pass on IR enabled by the user is probably more accessible and seems to be, from the discussions above, a shared solution.
So I propose to close this PR (the solution implemented being dropped), and have a more in-depth conversation, maybe at Grehack :smiley:. Aside, it would be great to start working on the IR transformation pass to answer your needs. I think [the veritesting paper](https://users.ece.cmu.edu/~aavgerin/papers/veritesting-icse-2014.pdf) is using a similar method (kind of "multi-path symbolic execution").

aguinet:
> That's why @commial said that generating an ExprCond for a cmovz doesn't resolve fully the problem.

Okay agreed! That's why I also proposed as an alternate solution the transformation at the IR level :)

> For me, the main advantage of merging if-then-else blocks in an ExprCond while doing symbolic + z3 analysis is, beside the potential speed-up, that you're giving to z3 the full information in one shot.
With several blocks, you'll have to manage a constraint on the path, basically saying to z3 "with the constraint 'x != 0', eax = 4". But z3 will not now what happen when "x != 0" doesn't stand.
After ExprCond transforms, you'll have the info on both possibilities. When asking for satisfiability / counter-example generation, it will be able to give a more educated answer.

I completely agree with this!

> As evoked by @a-vincent, another possibility is to change the representation of condition, to move them to edge annotation. This is the model BinCOA project is using, and you can have additional information here. But it likely involves huge changes in the whole framework.

I cross read the definition and it indeed seems quite interesting. I trust you on the fact that it might be hard to put into Miasm :)

> However, a transformation pass on IR enabled by the user is probably more accessible and seems to be, from the discussions above, a shared solution.

Yep!

> So I propose to close this PR (the solution implemented being dropped), and have a more in-depth conversation, maybe at Grehack 😃. Aside, it would be great to start working on the IR transformation pass to answer your needs. I think the veritesting paper is using a similar method (kind of "multi-path symbolic execution").

Fine by me, especially the part involving Grehack :) I'll take a look at the various pointers you gave me for in order to do this IR transformation!

Thanks for the interesting discussion!

PR #642, opened 2017-12-11
Remove some unnecessary parentheses when str()'ing expressions

a-vincent:
This patch removes quite a few useless parentheses from expressions
thanks to the usual priorities between operators.

It is conservative in the sense that not all such parentheses are
removed when we felt the priorities won't be known by most users.


serpilliere:
Thanks!

PR #652, opened 2018-01-05
Ida/Example: fix symbexec plugin

serpilliere:


commial:
FYI, tested & working on IDA:
* 7.0
* 6.95
* 6.9
* 6.8
* 6.7

For the 6.9 and below, the `menu.py` script is not working (error on depgraph import), but the `symbol_exec.py` script alone is.

PR #659, opened 2018-01-16
Jitter: remove french debug string

serpilliere:
@mkrzywix if you are ok with this, I merge this commit part


mkrzywix:
Sure, please do.

PR #660, opened 2018-01-16
Append b to the opens mode

serpilliere:
Python on Windows makes a distinction between text and binary files

@w4kfu: if you are ok with this, I merge this commit part!


w4kfu:
Sure!

PR #661, opened 2018-01-18
changing dot generation

GAJaloyan:
Changing the dot source code generated by miasm.

- splines=ortho
- adding URL to nodes and to edges (edges point to their tail node)
- adding constraint = false to backjumps (so that the nodes follow the address order).

### DIFF
--- a/example/ida/ctype_propagation.py
+++ b/example/ida/ctype_propagation.py
@@ -10,7 +10,7 @@ from miasm2.arch.x86.ctype import CTypeAMD64_unk, CTypeX86_unk
 from miasm2.arch.msp430.ctype import CTypeMSP430_unk
 from miasm2.core.objc import CTypesManagerNotPacked, ExprToAccessC, CHandler
 from miasm2.core.ctypesmngr import CAstTypes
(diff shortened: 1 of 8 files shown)


serpilliere:
We are currently testing this PR to see if the new graph style is ok.
As this PR is about "graph style", it takes some time 😄

GAJaloyan:
I am working on a version that uses dominators instead of looking for addresses at backjump. Not sure of the effect on execution speed though. I'll commit it on a separate PR.

serpilliere:
This is interesting: we were wondering with @a-vincent if it would be interesting to extend the idea and include our own graph layout.
There is a layout tool called https://github.com/bdcht/grandalf which implements the Sugiyama algorithm. It's roughly the dot layout, without the heuristic parts if I remember correctly.

But we were wondering if a simplier algorithm cannot be applied here, taking the fact that we have "special" graphs: program functions (I mean in most cases one head, directed graph, and classic patterns switch/if/jump)

For example, the https://github.com/jjyg/metasm project uses it's own layout based on pattern recognition (switch case, if/then/else,...) which seems to be simple, efficient and pretty close to what a reverser wants in term of "graph" pattern.

I didn't look at radare yet, maybe they have their own layout as well.

PR #677, opened 2018-02-12
Changing index from 1 to -1.

GAJaloyan:
It seems that the last argument is the one that is set to 0 (as in A op 0 =>A).


serpilliere:
Thank you!

Issue #695, opened 2018-03-09
Compiled with new XMM registers, but program stops abnormally on XMM instruction with no errors.


SneakyNachos:
I ran into a windows binary that used xmm0 through xmm2, so I decided to modify the C files of miasm in attempt to add support for the xmm registers. Currently the additions compile fine after 

`python setup.py install`

The issue that remains is that the Sandbox_Win_x86_32 seems to stop continuing to execute instructions once it hits an XMM instruction, almost as if told to stop on XMM instructions.

So my question is, what is required to force the Sandbox to continue walking the instructions after the XMM instruction is seen?

I'm just not sure if I made the right addition to the Sandbox to continue execution after a 128-bit transaction occurs.

The binary used can be acquired at the ctf github at
[https://github.com/ctfs/write-ups-2016/blob/master/csaw-ctf-2016-quals/reverse/key-125/key.exe](url)

My script used is:
```
from argparse import ArgumentParser
from miasm2.analysis.machine import Machine
from miasm2.expression.expression import *
from miasm2.core import asmblock
from miasm2.analysis.sandbox import Sandbox_Win_x86_32
from miasm2.jitter.csts import PAGE_READ,PAGE_WRITE
def kernel32_QueryPerformanceCounter(jitter):
	ret_addr,args = jitter.func_args_systemv(["lpPerformanceCount"])
	ret_value = 0
	jitter.func_ret_systemv(ret_addr,ret_value)
	pass
def main():	
	parser = Sandbox_Win_x86_32.parser(description="Sandbox")
	parser.add_argument("filename",help="filename")
	options = parser.parse_args()
	options.mimic_env = True
	sb = Sandbox_Win_x86_32(options.filename,options,globals())
	sb.jitter.init_run(sb.entry_point)
	sb.jitter.vm.add_memory_page(0x0,PAGE_READ|PAGE_WRITE,"\x00"*4096)
	sb.jitter.jit.log_mn=True
	sb.run()
	pass
main()
```

The command used to run the sandbox is 

```python s.py key.exe -a 0x401100```

Below are the additions made within the C files codegen.py, JitCore.c, JitCore.h, JitCore_x86.h, vm_mngr.c, and vm_mngr.h

First I setup the type definitions for 128-bit integers by adding in the headers

```
typedef __int128 int128_t;
typedef unsigned __int128 uint128_t;
```

I then added functions for MEM_WRITE_128, MEM_LOOKUP_128 in JitCore.c

```
[...]

PR #696, opened 2018-03-09
Allow python simulation of function with '-' in their name

commial:
See #695 .

In windows, DLL names could contains `-`, which is not a legit Python function name (for stubs definition).


serpilliere:
Hum.

Is it a real good idea to replace the - with a _?
I mean the _ is a character which is also authorized in input names.

commial:
I agree. I've use it because this is the "usual" way (such as `ArgumentParser`), but collision are still possible.
Do you have a better idea? (like mangling or something?) Or user have to use a `globals()[name] = func` trick?

PR #700, opened 2018-03-13
Fix/ppc cntlzw

commial:


commial:
Updated after discussion with @a-vincent and @serpilliere (ie. BSF and BSR now have default value, and then are defined on 0, instead of stopping execution)

commial:
Replaced by #740

PR #710, opened 2018-03-30
X86: fix segm's size generation

serpilliere:
The code fixed instruction's argument with segment, but not segment generated *in* mnemonic body


miasm2/arch/x86/sem.py
```diff
@@ -982,8 +983,8 @@ def cmps(ir, instr, size):
     if ir.do_str_segm:
         if instr.additional_info.g2.value:
             raise NotImplementedError("add segm support")
-        src1_sgm = m2_expr.ExprOp('segm', DS, src1)
-        src2_sgm = m2_expr.ExprOp('segm', ES, src2)
+        src1_sgm = ir.gen_segm_expr(DS, src1)
+        src2_sgm = ir.gen_segm_expr(DS, src2)
```
commial: Is the `ES` -> `DS` replacement wanted?

PR #713, opened 2018-04-11
Ssa transform

serpilliere:


serpilliere:
Hi!
@mrphrazer, news for the SSA: It's *really* an key feature and we *really* want to include it in Miasm. As your PR is old (2 year ago!) I reworked it and rebased it on the current master. Old comments from the original PR have been taken in account as well.


We have used it (for example in constant propagation), and it really gives instesting results :smile: 
Tell me if you are ok with this, because we are burning to release the next parts :smile:

mrphrazer:
Hi!

We are happy to hear this :) Go for it!

serpilliere:
PR seems ready to me

serpilliere:
Rebase done

miasm2/analysis/ssa.py
```diff
@@ -0,0 +1,587 @@
...
+
+
+class SSA(object):
+    """
+    Generic class for static single assignment (SSA) transformation
+
+    Handling of
+    - variable generation
+    - variable renaming
+    - conversion of an DiGraphIR block into SSA
```
commial: It's now an "IRCFG" ;)

miasm2/analysis/ssa.py
```diff
@@ -0,0 +1,587 @@
...
+            ids_rhs = self.get_regs(cur_rhs)
+
+            # add RHS ids to worklist
+            for id_rhs in ids_rhs:
+                if id_rhs in self.expressions:
+                    todo.add(id_rhs)
+        return expr
+
+
+class SSAPath(SSABlock):
```
commial: Why is `SSAPath` inheriting from `SSABlock` and not from `SSA`?

commial:
Thanks for the PR and for the work on re-integration to the current codebase!
Apart the few tiny comments, the patch is OK for me

miasm2/analysis/ssa.py
mrphrazer: Because you inherit `reassemble_expr`. Understanding a path as a sequence of block, this makes sense. A similar function on CFG-level would work differently since it has to consider all the different branches/phi nodes.

miasm2/analysis/ssa.py
```diff
@@ -0,0 +1,574 @@
...
+
+    The implemented SSA form is known as minimal SSA.
+    """
+
+    PHI_STR = 'Phi'
+
+
+    def __init__(self, ircfg):
+        """
+        Initialises SSA class for directed acyclic graphs
```
serpilliere: A last question for @mrphrazer 
The *acyclic* here seems strange to me here.
I think it's possible to have cylces in even in ssa form, passing through a phi operator.
Did I miss something here?

mrphrazer: You are right, the implemention works as it is also for cycles. The comment is a relic from a project where I had to enforce acylcic graphs via loop unrolling.

commial:
Thanks again @mrphrazer for this contribution, and @serpilliere for refactoring it. We were using it internally for some time now, and it's quite a nice tool!

PR #714, opened 2018-04-11
Asmblock: fix AsmSymbolPool API

serpilliere:


miasm2/core/asmblock.py
```diff
@@ -426,11 +437,18 @@ def set_offset(self, label, offset):
...
             self._offset2label[label.offset] = label
 
+    @property
+    def labels(self):
+        """Return all labels"""
+        return self._labels
+
     @property
     def items(self):
+        warnings.warn('DEPRECATION WARNING: use "labels" instead of "items"')
```
commial: You probably want to swap this two lines, to have the docstring right after the function definition

miasm2/core/asmblock.py
```diff
@@ -426,11 +437,18 @@ def set_offset(self, label, offset):
...
+    @property
+    def labels(self):
+        """Return all labels"""
+        return self._labels
+
     @property
     def items(self):
         """Return all labels"""
+        warnings.warn('DEPRECATION WARNING: use "labels" instead of "items"')
         return self._labels
```
commial: To follow the original behavior, maybe we can returns here a `list(self._labels)`?

serpilliere: Fixed

PR #720, opened 2018-04-15
Added support to parse  sxtb, sxth, sxtx, uxtw, uxth, uxtx, uxtb

su-vikas:
Hi, 

I have added the  code for parsing ARM register extender instructions, viz. sxtb, sxth, sxtx, uxtw, uxth, uxtx, uxtb. I have tested it for UXTW, and assumed the things will be same for other instructions as well. 

For reference, apart from ARM ASM documentation, I have used this blog https://community.arm.com/processors/b/blog/posts/a64-shift-and-extend-operations-operand-modifiers. 

Please let me know if anything is missing.

### DIFF
--- a/miasm2/arch/aarch64/sem.py
+++ b/miasm2/arch/aarch64/sem.py
@@ -124,9 +124,32 @@ def extend_arg(dst, arg):
         return arg
 
     op, (reg, shift) = arg.op, arg.args
-    if op == 'SXTW':
+    if op == "SXTB":
+        base = reg[:8].signExtend(dst.size)
+        op = "<<"
+    elif op == "SXTH":
+        base = reg[:16].signExtend(dst.size)
+        op = "<<"
+    elif op == 'SXTW':
+        base = reg[:32].signExtend(dst.size)
+        op = "<<"
+    elif op == "SXTX":
         base = reg.signExtend(dst.size)
         op = "<<"
+
+    elif op == "UXTB":
+        base = reg[:8].zeroExtend(dst.size)
+        op = "<<"
+    elif op == "UXTH":
+        base = reg[:16].zeroExtend(dst.size)
+        op = "<<"
+    elif op == 'UXTW':
+        base = reg[:32].zeroExtend(dst.size)
+        op = "<<"
+    elif op == "UXTX":
+        base = reg.zeroExtend(dst.size)
+        op = "<<"
+
     elif op in ['<<', '>>', '<<a', 'a>>', '<<<', '>>>']:
         base = reg.zeroExtend(dst.size)
     else:


miasm2/arch/aarch64/sem.py
```diff
@@ -813,7 +835,6 @@ def __init__(self, symbol_pool=None):
         self.pc = PC
         self.sp = SP
         self.IRDst = m2_expr.ExprId('IRDst', 64)
-        self.addrsize = 64
```
commial: It seems you have removed a line you don't want to

miasm2/arch/aarch64/sem.py
```diff
@@ -127,6 +127,28 @@ def extend_arg(dst, arg):
     if op == 'SXTW':
         base = reg.signExtend(dst.size)
         op = "<<"
+    elif op == "SXTB":
+        base = reg.signExtend(dst.size)
+        op = "<<"
```
serpilliere: Hi!

Ok there is an error in the semantic description here, but I can't blame you: the original SXTW was wrong :smile: 
As the documentation says the steps are:
- first truncate (Byte, Word)
- extend (signed /unsigned)
- shift (optionnal)

So the SXTW (and others !) should be something like:
```python
     if op == 'SXTW':
         base = reg[:16].signExtend(dst.size)
         op = "<<"
```

Do you agree with this?

su-vikas: I totally agree with it. I am guilty of skipping this info too. I just assumed SXTW was correctly written. I will update the patch and submit again.

miasm2/arch/aarch64/sem.py
su-vikas: You are correct. I am sorry for my oversight.

miasm2/arch/aarch64/sem.py
```diff
@@ -124,31 +124,32 @@ def extend_arg(dst, arg):
...
         op = "<<"
-    elif op == "SXTB":
-        base = reg.signExtend(dst.size)
+    elif op == 'SXTW':
+        base = reg[:16].signExtend(dst.size)
         op = "<<"
     elif op == "SXTH":
-        base = reg.signExtend(dst.size)
+        base = reg[:32].signExtend(dst.size)
         op = "<<"
```
serpilliere: Ok I did it again: My comment was wrong again: the `SXTW` is *word* sign extend so:
```python
     if op == 'SXTW':
         base = reg[:32].signExtend(dst.size)
         op = "<<"
```

The 16 bit is for `SXTH` !

So all the *W/*H are inverted. Sorry for that :confused:
By the way, can you squash your commits into one?

su-vikas: Gotcha. Fixed it, hope got it right the 3rd time :).

serpilliere:
Thanks a lot!

PR #733, opened 2018-05-04
Jitter: fix FPU C functions names

serpilliere:


miasm2/ir/translators/C.py
```diff
@@ -94,12 +94,17 @@ def from_ExprOp(self, expr):
                                                    self.from_expr(expr.args[0]),
                                                    self.from_expr(expr.args[1]),
                                                    size2mask(expr.args[0].size))
+            elif expr.op == 'cpuid':
+                return "%s(%s, %s)" % (expr.op,
+                                       self.from_expr(expr.args[0]),
+                                       self.from_expr(expr.args[1]))
             elif (expr.op.startswith('cpuid') or
```
commial: if `cpuid` is handled above, why continuing to check it here?

serpilliere:
Correct. Fixed

PR #734, opened 2018-05-05
Enhance parser

serpilliere:
This PR modify the parser structure:
- ParserAst is removed
- base_expr is common to all archs
- it is independant from archs
- the dependant stuff is done in `asm_ast_to_expr`
- the parser and the transformation to Expr is not in two different objects
- the parser should be easily replaced/rewritted


miasm2/arch/ppc/arch.py
```diff
@@ -1,5 +1,6 @@
 
 import logging
+from pdb import pm
```
commial: Probably a debug residue

miasm2/core/cpu.py
```diff
@@ -361,29 +254,187 @@ def neg_int(t):
...
+    assert len(t) == 1
+    reg = t[0]
+    return AstId(reg)
+
+
+def cb_deref_reg(t):
+    assert len(t) == 1
+    reg = t[0]
+    ptr = reg
+    mem = AstMem(ptr, 32)
```
commial: Why all `AstMem` in these functions have a default size of 32?

miasm2/core/cpu.py
```diff
@@ -361,29 +254,187 @@ def neg_int(t):
...
+        op, args = tokens[1], [tokens[0], tokens[2]]
+    elif len(tokens) > 3:
+        args = [tokens.pop(0)]
+        i = 0
+        while i < len(tokens):
+            op = tokens[i]
+            arg = tokens[i+1]
+            i += 2
+            if op == '-':
+                if isinstance(arg, AstInt):
```
commial: This is already done in `Ast` class code. `- arg` is then enough

miasm2/core/cpu.py
```diff
@@ -361,29 +254,187 @@ def neg_int(t):
...
+    return result
+
+
+def cb_op_sign(t):
+    assert len(t) == 1
+    op, value = t[0]
+    return -value
+
+
+def cb_op_div(t):
```
commial: In these function, it is not very clear what is the shape of `t` and what it contains. So, reading the function, it is hard to understand why `tokens[1]` is ignored here

commial:
Okay, so there is a lot of change in this PR and API cleaning, nice work!

I also see the `fromstring` and  ̀parse_txt` API have changed and are no longer compatible with previous ones, as `symbol_pool` is now a mandatory argument.

I understand there is a need to fill `symbol_pool` while parsing. But as the API is more and more complex in order to expose needed elements, it would be great to have in a near future the equivalent of `disasmEngine` for assembly, that is to say an object which store redundant arguments (every ones except the source).
What do you think?
FYI, Keystone API looks like:
```
ks = Ks(KS_ARCH_X86, KS_MODE_32)
# Optionally
ks.sym_resolver = sym_resolver

encoding, count = ks.asm(CODE)
```

miasm2/arch/x86/arch.py
```diff
@@ -121,148 +123,92 @@
...
-    r1 = reg2exprid(t[0][0])
-    i1 = ExprInt(t[1].arg, r1.size)
-    return r1 + i1
-
-
-def parse_deref_regreg(s, l, t):
-    t = t[0]
-    return t[0][0] + t[1][0]
+segm2enc = {CS: 1, SS: 2, DS: 3, ES: 4, FS: 5, GS: 6}
+enc2segm = dict([(x[1], x[0]) for x in segm2enc.items()])
```
commial: Defined twice

miasm2/arch/x86/arch.py
```diff
@@ -282,68 +228,121 @@ def ast_int2expr(a):
...
+                size = list(fixed_size)[0]
+                args = [self.asm_ast_to_expr(arg, symbol_pool, size, fixed_size) for arg in value.args]
+            else:
+                raise ValueError("Size conflict")
+            if None in args:
+                return None
+            return ExprOp(value.op, *args)
+        if isinstance(value, AstInt):
+            assert size_hint is not None
+            if 1 << size_hint < value.value:
```
commial: In which occasion this code is reached?

commial:
Apart these comments, LGTM. I don't know enough this code part to be able to go more in details.

miasm2/arch/ppc/arch.py
serpilliere: ack

miasm2/core/cpu.py
serpilliere: Correct: unused function: removed

miasm2/core/cpu.py
serpilliere: Ack

miasm2/core/cpu.py
serpilliere: Add assert tokens[1] == '/'

miasm2/arch/x86/arch.py
serpilliere: Ack

miasm2/arch/x86/arch.py
serpilliere: Ack. assert removed

miasm2/arch/x86/arch.py
serpilliere: The `fromstring` is added  here for the parsing of *one* instruction only. Adding the `symbol_pool` is necessary somewhere as you saw. As the `parse_asm` (which parses a complete code) takes the `symbol_pool`, I made the API similar. So function tend to have the same API :smile: 

For the form you proposed, I don't down which is the best honestly. Maybe one day we will switch to it!

miasm2/core/cpu.py
commial: This is also the case for others functions, such as `cb_op_mul`.
`t` is maybe not the best name also, maybe `parsed_tokens` if I correctly understand what it is?

miasm2/arch/x86/arch.py
commial: I was talking about the size change

miasm2/core/cpu.py
commial: Well, let me be more precise. From `AstNode` code:
```Python
    def __neg__(self):
        if isinstance(self, AstInt):
            value = AstInt(-self.value)
        else:
            value = AstOp('-', self)
        return value
```
So the whole `if isinstance(..., AstInt): ...; else: arg = -arg` is actually `arg = - arg`

commial:
> 
> The fromstring is added here for the parsing of one instruction only. Adding the symbol_pool is necessary somewhere as you saw. As the parse_asm (which parses a complete code) takes the symbol_pool, I made the API similar. So function tend to have the same API

I'm okay with that, and that similarity is a good thing. This is what I was talking about with `parse_txt` and `formstring`.
My point is that, in their current form, 2 of 3 arguments in `formstring` and 3 of 4 arguments in `parse_txt` are likely to always be the same in a given session. This does sounds like what `disasmEngine` is designed to do, that is, pre-fill repetitive arguments for the complex disasm APIs.

> For the form you proposed, I don't down which is the best honestly. Maybe one day we will switch to it!
Ack

serpilliere:
Yes, I have your point @commial. But the number of arguments is not as huge as in the disasm engine (about 13 arguments). So I think I will let this like that for the moment.

PR #736, opened 2018-05-09
add getby_offset in AsmCFG

losynix:


miasm2/core/asmblock.py
```diff
@@ -736,6 +736,14 @@ def get_bad_blocks_predecessors(self, strict=False):
                     yield predecessor
                     done.add(predecessor)
 
+    def getby_offset(self, offset):
+        """Return block containing @offset"""
+        for block in self:
+            if block.lines[0].offset <= offset and \
```
commial: As a tiny remarks, you have in Python the possibility to write:
```
if block.lines[0].offset <= offset <= block.lines[-1].offset + block.lines[-1].l - 1
```
as a common way to ensure x in [a, b]

losynix: fixed, thanks :)

PR #738, opened 2018-05-14
armtl change branch pc offset

marilafo:
Change armtl code use to branch when offset is larger than expected
Change armtl code use to branch when offset is negative

### DIFF
--- a/miasm2/arch/arm/arch.py
+++ b/miasm2/arch/arm/arch.py
@@ -907,6 +907,8 @@ class arm_offs(arm_imm):
         if (1 << (self.l - 1)) & v:
             v = -((0xffffffff ^ v) + 1)
         v = self.encodeval(v)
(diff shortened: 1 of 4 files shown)


miasm2/arch/arm/arch.py
```diff
@@ -1899,6 +1901,8 @@ def encodeval(self, v):
         # Remove pipeline offset
         v -= 2 + 2
         if v % 2 == 0:
+            if v > (1 << (self.l - 1)) - 1:
+                return False
             return v >> 1
         return False
```
serpilliere:
```python
        if v % 2 != 0:
                return False
        if v > (1 << (self.l - 1)) - 1:
                return False
        return v >> 1
```

serpilliere:
Thank you!

PR #753, opened 2018-06-09
Add XMM0-15 to gpreg_dict[]

acru3l:
Without XMM0-15 in gpreg_dict[], a ValueError occurs when taking/restoring a cpu snapshot with get_gpreg()/set_gpreg() like in Sibyl.


commial:
Nice catch, thanks!
To reproduce:
```Python
>>> from miasm2.analysis.machine import Machine
>>> jitter = Machine("x86_32").jitter()
>>> jitter.cpu.set_gpreg(jitter.cpu.get_gpreg())
unkown key: XMM10
  File "<stdin>", line 1, in <module>
ValueError: unkown reg
```

PR #754, opened 2018-06-09
Jitter: label to_string

serpilliere:


miasm2/core/asmblock.py
```diff
@@ -67,8 +67,17 @@ def set_label(self, loc_key):
 
     label = property(get_label, set_label)
 
+    def to_string(self, symbol_pool=None):
+        if symbol_pool is None:
+            return "%s:%s" % (str(self.c_t), str(self.loc_key))
```
commial: `str`s are useless, due to string format

miasm2/core/asmblock.py
```diff
@@ -67,8 +67,17 @@ def set_label(self, loc_key):
...
+        else:
+            return "%s:%s" % (
+                str(self.c_t),
```
commial: Idem (and `c_t` is already a string)

miasm2/core/asmblock.py
serpilliere: done

miasm2/core/asmblock.py
serpilliere: done

PR #759, opened 2018-06-11
Test for PPC sandbox (travis crash, sometimes)

commial:
Please ignore, this is only for bug reproduction on travis.


commial:
Likely fixed by #794 bug fixes

PR #763, opened 2018-06-15
Toshiba MeP support

guedou:
This PR adds a new architecture to miasm. Toshiba MeP-c4 can be assembled, disassembled and jited.

It might require some modifications:
- I tried to document it as much as possible in order to help future architectures developers. Some `# GV:` comments might be removed or replaced
- test are currently performed using `pytest`. I did not investigate how to integrate them to `test_all.py`

Comments are welcome =)


miasm2/arch/mep/sem.py
```diff
@@ -0,0 +1,1147 @@
...
+    PC = dst
+    ir.IRDst = dst
+
+
+@sbuild.parse
+def blti(reg_test, imm4, disp16):
+    """BLTI - Branch if the register is lower than imm4."""
+
+    # if(Rn< ZeroExt(imm4)) PC <- PC +SignExt((disp17)16..1||0) - (Signed comparison)
+    dst = disp16 if "<"(reg_test, imm4) else ExprLoc(ir.get_next_break_loc_key(instr), 32)
```
guedou: The failing tests are likely caused by this line. Locally, I am using PR #556.

What is the correct way to modify this line?

serpilliere: Hi
First, Thanks for the PR !
For the sign problem: We plan to do a PR which introduce comparison operators, but it's not finish at that time. So if you can modify your PR to use current signed/unsigned equations, this is great! Don't worry, we will patch each architecture (including MeP once merged) to use the future API :smiley:

Is this ok for you?

Another point:
I started a review of the code: It seems you are introducing the use of `pytest`, which is currently not a dependance of Miasm. As we use regressions tests for commit checks, each machine will need `pytest`. Is it possible to modify the `test_all.py` in order to include MeP regressions tests to `test_all.py` to avoid the pytest introduction?

The point is that we made `test_all.py` quickly in order to solve the test regressions problem. If one day we plan to change it to use a real test module (it may be pytest!), we want a bit more time to test different "tests managers" before modifying the whole framework. Here, the PR 'forces' a bit the use of `pytest`. 
You may add a main calling your tests to fix this.

guedou: The current version implements an abstraction layer in the `launch.py` that is compatible with `test_all.py` and `pytest`. The current drawback is that It duplicates the `launch_tests()` in 3 files =/
Does it sound ok ?

I did not manage to make the sign comparison work correctly.  With my local fixes (not pushed), the jit tests fail with `NotImplementedError: Unknown op: '<s'` in miasm2/ir/translators/C.py. I don't know how to fix it.

test/arch/mep/jit/launch.py
```diff
@@ -0,0 +1,7 @@
+# Toshiba MeP-c4 - pytest unit tests wrapper
+# Guillaume Valadon <guillaume@valadon.net>
+
+from ut_helpers_jit import launch_tests
+
+from test_jit_branchjump import TestBranchJump; launch_tests(TestBranchJump())
+from test_jit_repeat.py import TestRepeat; launch_tests(TestRepeat())
```
serpilliere: Tipo here. Remove the `.py` to be able to run the test

```python
from test_jit_repeat import TestRepeat; launch_tests(TestRepeat())
```

miasm2/arch/mep/sem.py
serpilliere: Ok here is the trick:
First, you have to rip the flags computation from `expression/simplifications_cond.py`, for example in `expr_simp_inf_signed` for the signed inf.
Then, recode a function in your `sem.py` for example:
```python
def compute_s_inf(arg1, arg2):
    return ((arg1 - arg2) ^ ((arg1 ^ arg2) & ((arg1 - arg2) ^ arg1))).msb()
```

Then, add it to your sembuilder:
```python

def compute_s_inf(arg1, arg2):
    return ((arg1 - arg2) ^ ((arg1 ^ arg2) & ((arg1 - arg2) ^ arg1))).msb()

# SemBuilder context
ctx = {"PC": PC, "SP": SP, "LP": LP, "SAR": SAR, "TP": TP,
       "RPB": RPB, "RPE": RPE, "RPC": RPC, "EPC": EPC, "NPC": NPC,
       "EXC": EXC, "HI": HI, "LO": LO, "PSW": PSW, "DEPC": DEPC, "DBG": DBG,
       "exception_flags": exception_flags,
       "compute_s_inf":compute_s_inf}
sbuild = SemBuilder(ctx)
```

Once this, you can call the function in the sembuilder, and fix uage of 'high" comparison operators. For example:
```python

@sbuild.parse
def blti(reg_test, imm4, disp16):
    """BLTI - Branch if the register is lower than imm4."""

    # if(Rn< ZeroExt(imm4)) PC <- PC +SignExt((disp17)16..1||0) - (Signed comparison)
    dst = disp16 if compute_s_inf(reg_test, imm4) else ExprLoc(ir.get_next_break_loc_key(instr), 32)
    PC = dst
    ir.IRDst = dst

```


I know the equation is a bit large, but as I said you we will use the comparison operators in a next PR, but it's not ready yet :confused:

Is it ok for you?

test/arch/mep/jit/launch.py
guedou: Done!

miasm2/arch/mep/sem.py
guedou: I think that I understood. What is the correct way to implement `compute_u_inf()` the unsigned counterpart?

serpilliere: As I said, you can extract the equation from `expression/simplifications_cond.py`:
```python
def expr_simp_inf_unsigned_inversed(expr_simp, e):
    "((x - y) ^ ((x ^ y) & ((x - y) ^ x))) ^ x ^ y [31:32] == x <u y"
   ...
```

So you can write:
```python
def is_unsigned_inf(x, y):
    #  x <u y"
    result = (((x - y) ^ ((x ^ y) & ((x - y) ^ x))) ^ x ^ y).msb()
    return result
```

guedou: Thanks! I did not understand that I needed to convert the equations myself. I pushed a second commit that adds the fixes. I could squash it later.

Two REPEAT jit examples fail and I do not understand why =/

miasm2/arch/mep/sem.py
```diff
@@ -236,23 +247,18 @@ def add3(ir, instr, reg_dst, reg_src, reg_or_imm):
...
+    """ADVCK3 - Check addition overflow."""
 
     # if(Overflow(Rn+Rm)) R0<-1 else R0<-0 (Signed)
-
-    # Get the register storing the highest value
-    max_rn_rm = rn.zeroExtend(64) if ">"(rn, rm) else rm.zeroExtend(64)
-
-    # Check for an overflow
-    r0 = i32(1) if "<"(rn.zeroExtend(64) + rm.zeroExtend(64), max_rn_rm) else i32(0)
+    r0 = i32(1) if compute_u_inf(i64(0xFFFFFFFF), rn.zeroExtend(64) + rm.zeroExtend(64)) else i32(0)
```
commial: It seems that you're testing `0xFFFFFFFF < rn.zeroExtend + rm.zeroExtend`

You may want to do the opposite?

commial: My bad, please ignore this remark

miasm2/arch/mep/sem.py
```diff
@@ -0,0 +1,1172 @@
...
+from miasm2.arch.mep.arch import mn_mep
+from miasm2.arch.mep.regs import PC, SP, LP, SAR, TP, RPB, RPE, RPC, EPC, NPC, \
+    take_jmp, in_erepeat
+from miasm2.arch.mep.regs import EXC, HI, LO, PSW, DEPC, DBG
+from miasm2.expression.expression import ExprId, ExprInt, ExprOp
+from miasm2.expression.expression import ExprAff, ExprCond, ExprMem
+from miasm2.core.cpu import sign_ext
+
+from miasm2.arch.mep.regs import exception_flags
+MEP_EXCEPTION_ZERO_DIVIDE = 1 << 0
```
serpilliere: Can you use:
```python
from miasm2.jitter.csts import EXCEPT_DIV_BY_ZERO
```

miasm2/arch/mep/ira.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class ir_a_mepl(ir_mepl, ir_a_mepb):
+    """MeP high level IR manipulations - Little Endian"""
+
+    def __init__(self, symbol_pool=None):
+        ir_a_mepb.__init__(self, symbol_pool)
+
+
+if __name__ == "__main__":
```
serpilliere: I know that it's classic in python to include test in module, but can you seprate this code and move it in the 'example' directory, or in the test directoty, and to add a call to it in `test_all.py` ?

miasm2/arch/mep/sem.py
guedou: Is there a way to make the constant seen by the sembuilder ?

I did no manage to do it, that's why I am using my own, see
https://github.com/cea-sec/miasm/pull/763/files#diff-199ffae7a66bdc2eeb8cc43346463088R941

guedou:
I can remove __main__ blocks. It is quite similar to the unit tests.

guedou:
Thanks!

I pushed two commits that fix everything.

serpilliere:
Perfect!

Can you add our test_ir.py in the test_all.py , and squash all 4 commits?
(the details of the commits are not really important here)

guedou:
Done =)

serpilliere:
Ok!

In the future, one could add the `REPEAT`  handling in gen_post_code for llvm.
This will allow llvm jitter for MeP

serpilliere:
Thank YOU!

PR #766, opened 2018-06-15
Example/IDA: fix depgraph api

serpilliere:


example/ida/depgraph.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, ira):
         self.Compile()
 
     @property
-    def label(self):
+    def loc_key(self):
         value = self.cbBBL.value
         for real_label in self.ira.blocks:
             if str(real_label) == value:
```
commial: The code should probably also reflect the 'label' - > 'loc_key' change

serpilliere: Correct!

PR #772, opened 2018-06-19
Jitter: add simple trace api

serpilliere:


miasm2/jitter/jitload.py
```diff
@@ -484,3 +484,36 @@ def eval_expr(self, expr):
         self.symbexec.update_cpu_from_engine()
 
         return ret
+
+    def set_trace_log(self,
+                      trace=False,
```
commial: I think this argument is no more needed, and the docstring could be rewritten accordingly.
In addition, it would be great if `trace_instr` and `trace_regs` are set to `True` by default, because this likely the default behavior

commial: Additionally, is there a specific purpose for the `kwargs` form?

PR #774, opened 2018-06-19
Jitter: rename jitter class into Jitter

serpilliere:


miasm2/jitter/jitload.py
```diff
@@ -484,3 +485,16 @@ def eval_expr(self, expr):
...
+    """
+    DEPRECATED object
+    Use Jitter instead of jitter
+    """
+
+
+    def __init__(self, *args, **kwargs):
+        warnings.warn("Deprecated API: use Jitter")
+        super(jitter, self).__init__(
+            ir_arch, jit_type=jit_type
```
commial: `*args and **kwargs` have to be used here

Issue #775, opened 2018-06-20
Readme out of date


Frky:
The examples given in the readme file seem to be out of date: when iterating over blocks, I have the following error: 

```
>>> blocks = mdis.dis_multiblock(0)
>>> for block in blocks:
...  pass
... 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "miasm2/core/asmblock.py", line 637, in __iter__
    raise RuntimeError("DEPRECATED")
RuntimeError: DEPRECATED

```

Also, is it normal that the link to clone Miasm repo refers to serpillere's repo instead of cea-sec (line 443)?

commial:
Done, thanks again!

PR #771, opened 2018-06-19
All: fix var name to reflect object types

serpilliere:

PR #776, opened 2018-06-20
Update README.md

Frky:
Proposed fix for Issue #775


serpilliere:
Hi!

The fix is already present in the PR #771 which is actually under review :smile: 
Are you trying to win a race condition ? :smile: 
But your fix on the URL is correct!

commial:
Nice spot!
Please rebase to reflect only the URL change

PR #777, opened 2018-06-20
Adding ccmp aarch64 instruction.

GAJaloyan:
adding the semantics for ccmp instruction in aarch64 architecture.


serpilliere:
I think you also have a corresponding modification in `aarch64/arch.py` which has not been pushed.
Is it correct ?

GAJaloyan:
I changed the original patch so that only sem.py gets modified. This version redefines the immediate to the appropriate size in sem.py.

serpilliere:
Thank you!

PR #780, opened 2018-06-21
Expr: add simplification shift mask

serpilliere:


commial:
Maybe I don't correctly understand the simplification here, but if 'int1' and 'int2' are different, I don't see how just a mask could be enough

miasm2/expression/simplifications_common.py
```diff
@@ -250,6 +250,26 @@ def simp_cst_propagation(e_s, expr):
             e_s(Y.msb()) == ExprInt(0, 1)):
             args = [args[0].args[0], X + Y]
 
+    # ((var >> int1) << int2) => var & mask
```
commial: This is just a detail, but it would be great if the comment reflects `int1` and `int2` equality, that is to say something like "# ((var << int1) >> int1) => var & mask"

commial:
I'm trying to check the formula with z3, but it get me a counter example (again, I'm may be wrong):

```Python
# $ python -i test/expression/simplifications.py --z3
>>> s = z3.Solver()
>>> s.add(trans.from_expr(b) < 32)
>>> s.add(((trans.from_expr(a) >> trans.from_expr(b)) << trans.from_expr(b)) != (trans.from_expr(a) & ((1 << (32 - trans.from_expr(b))) - 1)))
>>> s.assertions()
[b < 32, (a >> b) << b != a & (1 << 32 - b) - 1]
s.check()
sat
>>> s.model()
[b = 1, a = 2148556834]
```
And indeed:
```Python
# Origin
>>> hex(2148556834)
'0x80106022'
>>> (0x80106022 >> 1) << 1
2148556834

# Formula application
>>> 0x80106022 & ((1 << (32 - 1)) - 1)
1073186
>>> hex(1073186)
'0x106022'
```

serpilliere:
I think you made an error in the interpretation of the reduction:
In the case you have tested, the expression's form is:
```
(a >> b) << b
```

Here, in python, the op_name is "<<": this is the first "node" of the expression tree.
But in your test case, you test the result of the second case:
```
a & (1 << 32 - b) - 1
```
is the correspondance for:
```python
            if op_name == '>>':
                mask = ExprInt((1 << (expr.size - int(int1))) - 1, expr.size)
```

Are you ok with this?
Maybe you can test both cases with the z3 tool to be sure :smile: 

I am modifying the comment to reflect the code

commial:
Indeed, my mistake.

I've checked both with z3, and it's correct.

PR #786, opened 2018-06-30
Fix mips32 dstflow2label()

guedou:
Here is a small PR that I encounter while playing with the following small binary:
```
J          0x4
NOP
```

### DIFF
--- a/miasm2/arch/mips32/arch.py
+++ b/miasm2/arch/mips32/arch.py
@@ -97,8 +97,8 @@ class instruction_mips32(cpu.instruction):
         if self.name in ["J", 'JAL']:
             expr = self.args[0].arg
             addr = (self.offset & (0xFFFFFFFF ^ ((1<< 28)-1))) + expr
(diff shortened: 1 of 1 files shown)


miasm2/arch/mips32/arch.py
```diff
@@ -98,7 +98,7 @@ def dstflow2label(self, symbol_pool):
             expr = self.args[0].arg
             addr = (self.offset & (0xFFFFFFFF ^ ((1<< 28)-1))) + expr
             label = symbol_pool.getby_offset_create(addr)
-            self.args[0] = ExprLoc(label.loc_key, expr.size)
+            self.args[0] = ExprLoc(label, expr.size)
```
serpilliere: Nice catch!
Can you rename as well the "label" to "loc_key" to reflect the object type in the patch ?
Thank you!

guedou: Done.

This was discovered thanks to a r2m2 unit test =)

PR #797, opened 2018-07-12
Fix ARM C function calling

aguinetqb:
Arguments were not pushed correctly on the stack (for functions with
more than four arguments)


serpilliere:
Thanks!

PR #803, opened 2018-07-17
Feature/linux env

commial:
This PR:
- add support for x86 relocations (only basic, without consideration and symbol import in a LocationDB)
- add a module to simulate (naively) a Linux environment (file system, system info, ...)
- add a module to stub some x86_64 syscalls
- add a `dis_instr` method to DisasmEngine to complete `AsmBlock` and `AsmCFG` levels
- add few others CPUID
- add an example using these elements to emulate `file /bin/ls` from end to end

Note: for now, the syscall logic is mainly in syscall function. With the future implementation of syscall for others architectures, some of the code logic will probably move to `LinuxEnvironment` to keep only architectures specificities


Summus6:
Are you really interacting with **real** filedescriptors, files and sockets ?!
(I didn't look more than these "features" for now)

commial:
Indeed, this is a good question. This is a first step implementation, and it is probably not enough sandboxed. Files are really opened, but corresponding FD is not the one exposed to the program.
There is an intermediate layer which handle, and can therefore filter, uses of this FD. For instance, write operations in FD others than STDOUT and STDERR aren't yet performed nor authorized.
But still, one can "damage" a file by opening it with the appropriate flags.

The files are actually retrieved from a "chroot-like" environment, like in Windows emulation. That is to say, if the target program open "/bin/ls", it is actually "current_dir/file_sb/bin/ls" which is open, and it is on this file that action are performed.

Long story short, it would be preferable to completely abstract system calls, but this (at least at a first look) look like a big step further.
Still, I'm completely open to remarks and comment.

commial:
The PR has been updated to remove most of the real syscalls by stubbing them (ie, `stat`s ones and a few `ioctl`).

To give a maybe more comprehensive answer to @Summus6, remaining *real syscalls* are used only to interact with the filesystem (open with controlled flags, list directory, read from file, get file size, read link). The sandboxed filesystem is abstracted by the `FileSystem` class. It is responsible of wrapping the *real filesystem* and offers API acting like a chroot in it.
That way, the external use of this object (like in the syscall stubs) can "forget" the filesystem abstraction, and have less chance of using dangerous API.

Regarding the implementation, in the future, the FileSystem could be rewritten to offer a FUSE-like API, like the https://github.com/fusepy/fusepy one. Then, the `LinuxEnv` will be in charge to correctly transform filesystem syscalls into FUSE API calls. It would be also possible to use existing fusepy implementation or mounting the sandboxed one.

Summus6:
It's probably better than I thought the first time ^^'
But you're still doing "real" open on files (even if the emulated code does not access it directly) and I don't see a limit in your implementation.
A malicious (or bug) code may loop over an open syscall to block the system, no ?

commial:
> It's probably better than I thought the first time ^^'

Thanks, I'm happy to read that.

>  A malicious (or bug) code may loop over an open syscall to block the system, no ?

You're right. Unfortunately, I don't see any simple way to solve this problem, except adding a limit to the number of file descriptor created (which is something a user can still easy add if needed). I'm still open to suggestion if you have any.

Exhausting some of the host resource is already possible through other way (allocating huge amount of memory, infinite loop, etc.), so if there is no a proper / simple way to fix it, I'll consider it OK for now.

serpilliere:
I agree with you @commial : Linux & co already have a default limit on file descriptors so at worst, the user could change this limit to avoid those situation (same for memory consumption)
Does it seems legit to you @Summus6 ?

Summus6:
In my code I added a limit on the number of syscalls but it is a personal choice.
In a other way, as you suggested, you may limit the number of filedescriptor and return -1.
You can also limit the number of mmap, socket, etc.
Why ? If the malicious code do that in purpose, he may does something else after reaching the limit (when return is -1)

But, if we don't want to care about that, letting the user be careful (as he should be analyzing malicious code) is legit :)

PR #804, opened 2018-07-17
Support of ARM SVC in the Miasm VM

aguinet:
This is mainly done by copy-pasting things from the X86 "int" implementation. It looks like some of these stuff could be factorized. Do you want me to try something like that?

It also misses one test I need to polish.

FTR I used it in a recent study, and it worked in this case!


aguinet:
Ok thanks! This was more a "Request for comments" than a real PR :) I'll push the test today or tomorrow!

PR #809, opened 2018-07-19
Expression: fix slice assignment

serpilliere:


commial:
Wouldn't it better to have `expr[0:expr.size] -> expr` ? Idem for `ExprCompose(one_expr) -> one_expr`.

PR #812, opened 2018-07-26
Incorrect ir_mepb.addrsize definition

guedou:
Here is a small fixes, that allows to use Sibyl.


commial:
It also fix the "mep_l" jitter, thanks

PR #816, opened 2018-08-06
Operator high level

serpilliere:
New features:
- high level operators
- expression propagation


commial:
As a side note (for those using a custom expression simplifier), you may encounter new, unsimplified, operations as `FLAG_*` or `signExt_*`. You may want to enable  `PASS_HIGH_TO_EXPLICIT` passes to get back the old behavior.

serpilliere:
Noted!

PR #817, opened 2018-08-07
adding a new simplification for double zeroExtends.

GAJaloyan:
With the new push that beautifies the operators, there are a lot of double zeroExtends (zeroExt_64(zeroExt_32(..))) that have appeared in the IR. This commit simplifies them by removing the innermost zeroExtend. 
It could be possible to do the same with signExtend, but I have not found any such case (but there are some zeroExt_64(signExt_32), which may not be easy to simplify.


serpilliere:
Correct!
Can you add the twin reduction with "signext of signext" in the same commit?

GAJaloyan:
added the double signext also.

PR #827, opened 2018-08-20
Fix mips

serpilliere:


miasm2/arch/mips32/sem.py
```diff
@@ -76,23 +76,23 @@ def lb(arg1, arg2):
...
+    dst = arg3 if ExprOp("==", arg1, arg2) else ExprLoc(ir.get_next_break_loc_key(instr), ir.IRDst.size)
     PC = dst
     ir.IRDst = dst
 
 @sbuild.parse
 def bgez(arg1, arg2):
     """Branches on @arg2 if the quantities of register @arg1 is greater than or
     equal to zero"""
-    dst = ExprLoc(ir.get_next_break_loc_key(instr), ir.IRDst.size) if arg1.msb() else arg2
+    dst = ExprLoc(ir.get_next_break_loc_key(instr), ir.IRDst.size) if ExprOp('s<', arg1, ExprInt(0, arg1.size)) else arg2
```
commial: Typo in 's<'

serpilliere: Done.
I added a patch which uses TOK_* instead of strings as you suggested!

PR #829, opened 2018-08-23
X86: fix float

serpilliere:


miasm2/ir/translators/C.py
```diff
@@ -202,7 +202,11 @@ def from_ExprOp(self, expr):
...
             elif expr.op.startswith("sint_to_fp"):
                 dest_size = expr.size
-                arg_size = expr.args[0].size
+                arg = expr.args[0]
+                if arg.size < dest_size:
+                    # extend source first
+                    arg = arg.signExtend(dest_size)
+                arg_size = arg.size
                 if (arg_size, dest_size) in [
                         (32, 32), (64, 64), (32, 64),
```
commial: Due to the `signExtend` use, `(32,64)` case is no more reachable. Extending the size of the expression before passing through the floating point operation (ie. your way) is probably a better idea than the current implementation. So there is no more need for some of the corresponding C functions, and consideration for the case `arg_size != dest_size`.

serpilliere:
`sint_to_fpXX` is now replaced by `sint_to_fp`, and its source is sign extended to the final size. So `sint_to_fp` is not a size exception expression anymore.

PR #838, opened 2018-08-30
Remove malware sample

serpilliere:


commial:
For those who will later encounter a Virus Alert, **this is not an actual**, unintentionnal, **malware**. It's a connect back example, to 192.168.1.1.
One may ends with a virus alert on "git clone" due to git getting back the history.

PR #852, opened 2018-09-11
Code cleaning: lgtm.com

serpilliere:


miasm2/analysis/depgraph.py
```diff
@@ -195,6 +198,8 @@ class DependencyResult(DependencyState):
     """Container and methods for DependencyGraph results"""
 
     def __init__(self, ircfg, initial_state, state, inputs):
+
+        super(DependencyResult, self).__init__(state.loc_key, state.pending)
```
commial: `self.links = state.links` could already be done in the init and then removed from the current init, together with `_graph` init

miasm2/arch/ppc/sem.py
```diff
@@ -294,10 +294,10 @@ def mn_do_lswi(ir, instr, rd, ra, nb):
     if nb == 0:
         nb = 32
     i = 32
-    raise "%r not implemented" % instr
+    raise NotImplementedError("%r not implemented" % instr)
```
commial: Its better to raise a `RuntimeError` (as we do in others miasm code), because `NotImplementedError` is intended to be used for abstract method

miasm2/jitter/llvmconvert.py
```diff
@@ -398,7 +398,7 @@ def memory_write(self, func, addr, size, value):
             builder.store(value, ret)
             value_ptr = builder.bitcast(ret, llvm_ir.IntType(8).as_pointer())
 
-            ret = builder.call(
+            _ = builder.call(
```
commial: As you do in this same commit but for another file, the returned value can be ignored, so the `_ = ` be removed (but I note the Rust-ism :smile: )

PR #857, opened 2018-09-24
Update README.md

trietptm:
Update examples in README.md for these errors:
AttributeError: 'NoneType' object has no attribute 'get_or_create_offset_location'
TypeError: __new__() takes exactly 3 arguments (2 given)

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,7 +277,7 @@ Symbolic execution
 Initializing the IR pool:
 
 ```pycon
->>> ira = machine.ira()
+>>> ira = machine.ira(loc_db)
 >>> ircfg = ira.new_ircfg_from_asmcfg(asmcfg)
 ```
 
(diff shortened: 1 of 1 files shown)


trietptm:
Yah, I think a Python notebook would be good but I think there're not many changes. If it frequently raises errors because of updates to the framework, there's a bigger problem to the design :) .

serpilliere:
Hi!

Yes, there are design errors in Miasm: The code base was (and is still) a mess, and we are gradually trying to clean/improve it. We try to add "clean" code for new features, and to rewrite bad legacy code.

By doing this, some modules are rewritten to be cleaner and can use the same API, some modules need a complete API rewritting.

In the second case, we are trying to do our best to support old API. In the case we cannot do this, we raise an exception in which we explain how to fix the "client" code in order to match the new API.
Some very rare cases do a brute API modification/removal, indeed.

Additionally to this, there are bad french translations which are under modifications: That's why for example there is still a module called `asmbloc.py` (without the k) and the new `asmblock.py`. Today, it's been nearly 2 years we kept it to have a bit of "retro" compatibility, but we may remove it in a near future.
The last one for example is ExprAff (which stands for Expr Affectation which, in french, means expr assignment). But hopefully, we chose to use ExprAssign (instead of ExprAss)

We will try in the future to use and conform to a real CHANGELOG to explain all the things we may modify in order to keep things clear!

Hope it helps a bit on our intents and purposes :smile:

PR #858, opened 2018-09-27
In ARM & Aarch64 MSR and MRS instructions' exception wrong names used

su-vikas:
A minor bug fix, in ARM and Aarch64, in the respective exceptions of MRS and MSR, instruction names have been interchanged. Fixed it in the PR.


commial:
Thanks!

Issue #210, opened 2015-08-15
Support arm/armthumb switch during emulation


serpilliere:
For the moment the emulation is linked to only one architecture. A problem raises in arm emulation since executing a `BLX` switches from one arch to another.
Some modification are needed in the `jitter` to support such behavior.

hax0kartik:
I'm interested in adding this feature, could you point me to some docs or proposed changes from which I can adapt?

serpilliere:
Ok.
We have some ideas on this.
One of them, which may not be the most generic (as it solves only the problem for arm/armt) should be to merge the arm/armt in the *same* architecture. And the instruction disassembler by itself will chose which instruction set will be set, depending on the offset (like the arm processors does in fact)

I think it may imply to modify a little bit at least one API (the disassembler one) to force it ta always take a `bin_stream` as parameter.

But I am afraid it's not the simplest task to start with miasm :smiley:

PR #859, opened 2018-09-27
BLX ignore bit 0 as part of the address to branchto

jetsecurity:
BLX bit 0 indicates whether the destination code is interpreted as Thumb code.
However, Thumb code or not, bit 0 should be ignored to compute the destination address.


serpilliere:
Hum. This one is abit more difficult. See #210  and sub issues:
If we simply remove the bit in exec time, we won't be able to detect thumb/not thumb changes.
The patch is a bit more difficult, and we don't have any solutions for now :disappointed:

serpilliere:
It makes sense to me. But it may mask bugs in thumb -> arm attemps :disappointed:

PR #866, opened 2018-10-11
Bin_stream: add get_u8/u16/u32/u64

serpilliere:


miasm2/core/utils.py
```diff
@@ -12,6 +12,30 @@
 pck32 = lambda x: struct.pack('I', x)
 pck64 = lambda x: struct.pack('Q', x)
```
commial: Should we deprecate `pck64`, `pck32`, etc.? The endianess not being precised, the execution will depend on the host, not the best option....

serpilliere: Not for now: this patch is not intended to fix the endiannes everywhere.
Some code may use them elsewhere.

PR #869, opened 2018-10-12
Sandbox: change use_seh

serpilliere:
Option name change: use-windows-structs instead of use-seh.


commial:
Hello,

This was the purpose of #542. Let's agree on a name, and then update accordingly the option, relative APIs and the implementation file's name.

I'm not a big fan of "windows struct", because it does more (and what does "enabling windows structures" means?).
This option indeed add the TEB, PEB, SEH structures, but also, for instance, error handling. IHMO, this is more about the context/environment in which the process is emulated which behave like the Windows one.

PR #871, opened 2018-10-12
Simple: add multiplication simplification

serpilliere:
Best effort to simplify simple add/multiply operators:
```
a + a*3 => a*4
a*2 + a << 2 => a*6
...
```
This is useful as compilators often replace expensive multiplications by shift/add operations.

### DIFF
--- a/miasm2/expression/simplifications_common.py
+++ b/miasm2/expression/simplifications_common.py
@@ -308,6 +308,12 @@ def simp_cst_propagation(e_s, expr):
             return -ExprOp(op_name, *new_args)
         args = new_args
 
+    # -(a * b * int) => a * b * (-int)
+    if op_name == "-" and args[0].is_op('*') and args[0].args[-1].is_int():
+        args = args[0].args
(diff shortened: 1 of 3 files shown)


serpilliere:
Maybe one day we will have a real expander/factorizer, but this is an enhancement.

miasm2/expression/simplifications_common.py
```diff
@@ -1138,3 +1144,72 @@ def simp_slice_of_ext(expr_s, expr):
...
+            arg = arg.args[0]
+            if arg.is_op('<<') and arg.args[1].is_int():
+                base_expr, factor = arg.args
+                operands[base_expr] = operands.get(base_expr, 0) - (2 ** int(factor))
+            else:
+                operands[arg] = operands.get(arg, 0) - 1
+        else:
+            operands[arg] = operands.get(arg, 0) + 1
+    out = []
+
```
commial: As this code will be called for each additions, an early break right here could be interesting.
For instance, a boolean marking if `<<` or `*` has been encountered at least once, or if `len(operands) == len(expr.args)`

serpilliere: Camille, early optimization is the root of all evil.

PR #879, opened 2018-11-08
X86 sem : mov assignexpr rot / shift

w4kfu:
When an operation of type shift / rotate is translated a temporary basic block is generated to handle 0-rotate / 0-shift.
Thus when working in 64-bit mode with a 32-bit operand, the result will never be zero-extended to 64-bit.
E.g:

```py

main:
MOV        RAX, 0x00
MOV        RCX, RAX
MOV        RAX, 0x4142434445464748
MOV        RDX, RAX
ROR        EDX, CL
RET
''')
asmblock.asm_resolve_final(mn_x86, asmcfg, loc_db)
ir_arch = ir_a_x86_64(loc_db)
ircfg = ir_arch.new_ircfg_from_asmcfg(asmcfg)
symb = SymbolicExecutionEngine(ir_arch)
cur_addr = symb.run_at(ircfg, 0x00)
assert  symb.symbols[mn_x86.regs.RDX] == ExprInt(0x45464748, 64)
```

I don't know if it is the best option but I decided to update flags only in the temporary block and destination is updated in the original block.

Also attached another commit that fix a typeerror in the C translator (`TypeError: %d format: a number is required, not str`)

Waiting for your input.

### DIFF
--- a/miasm2/arch/x86/sem.py
+++ b/miasm2/arch/x86/sem.py
@@ -636,17 +636,18 @@ def _rotate_tpl(ir, instr, dst, src, op, left=False):
             m2_expr.ExprAssign(of, new_of),
             m2_expr.ExprAssign(dst, res)
             ]
+    e = []
+    if dst.size == 32 and dst in replace_regs[64]:
+        e.append(m2_expr.ExprAssign(dst[:dst.size], dst))
     # Don't generate conditional shifter on constant
     if isinstance(shifter, m2_expr.ExprInt):
         if int(shifter) != 0:
             return (e_do, [])
         else:
-            return ([], [])
-    e = []
+            return (e, [])
     loc_do, loc_do_expr = ir.gen_loc_key_and_expr(ir.IRDst.size)
     loc_skip = ir.get_next_loc_key(instr)
     loc_skip_expr = m2_expr.ExprLoc(loc_skip, ir.IRDst.size)
-
     e_do.append(m2_expr.ExprAssign(ir.IRDst, loc_skip_expr))
     e.append(m2_expr.ExprAssign(
         ir.IRDst, m2_expr.ExprCond(shifter, loc_do_expr, loc_skip_expr)))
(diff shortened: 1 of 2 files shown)


serpilliere:
Hi,
Holy cow! You got a big one here.
I agree with you, there is definitely something missing here.
I am currently reviewing the fix.

Thank you @w4kfu !

serpilliere:
Erf, I think there is a missing case in your fix. If the instruction is something like:
```
ROR EAX, 0
```
The code doesn't generate the EAX zero extend.
Do you agree?

w4kfu:
Hello,

I just tested right now, and it seems to work.

My first attempt commit : 7f2cd33f49a59e42a7bdf8cf5e3ba20b9580dc67 was a failed.

But f55c0e7aa61ac5d8ffd11b291edd7cb715c314a5 does the trick like you mention, check : https://github.com/cea-sec/miasm/blob/f55c0e7aa61ac5d8ffd11b291edd7cb715c314a5/miasm2/arch/x86/sem.py#L647 ?

No?

w4kfu:
Hello,

I just tested with this example:

```

main:
MOV        RAX, 0x4142434445464748
ROR        EAX, 0x00
RET
''')
asmblock.asm_resolve_final(mn_x86, asmcfg, loc_db)
ir_arch = ir_a_x86_64(loc_db)
ircfg = ir_arch.new_ircfg_from_asmcfg(asmcfg)
symb = SymbolicExecutionEngine(ir_arch)
cur_addr = symb.run_at(ircfg, 0x00)
assert  symb.symbols[mn_x86.regs.RAX] == ExprInt(0x45464748, 64)
```

And it seems to work.

FYI, my first attempt commit (7f2cd33f49a59e42a7bdf8cf5e3ba20b9580dc67) was not the right way to handle this problem.

The second one (f55c0e7aa61ac5d8ffd11b291edd7cb715c314a5) call the zero extend function "everytime" a 32 bits operand is used.

Regarding the constant 0x00 for your "ROR EAX, 0x00", check https://github.com/cea-sec/miasm/blob/f55c0e7aa61ac5d8ffd11b291edd7cb715c314a5/miasm2/arch/x86/sem.py#L647

Maybe I should also had a check if it is a 32 bits operand in the 0x00 case contant in _rotate_tpl, rotate_with_carry_tpl & _shift_tpl?

serpilliere:
Ho!

Correct. I missed this. (sorry)
Seems legit to me. Is it ok for you to squash the commit   7f2cd33f4 and f55c0e7a?

w4kfu:
Hello,

Commit squashed and cleaned up, I should have done it before.

serpilliere:
Perfect :smile: 
Thanks again for the fix @w4kfu !

PR #882, opened 2018-11-13
Revert 879 mov assignexpr rot shift ; Simplify assignexpr rot shift & patch simplification

w4kfu:
Hello,

You are gonna hate me but there is still some problems...

My test case was not good enough because the same bug is still present when using the GCC JIT.

I didn't catch the issue because when the IR was displayed, the assignation was as follow:

`RDX = zeroExt_64(RDX[0:32])`

But in fact the operation returned was:

`ExprAssign(ExprId('EDX', 32), ExprId('EDX', 32))`

So under the python jitter, the zeroextend was done, but not under GCC JIT, the operation disappeared in the generated C file.

Now the fix create the following asssignation expression:

`ExprAssign(ExprId('RDX', 64), ExprOp('zeroExt_32', ExprId('EDX', 32)))`

Now the test seems ok with python or GCC JIT.

Sorry for the inconvenience.

### DIFF
--- a/miasm2/arch/x86/sem.py
+++ b/miasm2/arch/x86/sem.py
@@ -638,7 +638,7 @@ def _rotate_tpl(ir, instr, dst, src, op, left=False):
             ]
     e = []
     if dst.size == 32 and dst in replace_regs[64]:
-        e.append(m2_expr.ExprAssign(dst[:dst.size], dst))
+        e.append(m2_expr.ExprAssign(dst, dst))
     # Don't generate conditional shifter on constant
     if isinstance(shifter, m2_expr.ExprInt):
         if int(shifter) != 0:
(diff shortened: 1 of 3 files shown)


w4kfu:
I tracked down the problem, why in the Python jitter it was working and not under GCC.

When using the python jitter, the function `post_add_asmblock_to_ircfg` will be called and function `irbloc_fix_regs_for_mode` will be called in order to replace `EAX = EAX` by `RAX = zeroExt_64(RAX[0:32])`.

When using GCC jitter, the funtion `block2assignblks` is called, inside this function a simplification passes is called (cf: https://github.com/cea-sec/miasm/blob/0f2f1b094b663caa4df3a375858e4a93df0227d0/miasm2/jitter/codegen.py#L173), the affectation is deleted.

By calling `irbloc_fix_regs_for_mode` before simplification seems to resolve the issue, but don't know if it is the best option:

```
            for irblock in irblocks:
                new_irblock = self.ir_arch.irbloc_fix_regs_for_mode(irblock, self.ir_arch.attrib)
                new_irblock = new_irblock.simplify(expr_simp_high_to_explicit)[1]
                out.append(new_irblock)
```

w4kfu:
Now only add assignation dst = dst for `ROL`/`ROR`/`RCL`/`RCR`/`SAR`/etc...

Add `fix_regs_for_mode` after getting the ir of the instruction and remove it in `gen_c`

serpilliere:
Seems ok for me.
Thanks for the fix of the fix @w4kfu !

PR #892, opened 2018-11-28
dis_multiblock can take multiple offsets as input

nguigo:
This change allows to pass an iterable to dis_multiblock. Besides convenience, it also allows the disassembly engine to transparently split overlapping blocks reachable from multiple roots. Backward compatibility is also preserved.

### DIFF
--- a/miasm2/core/asmblock.py
+++ b/miasm2/core/asmblock.py
@@ -1551,7 +1551,7 @@ class disasmEngine(object):
         warnings.warn('DEPRECATION WARNING: use "dis_block" instead of "dis_bloc"')
         return self.dis_block(offset)
 
-    def dis_multiblock(self, offset, blocks=None):
+    def dis_multiblock(self, offsets, blocks=None):
         """Disassemble every block reachable from @offset regarding
         specific disasmEngine conditions
         Return an AsmCFG instance containing disassembled blocks
@@ -1563,16 +1563,19 @@ class disasmEngine(object):
         job_done = set()
         if blocks is None:
             blocks = AsmCFG(self.loc_db)
-        todo = [offset]
+        try:
+            todo = list(iter(offsets))
+        except:
+            todo = [offsets]
 
         bloc_cpt = 0
         while len(todo):
+            target_offset = int(todo.pop(0))
             bloc_cpt += 1
             if self.blocs_wd is not None and bloc_cpt > self.blocs_wd:
-                log_asmblock.debug("blocks watchdog reached at %X", int(offset))
+                log_asmblock.debug("blocks watchdog reached at %X", int(target_offset))
                 break
 
-            target_offset = int(todo.pop(0))
             if (target_offset is None or
                     target_offset in job_done):
                 continue


nguigo:
The second change ensures the dependency search loop concludes the current iteration if a requested stop head has been reached.

serpilliere:
Hi!
I'm a bit reluctant to this one. Currently, the `dis_multiblock` takes an offset and gives back an asmcfg, which is a kind of function. Taking multiple offsets will give a very strange object: multiple heads, maybe disjoin graph (!), ... which does not seem to be a "classic" use case.

So maybe the best solution if you really want this is to keep the API unmodified, but to call multiple times dis_multiblock, and merging yourself the resulting asmcfgs. This may be achieved using `asmcfg.merge` + `asmcfg.apply_splitting`. 

Is it OK for you?

nguigo:
Hi again :)

I used this patch to get past jump tables and get all blocks from a single function in the same graph (albeit disjoint). The issue with the iterative _dis_multiblock()_ call workaround you are suggesting is that the splitting only works against blocks from the same call (due to the pending state management). 

Would you favor a fix located within the _apply_splitting()_ function?

nguigo:
Would you favor a fix around _apply_splitting()_ pending state management?

serpilliere:
Hi!
I will do some tests on the apply splitting. Give me a bit more time :smile:

serpilliere:
Hi @nguigo !
I have some news.
I did some tests, And maybe we can do this:
```
diff --git a/miasm2/core/asmblock.py b/miasm2/core/asmblock.py
index 34f11474..8a4a24f1 100644
--- a/miasm2/core/asmblock.py
+++ b/miasm2/core/asmblock.py
@@ -1551,7 +1551,7 @@ class disasmEngine(object):
         warnings.warn('DEPRECATION WARNING: use "dis_block" instead of "dis_bloc"')
         return self.dis_block(offset)

-    def dis_multiblock(self, offset, blocks=None):
+    def dis_multiblock(self, offset, blocks=None, job_done=None):
         """Disassemble every block reachable from @offset regarding
         specific disasmEngine conditions
         Return an AsmCFG instance containing disassembled blocks
@@ -1560,7 +1560,8 @@ class disasmEngine(object):
                 merge with
         """
         log_asmblock.info("dis bloc all")
-        job_done = set()
+        if job_done is None:
+            job_done = set()
         if blocks is None:
             blocks = AsmCFG(self.loc_db)
         todo = [offset]
```


So, the user code may be:
```python
from miasm2.analysis.binary import Container
from miasm2.core.asmblock import AsmCFG

cont = Container.from_string("\xff\xe0\x89\xc0\x89\xd8\xeb\x05\x89\xc8\xeb\x01\x90\xc3")
bs = cont.bin_stream

machine = Machine("x86_32")
mn, dis_engine = machine.mn, machine.dis_engine
ira, ir = machine.ira, machine.ir

mdis = dis_engine(bs, loc_db=cont.loc_db)
todo = [0, 2, 4, 8, 0xD, 0xC]

asmcfg = AsmCFG(mdis.loc_db)
job_done = set()
while todo:
    ad = todo.pop()
    asmcfg = mdis.dis_multiblock(ad, asmcfg, job_done)

for block in asmcfg.blocks:
    print block
open('out.dot', 'w').write(asmcfg.dot())
```

nguigo:
Just tried it out against additional test cases and it handles multiple paths + jumping in the middle of instructions gracefully. Thanks for taking the time to write this up! What do you think of the other commit?

serpilliere:
We are discussing about the good way to do this, so give us a bit more time :smiley:

nguigo:
Sounds good (didnt mean to imply urgency :relaxed:).

PR #894, opened 2018-12-03
Tests/qemu x86 64

commial:
Add sample, script and expected outputs for the QEMU x86_64 regression test.
The script and test behavior is basically the same than QEMU i386 tests.

For now, a few tests do not fully pass, while passing for i386:
* `xchg`
* `loop`
* `sse`

`fxsave` test (not passing) use specific x87 behavior, not emulated by Miasm.

`floats` test (not passing) requires the use of an external lib (libmath). While this is something doable now (since the last release), it raises floats divergence, for instance while running through sin / cos functions. Debugging them requires specific development, as Miasm traces and traces from a real system are needed (several operations are made before comparing the result with the expected QEMU one).


test/arch/x86/qemu/testqemu64.py
```diff
@@ -0,0 +1,133 @@
...
+    ret_ad, args = jitter.func_args_systemv(["out", "format"])
+    if args.out != 1:
+        raise RuntimeError("Not implemented")
+    fmt = jitter.get_str_ansi(args.format)
+    # Manage llx
+    fmt = fmt.replace("llx", "lx")
+    fmt = fmt.replace("%016lx", "%016z")
+
+    fmt_a = parse_fmt(fmt)
+    esp = jitter.cpu.ESP
```
serpilliere: Maybe this is `RSP` instead of `ESP` here?

PR #895, opened 2018-12-06
performance thing

Mizari:
Better replacing

### DIFF
--- a/miasm2/expression/expression.py
+++ b/miasm2/expression/expression.py
@@ -318,12 +318,10 @@ class Expr(object):
         @dct: dictionary of Expr -> *
         """
         if dct is None:
-            dct = {}
+            return self
 
         def my_replace(expr, dct):
-            if expr in dct:
-                return dct[expr]
-            return expr
+            return dst.get(expr, expr)
 
         return self.visit(lambda expr: my_replace(expr, dct))
 


Mizari:
buggy variable name

serpilliere:
I you are not awre, you don't need to close the PR: just push force your branch with a fixed version, it will automatically update the PR !

miasm2/expression/expression.py
```diff
@@ -318,12 +318,10 @@ def replace_expr(self, dct=None):
         @dct: dictionary of Expr -> *
```
serpilliere: After reading your patch, I wonder if the 'optional' argument is really interesting here.
I see no usecase where we want to call this function without an expression to replace.

So maybe we can change the api to have the `dct` mandatory.
What about this?

Mizari: Thats a good idea, I think.  
I also thought about using lighter version of this with only one mapping dst-src (I use it myself in my inherited expression class). This will boost perfomance a bit more for such simple replacings.

miasm2/expression/expression.py
```diff
@@ -318,12 +318,10 @@ def replace_expr(self, dct=None):
...
         """
         if dct is None:
-            dct = {}
+            return self
```
p-l-: Good point. While you are here, you might want to rewrite the test above as:
```Python
        if not dct:
            return self
```
(since there is no reason to do something either when `dct` is empty).

miasm2/expression/expression.py
```diff
@@ -318,12 +318,10 @@ def replace_expr(self, dct=None):
...
 
         def my_replace(expr, dct):
-            if expr in dct:
-                return dct[expr]
-            return expr
+            return dst.get(expr, expr)
 
         return self.visit(lambda expr: my_replace(expr, dct))
```
p-l-: Maybe since your (good) proposal to implement `my_replace()` is as simple as a function call, you could remove it and use here:
```Python
        return self.visit(lambda expr: dct.get(expr, expr))
```

PR #896, opened 2018-12-06
perfomance feature

Mizari:
Better replacing


serpilliere:
Hi,
Fine!
Are you ok to modify the `dct` argument to make it non optional as well, or do you prefer I merge this one now, and I will change the argument stuff in another PR?

Issue #823, opened 2018-08-16
SSA naive back-translation algorithm produces incorrect results


ecx86:
Hi! I remember I saw your talk at defcon.

So, I decided to take a look at the SSA destruction algorithm, since you mentioned that was one of the weak points of the project.

https://github.com/cea-sec/miasm/blob/master/miasm2/analysis/ssa.py

The problem is that this naive translation in which all variables `x_i` in a phi node `phi(x_1, x_2, x_3`, etc. are replaced with a single variable, `x`, produces incorrect results in some situations. Two famous examples of this in literature are the "swap problem" and "lost copy problems". Here is a good source on it: https://www.inf.ed.ac.uk/teaching/courses/copt/lecture-4-from-ssa.pdf

For example, since the semantics of phi statements is that all phi statements in the basic block are executed simultaneously, you could have for example:

```
y_1 = phi(y_0, x_1)
x_1 = phi(x_0, y_1)
```
here the variables `x` and `y` are swapped, but a naive backtranslation produces an incorrect result:

```
y = x
x = y
```
clearly the value of y is clobbered. You would need to introduce a temporary variable.

I'll contribute some code later to address this if I can. Anyways, great work on the project!

ecx86:
(shameless self-plug)

If you are looking for a reference implementation of some popular algorithms for it, there is [Sreedhar et al.](https://github.com/LLVM-but-worse/maple-ir/blob/master/org.mapleir.ir/src/main/java/org/mapleir/ir/algorithms/SreedharDestructor.java) and [Boissinot et al.](https://github.com/LLVM-but-worse/maple-ir/blob/master/org.mapleir.ir/src/main/java/org/mapleir/ir/algorithms/BoissinotDestructor.java).

serpilliere:
Hi!
Glad to hear from you :smiley:
I can't wait to see your  implementation!

By the way: in the example you gave, is it the very same assignblock (which means assignments are in parallel, and then x and y are really swapped), or is it a sequence of assignblock?

Thanks!

ecx86:
The phi statements are in parallel.

serpilliere:
Hi!


After an analysis done with @commial, it seems we don't have the swap problem, as the resulting created copies are also assigned in *parallel*.
For the lost copy, you are totally right: we did a little toy example, and it messes around!

Thanks for the issue!

serpilliere:
Hi!
Did you do some progress on the un-ssa stuff?
Cheers!

[...]

PR #899, opened 2018-12-08
SSA: fix ssa to cssa, and naive out of cssa

serpilliere:
This PR implements the common first step of "Translating Out of Static Single Assignment Form" and "Revisiting Out-of-SSA Translation for Correctness, Code Quality, and Efficiency" 

This patch fixes the current Miasm UNSSA algorithm, which is buggy.
 PR done with @commial. Big thanks to you man!

### DIFF
--- a/example/ida/graph_ir.py
+++ b/example/ida/graph_ir.py
@@ -2,23 +2,81 @@ import os
 import tempfile
 
 import idaapi
+import ida_kernwin
 import idc
+import ida_funcs
 import idautils
-
-from miasm2.core.bin_stream_ida import bin_stream_ida
 from miasm2.core.asmblock import is_int
+from miasm2.core.bin_stream_ida import bin_stream_ida
 from miasm2.expression.simplifications import expr_simp
+from miasm2.ir.ir import IRBlock, AssignBlock
 
-from miasm2.analysis.data_flow import dead_simp, DiGraphDefUse, \
-    ReachingDefinitions, merge_blocks, remove_empty_assignblks, \
-    PropagateExpr, replace_stack_vars, load_from_int, dead_simp, remove_empty_assignblks, \
-    read_mem, get_memlookup
+from miasm2.analysis.ssa import SSADiGraph, UnSSADiGraph, DiGraphLivenessSSA
 
-from miasm2.expression.simplifications import expr_simp
-from miasm2.analysis.ssa import SSADiGraph
-from miasm2.ir.ir import AssignBlock, IRBlock
+from miasm2.analysis.data_flow import dead_simp,  \
+    merge_blocks, remove_empty_assignblks, \
+    PropagateExpr, load_from_int
+
+
 from utils import guess_machine, expr2colorstr
-from miasm2.expression.expression import ExprLoc, ExprMem, ExprId, ExprInt
+
+
+
+
+TYPE_GRAPH_IR = 0
+TYPE_GRAPH_IRSSA = 1
+TYPE_GRAPH_IRSSAUNSSA = 2
+
+OPTION_GRAPH_CODESIMPLIFY = 1
+OPTION_GRAPH_DONTMODSTACK = 2
+OPTION_GRAPH_LOADMEMINT = 4
+
+
+class GraphIRForm(ida_kernwin.Form):
+
+    def __init__(self):
+        ida_kernwin.Form.__init__(
+            self,
+            r"""BUTTON YES* Launch
+BUTTON CANCEL NONE
+Graph IR Settings
+
+{FormChangeCb}
+Analysis:
+<Graph IR :{rGraphIR}>
+<Graph IR + SSA :{rGraphIRSSA}>
+<Graph IR + SSA + UnSSA :{rGraphIRSSAUNSSA}>{cScope}>
+
+Options:
+<Simplify code:{rCodeSimplify}>
+<Subcalls dont change stack:{rDontModStack}>
+<Load static memory:{rLoadMemInt}>{cOptions}>
+""",
+            {
+                'FormChangeCb': ida_kernwin.Form.FormChangeCb(self.OnFormChange),
+                'cScope': ida_kernwin.Form.RadGroupControl(
+                    (
+                        "rGraphIR",
+                        "rGraphIRSSA",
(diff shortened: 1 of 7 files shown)


serpilliere:
I remove the WIP as it seems we have interesting results here :smile:

serpilliere:
WIP because I am tracking a possible bug...

serpilliere:
Ok. It was a missuse of `immutable_ids` and `ssa_variables_expr`
Now, both are correctly used :smile: 

I used this PR to change the IDA `graph_ir` plugin, which is now a bit more customable and complete!

commial:
For those interested, here's what the GraphIR customization looks like with this PR:
ida

@ecx86, regarding #823 it seems that you've already played with Boissinot out-of-ssa. If you had time, it would be nice to hear your opinion on this implementation and/or if you have specific test cases that you want to try it on. To be more specific, this implementation also integrate the "phi-virtualisation" optimization indicated in the paper (to be able to have a pre-computed liveness in input).

example/ida/graph_ir.py
```diff
@@ -2,23 +2,81 @@
...
+
+{FormChangeCb}
+Analysis:
+<Graph IR :{rGraphIR}>
+<Graph IR + SSA :{rGraphIRSSA}>
+<Graph IR + SSA + UnSSA :{rGraphIRSSAUNSSA}>{cScope}>
+
+Options:
+<Simplify code:{rCodeSimplify}>
+<Subcall dont change stack:{rDontModStack}>
```
commial:
```suggestion
<Subcalls don't change stack:{rDontModStack}>
```

serpilliere:
@commial : I fixed the liveness analysis: the "phi" liveness was badly computed.
But there is a remaining bug in our implementation of the unssa algorithm: We have a minimalist example in `graph_8.dot` of `unssa.py` example.

My analysis: when we remove the `phi` of the `b` variable, we do this for each parents; here, lbl0/lbl2/lbl3. We generate the 'parallel copy' at the end of each of this parents; but when we generate the copy for block `lbl2`, we affect `b.0` which will interfere with the `b.0` at  `lbl3`

The liveness analysis seems good to me now. But I think  we have coalesced to quickly the `phi` predecessor variable (Maybe it's not clear, but it's not clear in my head neither...)
final_8

serpilliere:
Here is a graph (manually crafted...) with the parallel copy added:
ref_8_vars
And the same just afeter the trivial phi coalesced:
ref_8_coalesce
Here, it's clear we can coalesce `b` with `b.0`.
It seems we can coalesce `b` with `b.3` as well, but I don't see the possibility of coalescing `b` with `b.1`

serpilliere:
Ok, first bug: 
```python
                if merge_a_list[-1] < merge_b_list[-1]:
```
We are doing a 'string' comparison instead of comparing the `live_index` (lol)
But the result is not good anyway. We have to dig into the `live_index`...
I am pushing the modification to compare the `live_index`

serpilliere:
I think I have the bug: When we should create temporary variables, during the phi analysis, we should create pre `phi` variables (Here, `var0`, `var1`, `var2` for `b` phi, and  `var5`, `var6`, `var7` for `a` phi).
But we didn't create them, we only created post `phi`. So pre phi are not taken into account during the interference test: we merge `var0`, `var1`, `var2` with `var3` without putting `var0`, `var1`, `var2` in the new merge set.
The resulting interference computation is incorrect.

serpilliere:
The live check against the `varX` (which we don't need to create) is defined by the sentence:
```
Also,
to check the intersection with other variables, it is sufficient
to check the intersection at the parallel copies locations
```

But when we test the liveness of `b.0` and `var1`, we only test the liveness at the *post-phi* definition of `var1`, which does not comply with the paper: we have to check against *each* parallel copies, even the "pre-phi" locations.
So for liveness test, we have to check if one variable is a post phi, if so, check against every pre phi locations as well. Today, we check against only for one phi predecessor. this may be wrong

serpilliere:
I added the liveness test against pre phi definition, and it seems to work :tada: 
I will re analyse the code to be sure, and push the PR
Here is the result:
final_8
I may have missed a case as for me `a.0` and `a.2` should merge, but, erf, the result seems good :smile:

serpilliere:
Here is the last fix: I enable the expression propagation in the unssa regression tests, which generate interesting cases for it :smile:

commial:
Hey,

I agree with your different analyses:
* String comparison for merge_list... I guess it was working because variables are created in the dominance order, so `a.X` index were quite reflecting `live_index`...
* I do agree with the fact that we have to check for interference for pre-phi, and doing it in a way that all predecessors are checked before taking any decision
* Regrading your last example, I also agree with you. Live range of `a.2` is only in `lbl3`, so they should be able to interfere. It might help to have a look in the "interfere" test to check why they are considered as interfering

I should also have a look at the whole resulting code, because reviewing the different diffs make me loose the big picture.
I also wonder what happen if a given ssa var is used in two different phi, like:
```
                      a.1 = X
                     _______
                     /     \
| a.2 = phi(a.1, ...) |    | a.3 = phi(a.1, ...) |
```

I can't test the code for now, will do the next year.

serpilliere:
No problem @commial. I whish I pushed the different commits before merging them.
Maybe I can get back the temp commits.

I forgot to send the pic of the final last graph: I seems to be good.
final_8

serpilliere:
I have fixed an inversion which I did during some tests. This broke classic cases. I added another regression test to ensure this.
Note: In fact the regression tests are not "tested" here: I didn't find yet an interesting way of verifying the results.

miasm2/analysis/ssa.py
```diff
@@ -575,120 +574,505 @@ def _convert_phi(self):
...
-    Set the Assignblock in @graph at location @loc at @index using dictionary
-    @assignblk_dct
 
-    @graph: IRCFG instance
-    @loc: Location instance
-    @index: assignblock index
-    @assignblk_dct: dictionary representing the AssignBlock
+def get_var_assignment_src(ircfg, node, variables):
     """
+    Return a subset of @variables which are written by the irblock at @node
```
commial: It actually ensure and returns the only one variable of `variables` written in the assignblk (ie, the subset is always zero or one var)

PR #905, opened 2018-12-21
Enable easy pickling of dse snapshots

nguigo:
Pickling currently requires some additional piping but since `DSEEngine.restore_snapshot()` only requires the symbols dictionary (rather than the entire object), this small change allows to simply save DSE state on disk with:
```
snapshot = dse.take_snapshot()
pickle.dump(snapshot, file, pickle.HIGHEST_PROTOCOL)
```
and restore with:
```
snapshot = pickle.load(file)
dse.restore_snapshot(snapshot)
```

### DIFF
--- a/miasm2/analysis/dse.py
+++ b/miasm2/analysis/dse.py
@@ -390,7 +390,7 @@ class DSEEngine(object):
         snapshot = {
             "mem": self.jitter.vm.get_all_memory(),
             "regs": self._get_gpregs(),
-            "symb": self.symb.symbols.copy(),
+            "symb": {e: v for e,v in self.symb.symbols.items()}
         }
         return snapshot
 


commial:
Hi! Thanks for the PR. Indeed, you're right.
In the future, we should allow pickling of `symbols`, it would be also useful for other use cases (and several object in Miasm are already picklable).

Could you just relaunch AppVeyor test, just to be sure it is passing? (We sometime encounter issue with it)

nguigo:
Not sure how to do that, would amending the commit do the trick?

p-l-:
@commial I think you or @serpilliere can do that.

serpilliere:
Ok: we have a little problem with llvmlite, once it's ok, we will fix it and rerun the checks!

PR #907, opened 2018-12-22
Expression: add simplifications

serpilliere:


miasm2/expression/simplifications_common.py
```diff
@@ -1049,9 +1064,18 @@ def simp_zeroext_eq_cst(expr_s, expr):
         return ExprInt(0, 1)
     return ExprOp(TOK_EQUAL, src, ExprInt(int(arg2), src.size))
 
-def simp_ext_eq_ext(expr_s, expr):
-    # A.zeroExt(X) == B.zeroExt(X) => A == B
-    # A.signExt(X) == B.signExt(X) => A == B
+def simp_cond_zeroext(_, expr):
+    """X.zeroExt()?(A:B) => X ? A:B"""
+    if not (expr.cond.is_op() and expr.cond.op.startswith("zeroExt")):
```
commial: This is also true for `signExt` (ie `X.signExt` != 0 <=> `X` != 0)

miasm2/expression/simplifications_common.py
```diff
@@ -1075,13 +1099,101 @@ def simp_cond_eq_zero(expr_s, expr):
     new_expr = ExprCond(arg1, expr.src2, expr.src1)
     return new_expr
 
+def simp_sign_inf_zeroext(expr_s, expr):
+    """
+    X.zeroExt() <=s 0 => X == 0
+    X.zeroExt() <=s cst => X <=u cst[:X.size] (cond)
```
commial: First, this equation is not true if `X.zeroExt` is the same size of `X`, cause it would be rewrote as `X <=s cst => X <=u cst` (but it seems to be already checked by the code).

Second, maybe we can put it with a kind of more general rule:
X.zeroExt(size) with size > X.size is always positive.

A positive element <s or <=s X has three possibilities. Either X is always positive, and the 's' can be converted to a 'u' (so `X.zeroExt() <=s cst` => `X.zeroExt() <=u cst` if `cst` is positive, or `X.zeroExt <=s Y.zeroExt` => `X.zeroExt <=u Y.zeroExt`); either X is always negative, and this become always false or a comparison to 0; either we don't know and it can't be simplified.

With this rule, we can then add the rule `X.zeroExt() <=u cst` => `X <=u cst[:X.size]`, and `X <=u 0` => `X == 0`.

Those rules seem wider, more general, to me. What do you think?

Note: re-reading myself, I actually wonder if `X.zeroExt() <=s cst => X <=u cst[:X.size]` is even actually correct for negative `cst`. Like with X = 5, cst = 0xFFFFFFFF, 0x00000005 is not <=s 0xFFFFFFFF, but 0x5 is <=u 0xF

miasm2/expression/simplifications_common.py
```diff
@@ -1132,31 +1245,143 @@ def simp_ext_cst(expr_s, expr):
     return ret
 
 
-def simp_slice_of_ext(expr_s, expr):
-    # zeroExt(X)[0:size(X)] => X
-    if expr.start != 0:
-        return expr
+def simp_slice_of_ext(_, expr):
+    """
+    C.zeroExt(X)[A:B] => C[A:B] if B <= size(C)
```
commial: In the same style, `C.zeroExt(X)[A:B]` with `A >= size(C)` is 0.

miasm2/expression/simplifications_common.py
```diff
@@ -1132,31 +1245,143 @@ def simp_ext_cst(expr_s, expr):
...
+    if len(sizes) != 1:
+        return expr
+    size = list(sizes)[0]
+    args = [expr_s(arg[:size]) for arg in cond.args]
+    cond = ExprOp(cond.op, *args)
+    return ExprCond(cond, expr.src1, expr.src2)
+
+
+def simp_cond_sign_bit(_, expr):
+    """(a & .. & 0x80000000) ? A:B => (a & ...) <s ? A:B"""
```
commial: I imagine it's `(a & ...) <s 0 ...` (missing 0)

miasm2/expression/simplifications_common.py
```diff
@@ -1132,31 +1245,143 @@ def simp_ext_cst(expr_s, expr):
...
+    arg, cst = expr.args
+    if not (arg.is_op() and arg.op.startswith("signExt")):
+        return expr
+    if not cst.is_int():
+        return expr
+    base = arg.args[0]
+    tmp = int(mod_size2int[cst.size](int(cst)))
+    if -(1 << (base.size - 1)) <= tmp < (1 << (base.size - 1)):
+        # Can trunc integer
+        return ExprOp(expr.op, base, expr_s(cst[:base.size]))
```
commial: I'm not sure about this one, but as `tmp` comes from a bit to Python int translation, it should probably have to go from a Python int back to a bit representation at the end (using `mod_size2int` like transformation). I'm not sure the truncation is equivalent (it might be giving the range check).
You probably had a different view of this in mind, which is more explanatory of why it is working. If you can share it, that would help me to be sure it is indeed working :)

miasm2/expression/simplifications_common.py
serpilliere:
```
 I actually wonder if X.zeroExt() <=s cst => X <=u cst[:X.size] is even actually correct for negative cst
```
Not in all cases: if bits are stripped of during slice, it can be false:
```
ExprInt(0xFF, 8).zeroExt(16) <=s 0x7FF
```
cannot be sliced.


Anyway, I am re reading your comment to try to understand the final goal of the code.

miasm2/expression/simplifications_common.py
serpilliere: done

miasm2/expression/simplifications_common.py
serpilliere: Sorry but I dont understand your comment:
```
Those rules seem wider, more general, to me. What do you think?
```
What is/are your final reduction rule(s)?

Don't forget that  `simp_test_zeroext_inf` has additional conditions:
```
X.zeroExt() <=u cst => X <=u cst[:X.size]
```
Like previously, not valid in each case.

commial: We can add this rule:
`X.zeroExt() <=s cst` =>
* if cst is positive, `X.zeroExt() <=u cst`. Justification: signed comparison of positive number is the same than unsigned comparison
* if cst is strictly negative, it's always False. Justification: left part is a positive or null number, right part is strictly negative

This simplification is quite simple, and will report on the `simp_test_zeroext_inf` simplification the slice extraction / special case handling. In addition, we obtain the `False` case (when cst is strictly negative).

miasm2/expression/simplifications_common.py
serpilliere: Hum, I cannot parse:
```
but as tmp comes from a bit to Python int translation
```

The rule is this one:
```
A.signExt() <s int => A <s int[:]
```
Example:
```
ExprInt(-3, 8).signExt(32) <s ExprInt(12, 32)
```
Can be reduced in:
```
ExprInt(-3, 8) <s ExprInt(12, 8)
```
Idem for a negative constant:
```
ExprInt(-12, 8).signExt(32) <s ExprInt(-3, 32)
```
Can be reduced in:
```
ExprInt(-12, 8) <s ExprInt(-3, 8)
```
To check this, we need to verify if the constant (in signed form) fit in it's future truncated signed range.

If the constant fits in its future truncated signed range, this means the signed extend of this constant always gives the original constant value.
Exemple:
```
ExprId("A", 8).signExt(32) <s ExprInt(-3, 32)
```
* A is in range `[-128, 127]`
* `ExprInt(-3, 32)` is in this range `[-128, 127]`
* so we can compare against `ExprInt(-3, 8)`
* And effectively, `ExprInt(-3, 8).signExt(32)` gives `ExprInt(-3, 32)`

serpilliere: Note: we could better:
* If the value is bigger than the original range, the case is *always* true.
* if the value is lesser than the original range, the case is *always* false

Everything is the same for `simp_test_zeroext_inf`
Maybe this one os simpler as in unsigned domain.

These previous rules may be ok for unsigned as well:
```
ExprInt('a', 8).zeroExt(32) <u ExprInt(300, 32)
```
is always true.

serpilliere: I corrected a tipo in `simp_test_zeroext_inf` for size2uint

commial: Alright, I agree with you. I was missing that if the cst can fit in the range, it means it can be rewritten as small_cst.signExtend() (and then can be truncated).
Thanks for the rewording.

miasm2/expression/simplifications_common.py
serpilliere: Done

miasm2/expression/simplifications_common.py
```diff
@@ -1075,13 +1109,122 @@ def simp_cond_eq_zero(expr_s, expr):
...
+    if not (arg1.is_op() and arg1.op.startswith("zeroExt")):
+        return expr
+    src = arg1.args[0]
+    assert src.size < arg1.size
+
+    # If cst is zero
+    if arg2.is_int(0):
+        if expr.is_op(TOK_INF_SIGNED):
+            # X.zeroExt() <s 0 => 0
+            return ExprInt(0, 1)
```
commial: Is the size always 1? Shouldn't we use the size of the input expression?

serpilliere: done

PR #909, opened 2018-12-25
Make python version explicit

vbrandl:
On many platforms, the `python` binary links to python 3. This change makes the setup script use python2 explicitly


serpilliere:
Thank you!

PR #835, opened 2018-08-29
Adds Windows support and AppVeyor CI

0vercl0k:
Hey guys,

I've mentioned to you guys during BH that I'd have a look at how much effort it'd be to port miasm to Windows and so here it is.

This PR has all the changes required to have miasm working on Windows 32b/64b with Visual Studio (I've tested VS2017 but it'd be very easy to add different versions if you think it'd be useful in AppVeyor).

Here is the AppVeyor output for the below commit: https://ci.appveyor.com/project/0vercl0k/miasm-nc0e9. As you can see 460 tests out of the 471 of the Linux version are passing just fine.

I've disabled on purpose tests related to `dse_crackme.c` as the `dse_crackme.py` script is expecting an ELF to work properly. Adding support for a Windows PE or/and adding the ELF binary in the git tree would allow to have it tested on Windows too.

There is one last detail that I don't really know how to address though. Basically, if you guys know a way to have `setuptools` move extra files into the build directory (`.lib` files are necessary on Windows in order to have the jitter code being able to link against them). Right now, I manually copy them into the folder. The issue with that is that the content of the folder is moved into the installation directory prior to me moving the files. That's why in the `.appveyor.yml` file I run the install twice in a row; as the `.lib` files will be moved after the first install, the second actually places them into the installation directory. Not a huge deal, but clearly a hack and there must be a cleaner way of doing it :).

Worth mentioning: I've encountered at least two issues with `PyLong_FromLong` / `PyLong_FromUnsignedLong` and I'm unsure why those issues didn't show up on Linux. Maybe you guys have an idea?

One last thing, I've also disabled the `mn_div.py` test with the `llvm` jitter because it yields this on Windows:
```
LLVM ERROR: Program used external function '__umodti3' which could not be resolved!
```

And I'm not too sure who is responsible to expose this routine; again maybe you guys know :)

Happy to answer/discuss any questions you guys have!

Cheers

PS: Thanks to @commial  for helping out with questions and all 👍

PR #914, opened 2019-01-02
Travis: update to llvm 7.0

serpilliere:


serpilliere:
Ok. I applied the patch for building llvm with the hack from @aguinet, it seems to compile!
However, it seems we have errors on travis/appveyor at runtime on specific llvm binaries.

serpilliere:
I tracked down a faulty cause: If I set the `jit_maxline` jittter option to 1 (which gives the jitter the max number of lines to be jitted at once) the jit segfaults on the instruction `RCR` with llvmlite 0.27.

An interesting thing is before this one, the instruction `SHL`, which is very close to `RCR` in term of code generation doesn't make the jit segfault.

serpilliere:
Correction: it segfaults during *execution*, not during *jit*.

serpilliere:
Lol: it was in the logs from the beginning...

serpilliere:
So the bug is exactly what we had on windows, with #835, but on unix now as well.
Time to fix it !

serpilliere:
Ok. I finally got something: the trick is to link llvm lib before generating the library:


```python
    def enable_cache(self):
        "Enable cache of compiled object"
        # GNAAARRRFFF
        self.library_filenames.append('/usr/lib/libLLVM-7.so')
        # Load shared libraries
        for lib_fname in self.library_filenames:
            self.add_shared_library(lib_fname)

        # Activate cache
        self.exec_engine.set_object_cache(self.cache_notify,
                                          self.cache_getbuffer)
```

serpilliere:
Erf.

In fact, I tried to reproduce the bug in a simple file, but currently my example generates only a library, not a binary. So the result is a '.o' and the link is done by my complier, not by llvmlite.
So my example doesn't stand to show llvmlite guys the linkage problem
```python
from llvmlite import ir

import llvmlite.binding as llvm
from ctypes import CFUNCTYPE, c_int


SIZE = 128

# Create some useful types
myint = ir.IntType(SIZE)
fnty = ir.FunctionType(myint, (myint, myint))

# Create an empty module...
module = ir.Module(name=__file__)
# and declare a function named "fpadd" inside it
func = ir.Function(module, fnty, name="main")

# Now implement the function
block = func.append_basic_block(name="entry")
builder = ir.IRBuilder(block)
a, b = func.args
result = builder.urem(a, b, name="res")
builder.ret(result)


llvm.initialize()
llvm.initialize_native_target()
llvm.initialize_native_asmprinter()

target = llvm.Target.from_default_triple()
target_machine = target.create_target_machine()
mod = llvm.parse_assembly(str(module))
mod.verify()
open("%s.o" % "xxx", "wb").write(target_machine.emit_object(mod))
```

serpilliere:
Ok, in fact the code in llvm *already* tryed to link with libllvm. The problem is that the 'loading code' doesn't work for llvm 7 changed their naming convention:
```python
    def load_libraries(self):
        # Get LLVM specific functions
        name = "libLLVM-%d.%d" % (llvm.llvm_version_info[0],
                                  llvm.llvm_version_info[1],
        )
        try:
            # On Windows, no need to add ".dll"
            self.add_shared_library(name)
        except RuntimeError:
            try:
                # On Linux, ".so" is needed
                self.add_shared_library("%s.so" % name)
            except RuntimeError:
                pass

        # Load additional libraries
        for lib_fname in self.library_filenames:
            self.add_shared_library(lib_fname)
```
In llvm 7, the name is something like `libLLVM-7.so`. No more `libLLVM-6.0.so` ...

aguinet:
Bien vu !

Is this a code part of llvmlite?

serpilliere:
Nop, it's in Miasm.
Anyway, it seems to work on Linux, but not on Windows.

serpilliere:
The 'default triple' returns:
```
(Pdb) llvmlite.binding.get_process_triple()
'x86_64-unknown-linux-gnu'
```
Even on Windows. So I use the current process triple instead:
```diff
+        triple = llvm.get_process_triple()
+        target = llvm.Target.from_triple(triple)
```

serpilliere:
Hey, this will allow us to normalize regressions test between windows and linux: no more regession tests exceptions :tada:  (#835)

serpilliere:
Ok : we have to keep the ELF, as llvm MCJIT only support this at runtime. So back to the original problem. In fact llvm is able to link the elf at runtime with dll: it works for the library `JitCore_arch` for example. So the problem was not here.

After some tests, it is evident that the code doesn't find LLVM library:
```
LOAD DYN
X libLLVM-7.0
X libLLVM-7
X libclang
NO lib LOADED!!!
```
None of those libraries are found at runtime. I don't know which library has to be linked at runtime to implement umodi3 for now.

commial:
Another possibility would be to disable the use of "builtins" (named according to https://compiler-rt.llvm.org/). In llvmlite, there are a few trace of "builtin" support, especially in https://github.com/numba/llvmlite/blob/master/llvmlite/llvmpy/passes.py (but it looks like the code is kind of dead), and also in `FunctionAttributes` (which seems related to [this](https://groups.google.com/forum/#!msg/llvm-dev/HaUw3JxO-BE/vfeYJXR1j14J) discussion).

I didn't find a way to reach the "nobuiltins" possibility of llvmlite for now, and adding a `nobuiltin` attribute on the generated function doesn't change anything.
(Note: the compiler-rt page mention a full C version of these so-called builtins)

commial:
By the note, I mean, we may explicitly compile it https://github.com/llvm-mirror/compiler-rt for our needs...

PR #926, opened 2019-01-10
cleanup travis.yml and appveyor.yml

stephengroat:
Include pip requirements.txt files and pin versions where needed
No double venv

@serpilliere this should be step 1 on #499


stephengroat:
final change made, good 2 go

serpilliere:
Great!
Thank you @stephengroat

commial:
Yeah, thank you!
I guess as llvmlite version is explicitely fixed in the requirements, we can rebase the PRs and have tests working (without waiting for #914)

PR #929, opened 2019-01-11
Not shown instruction prefix

nofiv:
Instruction prefix wasn't shown when to_string method was used since instruction_x86(https://github.com/cea-sec/miasm/blob/master/miasm2/arch/x86/arch.py#L452) used just the inherited one from instruction(https://github.com/cea-sec/miasm/blob/master/miasm2/core/cpu.py#L997) which doesn't support these prefixes. It is used perhaps by AsmBlock(https://github.com/cea-sec/miasm/blob/master/miasm2/core/asmblock.py#L147) and it makes the prefixes invisible for it and many others.


serpilliere:
Perfect. Thank you @nofiv !

PR #912, opened 2018-12-29
Fix for undeclared identifier __BIG_ENDIAN and __LITTLE_ENDIAN on macOS

dnutiu:
This is a newer pull requests that resolves conflicts from https://github.com/cea-sec/miasm/pull/438

PR #930, opened 2019-01-11
Macos

stephengroat:
closes #438, closes #875, closes #912

still an issue with `gcc` tests on macos


stephengroat:
needs to run on appveyor to check for windows issues

commial:
Not sure to understand what you're asking.
If you want to try without `gcc`, you just have to remove:
`-o qemu -t python,llvm`
`-o gcc -t qemu,long`

And replace `-o long` by `-o long -t gcc`. `-o` is whitelisting, `-t` is blacklisting.

Regarding the necessity of `gcc` on MacOS X, I'm not sure. If we have at least one of the fast jitter (either CC based or LLVM), it should be OK. Thoughts @serpilliere ?

Note: maybe it could work with Clang or specific options. For instance, on Windows, it is actually using `cl` with others flags. Options can be customized in `JitCore_Gcc.add_block`

XVilka:
- Should be rebased I think
- Travis now supports Windows too, so you might extend with it too.

serpilliere:
@XVilka : note that we already have windows regressions tests run on appveyor.
Do you propose removing appveyor in favor of travis?
Can you argu on this ?

XVilka:
No, I am not insisting. Missed the AppVeyor thing, sorry.

serpilliere:
No problem!

PR #932, opened 2019-01-11
parallel tests

stephengroat:
should improve test speed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,17 @@ addons:
       - llvm-6.0
       - llvm-6.0-dev
       - g++-5
-env: CXX=g++-5 LLVM_CONFIG=llvm-config-6.0
+env:
+  global: CXX=g++-5 LLVM_CONFIG=llvm-config-6.0
+  matrix:
+    - MIASM_TEST_TAG=regression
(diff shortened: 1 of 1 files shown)


commial:
Well, it seems a good idea, but I'm afraid there might be a lot of redundancy. Does travails launch several instances for matrix?

stephengroat:
yup, launches a bunch https://travis-ci.org/cea-sec/miasm/builds/478452513

don't know why travis didn't associate the build with this pr

stephengroat:
@commial is the `tcc` code in `tests/test_all.py` still valid or should the references be removed?

commial:
I don't see any TCC reference in test_all :/

stephengroat:
@commial sorry about that, must have been looking at an old branch. cleaned it up

serpilliere:
Thank @stephengroat !

commial:
I'll take a closer look at how we can better dispatch tests (for now, I think example + regression already map 100% of the testset)

stephengroat:
@commial great! if you could keep in mind still being able to split out the gcc tests, could probably get macos build and test working on #930

PR #946, opened 2019-01-17
Armt: fix flags update in ITT blocks

serpilliere:


a-vincent:
Hi, yes, the flags are re-evaluated at each instruction (and negated depending on the T or E). The best witness to this in the ARMv7 architecture reference manual is paragraph A8.3.1 Pseudocode details of conditional execution: a part of the ITSTATE internal register is basically used as a substitute for the usual 4 bit prefix in A type instructions.

serpilliere:
Ok. Thanks @a-vincent. So I am removing the wip here.

PR #952, opened 2019-01-26
PPC: Explicit flags

serpilliere:


serpilliere:
I have added the support of high level flags in the jitter, since  some architecture can use them directly (here ppc).
The other solution could be to add explicit logical computation for those operators (`<s`, `<u`, ...) but maybe it's less interesting: for example llvm will directly have native translation, resulting in quicker jitted code.

commial:
I agree. Actually, the translation of some of x86 flags could also benefit from this, to avoid having a really complicated expression, but comparison instead. I think at least llvm, but probably gcc too, will be able to better optimize and jit them.

PR #953, opened 2019-01-27
Elf add sym

serpilliere:


miasm2/core/locationdb.py
```diff
@@ -252,7 +252,7 @@ def add_location(self, name=None, offset=None, strict=True):
             # Non-strict mode
             if name_loc_key is not None:
                 known_offset = self.get_offset_location(name_loc_key)
-                if known_offset != offset:
+                if known_offset is not None and known_offset != offset:
```
commial: IMHO, this is not enough: if `known_offset` is actually `None`, the returned `name_loc_key` will not have its offset set, so there is a set offset to do.

miasm2/analysis/binary.py
```diff
@@ -209,6 +209,16 @@ def parse(self, data, vm=None, addr=0, apply_reloc=False, **kwargs):
             # Add known symbols (vm_load_elf already does it)
             fill_loc_db_with_symbols(self._executable, self.loc_db, addr)
 
+            if hasattr(self._executable.sh, 'dynsym'):
```
commial: This should rather be done in `fill_loc_db_with_symbols`

serpilliere: I agree with you: I moved the code

miasm2/core/locationdb.py
serpilliere: Yes! I added the set_offset.

serpilliere: I agree with you: I moved the code

serpilliere: Yes! I added the set_offset.

miasm2/analysis/binary.py
serpilliere: I agree with you: I moved the code

Issue #954, opened 2019-01-29
dead branches of the IR-CFG are not removed


KosBeg:
Hello! I'm new user of miasm, so maybe this is not a issue, but just a question.

```
python full.py binary -m x86_32 -g -s -p -x -y -e -j
```
[binary.zip](https://github.com/cea-sec/miasm/files/2807328/binary.zip)

part of final.dot

![2019-01-29_12-44-02](https://user-images.githubusercontent.com/10928175/51903783-b6910e80-23c5-11e9-8ac2-a52aa7b16300.png)

Because of this non-optimization, we cannot optimize the rest of the code.

Is this a bug or an expected output?

serpilliere:
Hi!

Yes, we saw this behavior a couple of days ago; And yes, we have to fix this in a near future.

Here is the explanation:
First here, we are in `SSA` form. Then you may have applied some modifications (expression propagation, simplification, ...) which resulted in a simplification of `IRDst`: from an `ExprCond` for example, to a constant here.

The trick is that for now, even if the `IRDst` has been updated, Miasm doesn't want to update the graph. Why? Because if we simplify the graph and del some blocks, we will loose some `SSA` assignments, which will loose some sources of `Phi` equations. (which are necessary for the `out-of-ssa` algorithm for example)

But we worked a bit on this and we may have a solution. I will push it, as you can test it and we can decide on integrated it in Miasm core.

Result: the link *to* `loc_key_9` will be removed, so the `Phi` at `loc_1d` will be simplified, to the constant will be propagated, and the simplifier machine will go on the next step!


For the moment, there are errors in the merging of block, so it's not completely working!

Anyway, thanks for the issue @KosBeg  and the example, it's a perfect one :smile:

serpilliere:
Here is a temporary result: the bug is in the last `Phi` which should not be here, and should result in edx propagation, which should give more reductions ...
![ssa_6](https://user-images.githubusercontent.com/9433661/51941178-3c3bab00-2414-11e9-89c9-c661aa0e1851.png)

KosBeg:
> I will push it, as you can test it and we can decide on integrated it in Miasm core.
it would be great!)

> Anyway, thanks for the issue @KosBeg and the example, it's a perfect one 😄
it's one of 34 VM handlers from one crackme. The smallest one, so I have binaries for tests)))

Thank you for answer!

serpilliere:
Hey, it's seems this can be patched!
Here is the final graph.
[...]

PR #955, opened 2019-02-01
Del unused dst in ir simp

serpilliere:
When doing IR simplifications, `IRDst` can be simplified from an `ExprCond` to a simple  `ExprLoc` for example in obfuscated codes. The `ircfg` may then be updated to reflect the lost of an edge.
Until now, this job couldn't be done as some passes (like unssa) takes a sanitized `ircfg` in ssa form, meaning each `phi` source can be tracked back to it's assignments... which is not the case if we simply delete edges during graph simplifications.

This patch add the `Phi` update after edges deletion. This gives more interesting IR simplifications. 
Example here with the following input `ircfg`:
ircfg_0

which gives (after ssa form transformation and simplifications):
unssa_0

This PR resolves the first question of #954 by @KosBeg

Note: regression tests will be added after the factorization and "librarization" of the "simplification code" part present in `full.py`

### DIFF
--- a/miasm2/analysis/data_flow.py
+++ b/miasm2/analysis/data_flow.py
@@ -4,9 +4,12 @@ from collections import namedtuple
 from miasm2.core.graph import DiGraph
 from miasm2.ir.ir import AssignBlock, IRBlock
 from miasm2.expression.expression import ExprLoc, ExprMem, ExprId, ExprInt,\
-    ExprAssign
+    ExprAssign, ExprOp
 from miasm2.expression.simplifications import expr_simp
 from miasm2.core.interval import interval
+from miasm2.expression.expression_helper import possible_values
+from miasm2.analysis.ssa import get_phi_sources_parent_block, \
+    irblock_has_phi
 
 class ReachingDefinitions(dict):
     """
@@ -1212,3 +1215,201 @@ class DiGraphLivenessIRA(DiGraphLiveness):
             var_out = ir_arch_a.get_out_regs(irblock)
             irblock_liveness = self.blocks[node]
             irblock_liveness.infos[-1].var_out = var_out
+
+
+def discard_phi_sources(ircfg, deleted_vars):
+    """
+    Remove phi sources in @ircfg belonging to @deleted_vars set
+    @ircfg: IRCFG instance in ssa form
+    @deleted_vars: unused phi sources
+    """
+    for block in ircfg.blocks.values():
+        if not block.assignblks:
+            continue
+        assignblk = block[0]
+        todo = {}
+        modified = False
+        for dst, src in assignblk.iteritems():
+            if not src.is_op('Phi'):
+                todo[dst] = src
+                continue
+            srcs = set(expr for expr in src.args if expr not in deleted_vars)
+            assert(srcs)
+            if len(srcs) > 1:
+                todo[dst] = srcs
+                continue
+            todo[dst] = srcs.pop()
+            modified = True
+        if not modified:
+            continue
+        assignblks = list(block)
+        assignblk = dict(assignblk)
+        assignblk.update(todo)
+        assignblk = AssignBlock(assignblk, assignblks[0].instr)
+        assignblks[0] = assignblk
+        new_irblock = IRBlock(block.loc_key, assignblks)
+        ircfg.blocks[block.loc_key] = new_irblock
+    return True
+
+
+def get_unreachable_nodes(ircfg, edges_to_del, heads):
+    """
+    Return the unreachable nodes starting from heads and the associated edges to
+    be deleted.
+
+    @ircfg: IRCFG instance
+    @edges_to_del: edges already marked as deleted
+    heads: locations of graph heads
+    """
+    todo = set(heads)
+    visited_nodes = set()
+    new_edges_to_del = set()
+    while todo:
+        node = todo.pop()
+        if node in visited_nodes:
+            continue
+        visited_nodes.add(node)
+        for successor in ircfg.successors(node):
+            if (node, successor) not in edges_to_del:
+                todo.add(successor)
+    all_nodes = set(ircfg.nodes())
+    nodes_to_del = all_nodes.difference(visited_nodes)
+    for node in nodes_to_del:
+        for successor in ircfg.successors(node):
+            if successor not in nodes_to_del:
+                # Frontier: link from a deleted node to a living node
+                new_edges_to_del.add((node, successor))
+    return nodes_to_del, new_edges_to_del
+
+
+def update_phi_with_deleted_edges(ircfg, edges_to_del):
+    """
+    Update phi which have a source present in @edges_to_del
+    @ssa: IRCFG instance in ssa form
+    @edges_to_del: edges to delete
+    """
+
+    modified = False
+    blocks = dict(ircfg.blocks)
+    for loc_src, loc_dst in edges_to_del:
+        block = ircfg.blocks[loc_dst]
+        assert block.assignblks
+        assignblks = list(block)
+        assignblk = assignblks[0]
+        out = {}
+        for dst, phi_sources in assignblk.iteritems():
+            if not phi_sources.is_op('Phi'):
+                out = assignblk
+                break
+            var_to_parents = get_phi_sources_parent_block(
+                ircfg,
+                loc_dst,
+                phi_sources.args
+            )
+            to_keep = set(phi_sources.args)
+            for src in phi_sources.args:
+                parents = var_to_parents[src]
+                if loc_src in parents:
+                    to_keep.discard(src)
+                    modified = True
+            assert to_keep
+            if len(to_keep) == 1:
+                out[dst] = to_keep.pop()
+            else:
+                out[dst] = ExprOp('Phi', *to_keep)
+        assignblk = AssignBlock(out, assignblks[0].instr)
+        assignblks[0] = assignblk
+        new_irblock = IRBlock(loc_dst, assignblks)
+        blocks[block.loc_key] = new_irblock
+
+    for loc_key, block in blocks.iteritems():
+        ircfg.blocks[loc_key] = block
+    return modified
+
+
+def del_unused_edges(ircfg, heads):
+    """
+    Delete non accessible edges in the @ircfg graph.
+    @ircfg: IRCFG instance in ssa form
+    @heads: location of the heads of the graph
+    """
+
+    deleted_vars = set()
+    modified = False
+    edges_to_del_1 = set()
+    for node in ircfg.nodes():
+        successors = set(ircfg.successors(node))
+        block = ircfg.blocks[node]
+        dst = block.dst
+        possible_dsts = set(solution.value for solution in possible_values(dst))
+        if not all(dst.is_loc() for dst in possible_dsts):
+            continue
+        possible_dsts = set(dst.loc_key for dst in possible_dsts)
+        if len(possible_dsts) == len(successors):
+            continue
+        dsts_to_del = successors.difference(possible_dsts)
+        for dst in dsts_to_del:
+            edges_to_del_1.add((node, dst))
+
+    # Remove edges and update phi accordingly
+    # Two cases here:
+    # - edge is directly linked to a phi node
+    # - edge is indirect linked to a phi node
+    nodes_to_del, edges_to_del_2 = get_unreachable_nodes(ircfg, edges_to_del_1, heads)
+    modified |= update_phi_with_deleted_edges(ircfg, edges_to_del_1.union(edges_to_del_2))
+
+    for src, dst in edges_to_del_1.union(edges_to_del_2):
+        ircfg.del_edge(src, dst)
+    for node in nodes_to_del:
+        block = ircfg.blocks[node]
+        ircfg.del_node(node)
+        for assignblock in block:
+            for dst in assignblock:
+                deleted_vars.add(dst)
+
+    if deleted_vars:
+        modified |= discard_phi_sources(ircfg, deleted_vars)
+
+    return modified
+
+
+class DiGraphLivenessSSA(DiGraphLivenessIRA):
+    """
+    DiGraph representing variable liveness is a SSA graph
+    """
+    def __init__(self, ircfg):
+        super(DiGraphLivenessSSA, self).__init__(ircfg)
+
+        self.loc_key_to_phi_parents = {}
+        for irblock in self.blocks.values():
+            if not irblock_has_phi(irblock):
+                continue
+            out = {}
+            for sources in irblock[0].itervalues():
+                var_to_parents = get_phi_sources_parent_block(self, irblock.loc_key, sources.args)
+                for var, var_parents in var_to_parents.iteritems():
+                    out.setdefault(var, set()).update(var_parents)
+            self.loc_key_to_phi_parents[irblock.loc_key] = out
+
+    def back_propagate_to_parent(self, todo, node, parent):
+        parent_block = self.blocks[parent]
+        cur_block = self.blocks[node]
+        irblock = self.ircfg.blocks[node]
+        if cur_block.infos[0].var_in == parent_block.infos[-1].var_out:
+            return
+        var_info = cur_block.infos[0].var_in.union(parent_block.infos[-1].var_out)
+
+        if irblock_has_phi(irblock):
+            # Remove phi special case
+            out = set()
+            phi_sources = self.loc_key_to_phi_parents[irblock.loc_key]
+            for var in var_info:
+                if var not in phi_sources:
+                    out.add(var)
+                    continue
+                if parent in phi_sources[var]:
+                    out.add(var)
+            var_info = out
+
+        parent_block.infos[-1].var_out = var_info
+        todo.add(parent)
--- a/miasm2/analysis/ssa.py
+++ b/miasm2/analysis/ssa.py
@@ -791,7 +791,6 @@ class UnSSADiGraph(object):
                 for src in sources.args:
                     parents = var_to_parents[src]
                     self.new_var_to_srcs_parents.setdefault(new_var, set()).update(parents)
-                    parent_dsts = set((parent, src) for parent in parents)
                     for parent in parents:
                         self.phi_parent_sources.setdefault(dst, set()).add((parent, src))
 
--- a/miasm2/analysis/ssa.py
+++ b/miasm2/analysis/ssa.py
@@ -1,6 +1,5 @@
 from collections import deque
 
-from miasm2.analysis.data_flow import DiGraphLivenessIRA, remove_empty_assignblks
 from miasm2.expression.expression import ExprId, ExprAssign, ExprOp, get_expr_ids
 from miasm2.ir.ir import AssignBlock, IRBlock
 
@@ -1079,47 +1078,3 @@ class UnSSADiGraph(object):
                     out[dst] = src
                 assignblks[i] = AssignBlock(out, assignblk.instr)
             self.ssa.graph.blocks[irblock.loc_key] = IRBlock(irblock.loc_key, assignblks)
-
-
-class DiGraphLivenessSSA(DiGraphLivenessIRA):
-    """
-    DiGraph representing variable liveness is a SSA graph
-    """
-    def __init__(self, ircfg):
-        super(DiGraphLivenessSSA, self).__init__(ircfg)
-
-        self.loc_key_to_phi_parents = {}
-        for irblock in self.blocks.values():
-            if not irblock_has_phi(irblock):
-                continue
-            out = {}
-            for sources in irblock[0].itervalues():
-                var_to_parents = get_phi_sources_parent_block(self, irblock.loc_key, sources.args)
-                for var, var_parents in var_to_parents.iteritems():
-                    out.setdefault(var, set()).update(var_parents)
-            self.loc_key_to_phi_parents[irblock.loc_key] = out
-
-    def back_propagate_to_parent(self, todo, node, parent):
-        parent_block = self.blocks[parent]
-        cur_block = self.blocks[node]
-        irblock = self.ircfg.blocks[node]
-        if cur_block.infos[0].var_in == parent_block.infos[-1].var_out:
-            return
-        var_info = cur_block.infos[0].var_in.union(parent_block.infos[-1].var_out)
-
-        if irblock_has_phi(irblock):
-            # Remove phi special case
-            out = set()
-            phi_sources = self.loc_key_to_phi_parents[irblock.loc_key]
-            for var in var_info:
-                if var not in phi_sources:
-                    out.add(var)
-                    continue
-                if parent in phi_sources[var]:
-                    out.add(var)
-            var_info = out
-
-        parent_block.infos[-1].var_out = var_info
-        todo.add(parent)
-
-
--- a/example/disasm/full.py
+++ b/example/disasm/full.py
@@ -8,9 +8,12 @@ from miasm2.core.interval import interval
 from miasm2.analysis.machine import Machine
 from miasm2.analysis.data_flow import dead_simp, DiGraphDefUse, \
     ReachingDefinitions, merge_blocks, remove_empty_assignblks, \
-    PropagateExpr, replace_stack_vars, load_from_int
+    PropagateExpr, replace_stack_vars, load_from_int, \
+    del_unused_edges
 from miasm2.expression.simplifications import expr_simp
-from miasm2.analysis.ssa import SSADiGraph, UnSSADiGraph, DiGraphLivenessSSA
+from miasm2.analysis.ssa import SSADiGraph
+from miasm2.analysis.outofssa import UnSSADiGraph
+from miasm2.analysis.data_flow import DiGraphLivenessSSA
 from miasm2.ir.ir import AssignBlock, IRBlock
 
 
@@ -395,6 +398,9 @@ if args.propagexpr:
                 if args.verbose > 3:
                     open('tmp_after_%d.dot' % index, 'w').write(ircfg_a.dot())
                 simp_modified |= remove_empty_assignblks(ircfg_a)
+                simp_modified |= del_unused_edges(ircfg_a, heads)
+                simp_modified |= merge_blocks(ircfg_a, heads)
+
                 if args.loadint:
                     simp_modified |= load_from_int(ircfg_a, bs, is_addr_ro_variable)
                 modified |= simp_modified
--- a/miasm2/analysis/outofssa.py
+++ b/miasm2/analysis/outofssa.py
@@ -1,0 +1,411 @@
+from miasm2.expression.expression import ExprId
+from miasm2.ir.ir import IRBlock, AssignBlock
+from miasm2.analysis.ssa import get_phi_sources_parent_block, \
+    irblock_has_phi
+
+
+class Varinfo(object):
+    """Store liveness information for a variable"""
+    __slots__ = ["live_index", "loc_key", "index"]
+
+    def __init__(self, live_index, loc_key, index):
+        self.live_index = live_index
+        self.loc_key = loc_key
+        self.index = index
+
+
+class UnSSADiGraph(object):
+    """
+    Implements unssa algorithm
+    Revisiting Out-of-SSA Translation for Correctness, Code Quality, and
+    Efficiency
+    """
+
+    def __init__(self, ssa, head, cfg_liveness):
+        self.cfg_liveness = cfg_liveness
+        self.ssa = ssa
+        self.head = head
+
+        # Set of created variables
+        self.copy_vars = set()
+        # Virtual parallel copies
+
+        # On loc_key's Phi node dst -> set((parent, src))
+        self.phi_parent_sources = {}
+        # On loc_key's Phi node, loc_key -> set(Phi dsts)
+        self.phi_destinations = {}
+        # Phi's dst -> new var
+        self.phi_new_var = {}
+        # For a new_var representing dst:
+        # new_var -> set(parents of Phi's src in dst = Phi(src,...))
+        self.new_var_to_srcs_parents = {}
+        # new_var -> set(variables to be coalesced with, named "merge_set")
+        self.merge_state = {}
+
+        # Launch the algorithm in several steps
+        self.isolate_phi_nodes_block()
+        self.init_phis_merge_state()
+        self.order_ssa_var_dom()
+        self.aggressive_coalesce_block()
+        self.insert_parallel_copy()
+        self.replace_merge_sets()
+        self.remove_assign_eq()
+
+    def insert_parallel_copy(self):
+        """
+        Naive Out-of-SSA from CSSA (without coalescing for now)
+        - Replace Phi
+        - Create room for parallel copies in Phi's parents
+        """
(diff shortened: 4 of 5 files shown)


KosBeg:
when I use `miasm2` and `full.py` from this PR with this binary

[binary_1D.zip](https://github.com/cea-sec/miasm/files/2820851/binary_1D.zip)

```
PS C:\!miasm_test> python full.py binary_1D -m x86_32 -g -s -s -p -x -y -e -j
INFO : Load binary
INFO : ok
INFO : import machine...
INFO : ok
INFO : func ok 0000000000000000 (0)
INFO : generate graph file
INFO : generate intervals
[0x0 0x244]
INFO : total lines 0
INFO : generating IR and IR analysis
INFO : generating IR... 0
INFO : Print blocks (without analyse)
INFO : Gen Graph... 0
INFO : Print blocks (with analyse)
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
  File "full.py", line 392, in <module>
    simp_modified |= dead_simp(ir_arch_a, ircfg_a)
  File "C:\Python27_x64\lib\site-packages\miasm2\analysis\data_flow.py", line 259, in dead_simp
    reaching_defs = ReachingDefinitions(ircfg)
  File "C:\Python27_x64\lib\site-packages\miasm2\analysis\data_flow.py", line 42, in __init__
    self.compute()
  File "C:\Python27_x64\lib\site-packages\miasm2\analysis\data_flow.py", line 57, in compute
    modified |= self.process_block(block)
  File "C:\Python27_x64\lib\site-packages\miasm2\analysis\data_flow.py", line 66, in process_block
    pred = self.ircfg.blocks[pred_lbl]
KeyError: <LocKey 17>
```

and log for `miasm2` and `full.py` from `master`

```
PS C:\!miasm_test> python full_orig.py binary_1D -m x86_32 -g -s -s -p -x -y -e -j
INFO : Load binary
INFO : ok
INFO : import machine...
INFO : ok
INFO : func ok 0000000000000000 (0)
INFO : generate graph file
INFO : generate intervals
[0x0 0x244]
INFO : total lines 0
INFO : generating IR and IR analysis
INFO : generating IR... 0
INFO : Print blocks (without analyse)
INFO : Gen Graph... 0
INFO : Print blocks (with analyse)
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
INFO : dead simp...
INFO : ok...
******************************************************************************** Remove phi
```

ofc `miasm2` from `master` don't do this optimization, but PR looks like a regression(((

serpilliere:
Hey,
Thanks for the test!
I fixed (I forgot an edge deletion)

Can you test it again? Here, it seems to be good with `binary` and `binary_1C`
Thanks again for the feedback @KosBeg !
+
serpi

KosBeg:
it seems that everything is ok 😃

but I can't test this file
for me miasm work more than 2 hours and still work...

[binary_MAX.zip](https://github.com/cea-sec/miasm/files/2823372/binary_MAX.zip)

it's biggest VM handler, just `1701 bytes`, but miasm can't simplify it 🙁

serpilliere:
Ok I think I have found the problem:
As I said, Miasm for the moment doesn't propagate constant through memory storage.
If you look at your binary and to the propagation trick on ly for the first 512 bytes (force RET for last instruction) you will obtain:

final

As you can see, you have sometime:
```
[XXX] = CST
TOTO = [XXX]
```

For now, Miasm doesn't know he can propagate `CST` into `TOTO` directly.
This is a first point.

For the "size" of the graph, if you do the following code:
```
A = A + A
A = A + A
A = A + A
A = A + A
A = A + A
A = A + A
A = A + A
A = A + A

```
The size of the final expression (developed) will be a tree of 256 nodes :smile: 
Four more additions and it's 1024 !
This one is in fact a bad example as Miasm known that A+A = 2*A and so on
But if it's a XOR then ADD the OR, there are no simplifications... except if arguments are constants, which as you see in the first point can be missed by Miasm for now!

serpilliere:
Ok, maybe we can try to do really propagation constant through memory in very simple cases.
I will have a try, and keep you in touch!

KosBeg:
thank you very much!
I think that this feature will be useful for everyone)


there is another question about this PR:
I noticed that if the last instr of the out-node(last node) not a jcc or ret(or any other instr that change IRDst) - miasm falls.
eg we have
```
nop ; 90
ret ; c3
```
it's will be ok for this pr and for `master` branch
but if we remove ret, eg just
```
nop ; 90
```

error for `full.py` from this pr
```
  File "full.py", line 399, in <module>
    simp_modified |= del_unused_edges(ircfg_a, heads)
  File "C:\Python27_x64\lib\site-packages\miasm2\analysis\data_flow.py", line 1391, in del_unused_edges
    block = ircfg.blocks[node]
KeyError: <LocKey 2>
```

and for `full.py` from `master`
```
  File ".\full_master.py", line 420, in <module>
    cfg_liveness = DiGraphLivenessSSA(ircfg_a)
  File "C:\Python27_x64\lib\site-packages\miasm2\analysis\ssa.py", line 1035, in __init__
    super(DiGraphLivenessSSA, self).__init__(ircfg)
  File "C:\Python27_x64\lib\site-packages\miasm2\analysis\data_flow.py", line 1100, in __init__
    irblock = ircfg.blocks[node]
KeyError: <LocKey 2>
```

serpilliere:
Ok I think I have something:
In fact Miasm can handle the case without problem: this is a question of order of simplification passes.
In the current case, Miasm applies simplification cases one after other.
The problem here is that the `propagate_expr` propagate constant, and expressions.
As we have cases near the `A+A, A+A...` trick, if we apply this one, this will give very big expressions.

So we have to first propagate *constant* in a first time (warning: this simple pass does not exist: it's the same as propag_expr for now :cry:) then apply expressions reduction, then remove dummy jumps (resolved `IRDst` from `ExprCond` to fix destination, the `binary_MAX` is solved in 7 seconds :smile: 

We obtain the following graph (I keep the stack assignment to see all side effects):
final

This is done using a quick and dirty propagation through memory. I will push this soon

Note: It seem we have a pattern here: more over, some memory variables are written multiple times *without* being read between writes. Miasm doesn't remove this for now, but maybe you can write a simple script (for your case) to delete them

 
This might result in a *really* simple final graph  !

commial:
As a quick remark, we could also apply a limit in the propagate pass on the resulting expression depth.
Something like if `resulting_expression.graph.depth > LIMIT`, do not propagate. This could, at least, avoid having really long computation time on obfuscated code (and let the user add simplification to further reduce the result).

serpilliere:
Yes, I did this at first, but maybe the priority of propagation is cleaner: we may reduce leaves of a tree before moving big branches.

commial:
I think both of them are of interest: the const one for efficiency, the depth limit one to avoid future behavior like this in other cases

serpilliere:
Yes: I started rewriting a constant propagation only simplification, and an expression propgation using the optional limit.
But we need an object above this to select/change the applied simplification passes: this will allow to factor the code in `full.py`, `graph_ir.py`, ...

serpilliere:
Anyway, if you are ok, this PR can be done independently of the modification we are doing in the other passes I think.

serpilliere:
Hi @KosBeg !
I have done a branch (https://github.com/serpilliere/miasm/tree/obj_simp_pass_manager) which includes the #955 and an alpha candidate for a PR of Miasm (which is not submitted yep).

It allow you to get the simplified IR of every binaries you have uploaded here (`binary`, `binary_1D` and  `binary_MAX`). The biggest one takes around 7 sec to compute.
```shell
time python2   ~/projet/miasm/example/disasm/full.py binary_MAX 0x0 -m x86_32 -g -s -p -x -y -e -j
```

It gives the graph I uploaded some days ago.
Feel free to test it !

KosBeg:
tested with all vm handlers. for handlers it work as expected,
but 😄 if I try to deobfuscate movfuscator(yeah, [movfuscator](https://github.com/xoreaxeaxeax/movfuscator), probably crazy idea 😅) - it give some strange results

no_name

marked with green and blue border - not propagated data, seem like a regression.

marked with gray border - different result for miasm from master and from this branch. It can be a regression, or on the contrary - it's a progress.

unfortunately I could not built the movfuscator from git, but I have 1 movfuscated binary, if necessary I can upload it. (stripped file, without debug info and source code)

serpilliere:
Humm, maybe the binary could be interesting here :smile: if you don't mind sharing it!

KosBeg:
yeah, no problem)))

[industries.zip](https://github.com/cea-sec/miasm/files/2845871/industries.zip)

it's from this ctf - https://ctftime.org/event/742

serpilliere:
Ok. The trick is that you used `-e` to force loading memory access based on integers from the binary.
This can't be assumed here as globals are in `.data` section, which is RW, and the binary *does* some `.data` modifications.
So the `-e` option can't be used here: the result is different because it depends on which passes are run first, as simplification passes are not safe here.

serpilliere:
Anyway, thank you for the help @KosBeg !
We will try to merge more or less everything here to make simplifications passes usefull.

KosBeg:
`obj_simp_pass_manager` branch vs `master` branch
1.5x speed-down on this file
[BabyCrackme.zip](https://github.com/cea-sec/miasm/files/2846172/BabyCrackme.zip)

serpilliere:
Ok: for the BabyCrackme, I spotted an error in `popfw` which result in a 16 bit slice of a 32 bit pop (which is bad: the documentation says popfw does directly a 16 bit access on the stack)

Result: Miasm refuses to reduce `(@32[ESP + XXX])[:16]` to `@16[ESP + XXX]` (as the 32 bit lookup can hit the end of a page and not the 16 bit lookup)
The result is the following (starting at 171E)
final

I fixed the `popfw`, so the lookup is done in 16 bit mode. The result (for the same adress) is:
final_fix


This unlock more simplifications :smiley:
After a little analysis, we can a missing pattern on dual memory write (without read):
```python
...
[ESP + X] = Y
[ESP + X] = Y
...
```

But maybe it's up to the user: the simplification may not be true for example if the destination address targets an memory mapped hardware..
Anyway, this is way does (a bit) the `-y` option, and, hey, the result seems to be very interesting :smile: : 
final_end
We can see this result in a RET and a big equation in EAX, which seems to be a check of a string with 0x1337 constant and so on:
```python
src = ExprCompose(ExprOp('&', ExprOp('^', ExprOp('zeroExt_8', ExprCond(ExprOp('<=u', ExprOp('+', ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x2, 32)), 32), 8, 16), ExprInt(0xD0, 8)), ExprInt(0x9, 8)), ExprInt(0x0, 1), ExprInt(0x1, 1))), ExprInt(0xFF, 8)), ExprOp('^', ExprOp('zeroExt_8', ExprOp('<s', ExprOp('+', ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x2, 32)), 32), 8, 16), ExprInt(0xD0, 8)), ExprInt(0x0, 8))), ExprInt(0xFF, 8)), ExprOp('^', ExprOp('zeroExt_8', ExprOp('CC_EQ', ExprOp('^', ExprOp('FLAG_EQ_CMP', ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x10, 32)), 16), ExprInt(0x676E, 16)), ExprInt(0x1, 1)))), ExprInt(0xFF, 8)), ExprOp('zeroExt_8', ExprOp('==', ExprMem(ExprId('EDI', 32), 8), ExprInt(0x6C, 8))), ExprOp('zeroExt_8', ExprOp('==', ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x1, 32)), 8), ExprInt(0x69, 8))), ExprOp('zeroExt_8', ExprOp('==', ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x2, 32)), 8), ExprInt(0x6E, 8))), ExprOp('zeroExt_8', ExprOp('==', ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x6, 32)), 8), ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x12, 32)), 8))), ExprOp('zeroExt_8', ExprOp('==', ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0xF, 32)), 8), ExprInt(0x69, 8))), ExprOp('zeroExt_8', ExprOp('==', ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x18, 32)), 8), ExprInt(0x0, 8))), ExprOp('zeroExt_8', ExprOp('==', ExprOp('+', ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x17, 32)), 8), ExprOp('-', ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x3, 32)), 8))), ExprInt(0x2, 8))), ExprOp('zeroExt_8', ExprOp('==', ExprOp('+', ExprOp('-', ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x7, 32)), 8)), ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x7, 32)), 32), 8, 16)), ExprInt(0x2, 8))), ExprOp('zeroExt_8', ExprOp('==', ExprOp('<<', ExprOp('+', ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x2, 32)), 32), 8, 16), ExprInt(0xD0, 8)), ExprInt(0x2, 8)), ExprInt(0xC, 8))), ExprOp('zeroExt_8', ExprOp('==', ExprOp('^', ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x2, 32)), 32), 16, 32), ExprInt(0x1337, 16)), ExprInt(0x6156, 16))), ExprOp('zeroExt_8', ExprOp('==', ExprOp('^', ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0xB, 32)), 32), 16, 24), ExprInt(0x11, 8)), ExprInt(0x7C, 8))), ExprOp('zeroExt_8', ExprOp('==', ExprOp('zeroExt_32', ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x12, 32)), 8)), ExprCompose(ExprInt(0x5F, 8), ExprSlice(ExprOp('+', ExprOp('-', ExprOp('zeroExt_32', ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x12, 32)), 8))), ExprOp('zeroExt_32', ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x6, 32)), 8))), 8, 32)))), ExprOp('zeroExt_8', ExprOp('==', ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x7, 32)), 32), 8, 16), ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0xB, 32)), 8))), ExprOp('zeroExt_8', ExprOp('==', ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x7, 32)), 32), 8, 16), ExprInt(0x72, 8))), ExprOp('zeroExt_8', ExprOp('==', ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x7, 32)), 32), 16, 32), ExprInt(0x6730, 16))), ExprOp('zeroExt_8', ExprOp('==', ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0xB, 32)), 32), 24, 32), ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0xB, 32)), 32), 16, 24))), ExprOp('zeroExt_8', ExprOp('==', ExprInt(0x41, 8), ExprOp('+', ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0xB, 32)), 32), 8, 16), ExprInt(0x1, 8)))), ExprOp('zeroExt_8', ExprOp('==', ExprInt(0x6766717A, 32), ExprOp('^', ExprCompose(ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x13, 32)), 32), 24, 32), ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x13, 32)), 32), 16, 24), ExprSlice(ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x13, 32)), 32), 8, 16), ExprMem(ExprOp('+', ExprId('EDI', 32), ExprInt(0x13, 32)), 8)), ExprInt(0x14131211, 32)))), ExprInt(0x1, 8)), ExprInt(0x0, 24))
>>> print src
{(zeroExt_8(((@32[EDI + 0x2][8:16] + 0xD0) <=u 0x9)?(0x0,0x1)) ^ 0xFF) & (zeroExt_8((@32[EDI + 0x2][8:16] + 0xD0) <s 0x0) ^ 0xFF) & (zeroExt_8(CC_EQ(FLAG_EQ_CMP(@16[EDI + 0x10], 0x676E) ^ 0x1)) ^ 0xFF) & zeroExt_8(@8[EDI] == 0x6C) & zeroExt_8(@8[EDI + 0x1] == 0x69) & zeroExt_8(@8[EDI + 0x2] == 0x6E) & zeroExt_8(@8[EDI + 0x6] == @8[EDI + 0x12]) & zeroExt_8(@8[EDI + 0xF] == 0x69) & zeroExt_8(@8[EDI + 0x18] == 0x0) & zeroExt_8((@8[EDI + 0x17] + -@8[EDI + 0x3]) == 0x2) & zeroExt_8((-@8[EDI + 0x7] + @32[EDI + 0x7][8:16]) == 0x2) & zeroExt_8(((@32[EDI + 0x2][8:16] + 0xD0) << 0x2) == 0xC) & zeroExt_8((@32[EDI + 0x2][16:32] ^ 0x1337) == 0x6156) & zeroExt_8((@32[EDI + 0xB][16:24] ^ 0x11) == 0x7C) & zeroExt_8(zeroExt_32(@8[EDI + 0x12]) == {0x5F 0 8, (-zeroExt_32(@8[EDI + 0x12]) + zeroExt_32(@8[EDI + 0x6]))[8:32] 8 32}) & zeroExt_8(@32[EDI + 0x7][8:16] == @8[EDI + 0xB]) & zeroExt_8(@32[EDI + 0x7][8:16] == 0x72) & zeroExt_8(@32[EDI + 0x7][16:32] == 0x6730) & zeroExt_8(@32[EDI + 0xB][24:32] == @32[EDI + 0xB][16:24]) & zeroExt_8(0x41 == (@32[EDI + 0xB][8:16] + 0x1)) & zeroExt_8(0x6766717A == ({@32[EDI + 0x13][24:32] 0 8, @32[EDI + 0x13][16:24] 8 16, @32[EDI + 0x13][8:16] 16 24, @8[EDI + 0x13] 24 32} ^ 0x14131211)) & 0x1 0 8, 0x0 8 32}
```

It's a AND between multiple condition:
```python
>>> print src.args[0]
(zeroExt_8(((@32[EDI + 0x2][8:16] + 0xD0) <=u 0x9)?(0x0,0x1)) ^ 0xFF) & (zeroExt_8((@32[EDI + 0x2][8:16] + 0xD0) <s 0x0) ^ 0xFF) & (zeroExt_8(CC_EQ(FLAG_EQ_CMP(@16[EDI + 0x10], 0x676E) ^ 0x1)) ^ 0xFF) & zeroExt_8(@8[EDI] == 0x6C) & zeroExt_8(@8[EDI + 0x1] == 0x69) & zeroExt_8(@8[EDI + 0x2] == 0x6E) & zeroExt_8(@8[EDI + 0x6] == @8[EDI + 0x12]) & zeroExt_8(@8[EDI + 0xF] == 0x69) & zeroExt_8(@8[EDI + 0x18] == 0x0) & zeroExt_8((@8[EDI + 0x17] + -@8[EDI + 0x3]) == 0x2) & zeroExt_8((-@8[EDI + 0x7] + @32[EDI + 0x7][8:16]) == 0x2) & zeroExt_8(((@32[EDI + 0x2][8:16] + 0xD0) << 0x2) == 0xC) & zeroExt_8((@32[EDI + 0x2][16:32] ^ 0x1337) == 0x6156) & zeroExt_8((@32[EDI + 0xB][16:24] ^ 0x11) == 0x7C) & zeroExt_8(zeroExt_32(@8[EDI + 0x12]) == {0x5F 0 8, (-zeroExt_32(@8[EDI + 0x12]) + zeroExt_32(@8[EDI + 0x6]))[8:32] 8 32}) & zeroExt_8(@32[EDI + 0x7][8:16] == @8[EDI + 0xB]) & zeroExt_8(@32[EDI + 0x7][8:16] == 0x72) & zeroExt_8(@32[EDI + 0x7][16:32] == 0x6730) & zeroExt_8(@32[EDI + 0xB][24:32] == @32[EDI + 0xB][16:24]) & zeroExt_8(0x41 == (@32[EDI + 0xB][8:16] + 0x1)) & zeroExt_8(0x6766717A == ({@32[EDI + 0x13][24:32] 0 8, @32[EDI + 0x13][16:24] 8 16, @32[EDI + 0x13][8:16] 16 24, @8[EDI + 0x13] 24 32} ^ 0x14131211)) & 0x1
```

conditions:
```
conds = src.args[0]
>>> for arg in conds.args: print arg
...
zeroExt_8(((@32[EDI + 0x2][8:16] + 0xD0) <=u 0x9)?(0x0,0x1)) ^ 0xFF
zeroExt_8((@32[EDI + 0x2][8:16] + 0xD0) <s 0x0) ^ 0xFF
zeroExt_8(CC_EQ(FLAG_EQ_CMP(@16[EDI + 0x10], 0x676E) ^ 0x1)) ^ 0xFF
zeroExt_8(@8[EDI] == 0x6C)
zeroExt_8(@8[EDI + 0x1] == 0x69)
zeroExt_8(@8[EDI + 0x2] == 0x6E)
zeroExt_8(@8[EDI + 0x6] == @8[EDI + 0x12])
zeroExt_8(@8[EDI + 0xF] == 0x69)
zeroExt_8(@8[EDI + 0x18] == 0x0)
zeroExt_8((@8[EDI + 0x17] + -@8[EDI + 0x3]) == 0x2)
zeroExt_8((-@8[EDI + 0x7] + @32[EDI + 0x7][8:16]) == 0x2)
zeroExt_8(((@32[EDI + 0x2][8:16] + 0xD0) << 0x2) == 0xC)
zeroExt_8((@32[EDI + 0x2][16:32] ^ 0x1337) == 0x6156)
zeroExt_8((@32[EDI + 0xB][16:24] ^ 0x11) == 0x7C)
zeroExt_8(zeroExt_32(@8[EDI + 0x12]) == {0x5F 0 8, (-zeroExt_32(@8[EDI + 0x12]) + zeroExt_32(@8[EDI + 0x6]))[8:32] 8 32})                                                                                                                     
zeroExt_8(@32[EDI + 0x7][8:16] == @8[EDI + 0xB])
zeroExt_8(@32[EDI + 0x7][8:16] == 0x72)
zeroExt_8(@32[EDI + 0x7][16:32] == 0x6730)
zeroExt_8(@32[EDI + 0xB][24:32] == @32[EDI + 0xB][16:24])
zeroExt_8(0x41 == (@32[EDI + 0xB][8:16] + 0x1))
zeroExt_8(0x6766717A == ({@32[EDI + 0x13][24:32] 0 8, @32[EDI + 0x13][16:24] 8 16, @32[EDI + 0x13][8:16] 16 24, @8[EDI + 0x13] 24 32} ^ 0x14131211))                                                                                          
0x1
```
Maybe there is an error, as I don't see the test of the character 5.

commial:
Just say'in, but isn't it the time to call z3? :)

KosBeg:
> and, hey, the result seems to be very interesting 😄
> Maybe there is an error, as I don't see the test of the character 5
no, there is no error, it's ok)
this code checks the flag-string

this one from ctf.ma - 
[BabyCrackme.zip](https://github.com/cea-sec/miasm/files/2846523/BabyCrackme.zip) - original binary)


> Just say'in, but isn't it the time to call z3? :)
yea, I solved it this way using z3 :D

serpilliere:
> Just say'in, but isn't it the time to call z3? :)

@commial : erf: #965 
There was a tipo on the signed operators name in z3.
SLT does not exist: it's the native python "<" or "<="

serpilliere:
> no, there is no error, it's ok)

I admit I didn't checked back the original code. Shame on me :confused:

KosBeg:
> Maybe there is an error, as I don't see the test of the character 5.

> zeroExt_8((@32[EDI + 0x2][16:32] ^ 0x1337) == 0x6156)

`@32[EDI + 0x2][16:32]` is `@8[edi + 4] | (@8[edi + 5] << 8)` or `(@8[edi + 4] << 8) | @8[edi + 5]`
maybe add optimization like this?
eg we have:
1) `@32[EDI + 0x7][0:8]` -> `@8[EDI + 0x7]`
2) `@32[EDI + 0x7][8:16]` -> `@8[EDI + 0x8]`
and so on...

PS: I'll try to find more binaries with interesting obfuscation for tests 😄

serpilliere:
Yes: the trick is that this one is *endianess* dependent !

KosBeg:
try to deobfuscate this one - https://www.lhaihai.wang/post/2018-10-7-inctf-2018/#reverse-dragon-slayer :D

[dragonslayer.zip](https://github.com/cea-sec/miasm/files/2846682/dragonslayer.zip)

serpilliere:
I see what you mean @KosBeg !
It's very long only to *disassemble* the binary!

miasm2/analysis/ssa.py
```diff
@@ -1,6 +1,8 @@
 from collections import deque
 
-from miasm2.analysis.data_flow import DiGraphLivenessIRA, remove_empty_assignblks
+from miasm2.analysis.data_flow import DiGraphLivenessIRA, \
```
commial: IMHO, it would be better to have SSA related function in the `ssa.py` file, such as `get_phi_sources_parent_block` and `irblock_has_phi`.
To avoid circular dependency, OutOfSSA can be moved to a separated, dedicated file

miasm2/analysis/data_flow.py
```diff
@@ -1212,3 +1223,203 @@ def init_var_info(self, ir_arch_a):
...
+    """
+
+    deleted_vars = set()
+    modified = False
+    edges_to_del_1 = set()
+    for node in ircfg.nodes():
+        successors = set(ircfg.successors(node))
+        block = ircfg.blocks[node]
+        dst = block.dst
+        possible_dsts = set([solution.value for solution in possible_values(dst)])
```
commial:
```suggestion
        possible_dsts = set(solution.value for solution in possible_values(dst))
```

miasm2/analysis/data_flow.py
```diff
@@ -1212,3 +1223,203 @@ def init_var_info(self, ir_arch_a):
...
+        if not all(dst.is_loc() for dst in possible_dsts):
+            continue
+        possible_dsts = set([dst.loc_key for dst in possible_dsts])
```
commial:
```suggestion
        possible_dsts = set(dst.loc_key for dst in possible_dsts)
```

commial:
Despite the few comments, do you have (@serpilliere or @KosBeg) remaining issues or tests to perform for this PR?

miasm2/analysis/ssa.py
serpilliere: In fact it depends where you put your "SSA related stuff".
It seems you do the choice here to say that `OutOfSSA` is not linked to `SSA` to avoid this particular circular dependency.
By the way, the `del_unused_edge` depends on `update_phi_with_deleted_edges` which depends on `get_phi_sources_parent_block`...
In fact this may result in the move of a lot of function in the SSA file.
In the current state, SSA contains the bases classes.

miasm2/analysis/data_flow.py
serpilliere: done

miasm2/analysis/data_flow.py
serpilliere: done

serpilliere:
Nop for me!

miasm2/analysis/ssa.py
commial: Yes, and indeed, I also see phi functions in SSA. The idea behind is that SSA could be a kind of self contained file with ssa related stuff, and elements which need ssa will import from this file their necessary stuff. For instance, if I'm looking for the function saying if a block contains a phi or not, I'll probably first have a look at the SSA file. It also avoid putting too many things in data_flow, keeping data flow related stuff.

KosBeg:
I also have no questions.
and on all my tests it worked as expected.

miasm2/analysis/ssa.py
serpilliere: The fact is that today, the SSA is only a *transformation* from one ircfg to another. So the look up argument is somehow not a good one for me.
More over, we will have *a lot* of simplificaiton passes which will operate on the ircfg in SSA form: maybe it won't be a good idea to them in the SSA file neither, I think we agree on this.
As all this is due to the out of ssa, maybe I can set back the SSA with it's minimalist form, let the basic function in it, and put outofssa in the file `outofssa.py` (or `unssa.py`?) which will depend on ssa.
We will have:
- ssa stand alone (class + basic functions)
- outofssa: depends on ssa 
- data_flow: classic dependencies

Does it sound good to you @commial ?

serpilliere:
Thank you for the feedback @KosBeg

miasm2/analysis/ssa.py
commial: Yep, sounds good

serpilliere: To me a draw back will the the `DiGraphLivenessSSA` can't be placed in SSA. The solutions are in `unssa.py` or in `data_flow.py`, as the function depends on `ssa.py` AND on `data_flow.py` :cry:

serpilliere: Ok, If we put `DiGraphLivenessSSA` in data_flow and remove the `remove_empty_block` in unssa, unssa only depends on SSA.
so:
- ssa self content
- unssa depends on ssa
- data_flow depends on both

Sounds good to me: data_flow does high level stuff, so it's the user of those "low" modules.

serpilliere:
Here is the version with `outofssa.py` fix.
I tested a rebased version of `https://github.com/serpilliere/miasm/tree/obj_simp_pass_manager` it seems ok for the binaries above.
I will make PRs to progressively push the remaining fixes to process those binaries with the master branch!

miasm2/analysis/ssa.py
commial: Perfect, thanks for the refactoring

commial:
Thanks @KosBeg and @serpilliere for your extended tests

PR #963, opened 2019-02-08
Added function find_path_from_src

nofiv:
The function find_path_from_src does the same as function find_path, but it searches the paths from src to dst, not vice versa like find_path, which might be more efficient in multiple cases.


commial:
Hi!
First, thanks for the PR. It would be nice to add, in the docstring, what the function expects in args (for instance, what is `cycle_count`), and what the output is (list of path, src to dst, from what I understand).

Otherwise, it looks good to me

commial:
Thanks for your contribution!

PR #964, opened 2019-02-08
API: replace shift_offset by base_address

serpilliere:
WARNING: base address is the negative of shift_offset


serpilliere:
Thank you @a-vincent !

PR #970, opened 2019-02-12
ObjC: ctxt is a dictionnary with str keys

commial:


commial:
As discussed offline, this patch is wrong. We believe that:
* Coming from C, we have to identify a type by its name, without any size expression
* Coming from Expr, we want to be able to type any expression, from ExprId to complex ExprMem. So the size is relevant, and str identification is not

Additionally, C <-> Expr is not a bijection: there are ambiguity coming from Expr, especially for Union.

PR #972, opened 2019-02-13
Stop disassembling at INT

nofiv:
When there are INT 3 instructions perhaps behind ExitProcess call and it's in the last basic block of a function(in memory), dis_multiblock(https://github.com/cea-sec/miasm/blob/master/miasm2/core/asmblock.py#L1554) disassembles also the following function if there's no other padding

### DIFF
--- a/miasm2/arch/x86/arch.py
+++ b/miasm2/arch/x86/arch.py
@@ -500,7 +500,7 @@ class instruction_x86(instruction):
         if self.name.startswith('LOOP'):
             return True
         if self.name.startswith('INT'):
-            return True
+            return False
         if self.name.startswith('SYS'):
             return True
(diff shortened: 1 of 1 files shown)


serpilliere:
For such cases, we may implement in the future a basic database which links known functions to some informations, for example (a bit like` IDA`):
- its prototype (the number of parameters, types, ...)
- if the function returns or not (for example `exitprocess`/`exitthread`, ...)
- its ABI (cdecl, stdcall, ...)

Thus, the disassembler may be lead by those information.
For instance, [retdec](https://github.com/avast-tl/retdec) has these precious information.
Are you ok with this?

nofiv:
I agree with you, it would be a better solution in general

PR #973, opened 2019-02-13
Fix for bin_stream_ida after change in shift_offset API

su-vikas:
API change from shift_offset to base_address,  broke bin_stream_ida.py. Fixing the issue in this PR.


serpilliere:
Hi @su-vikas,
Correct: I totally forgot to fix `bin_stream_ida`. Thank you!
But I think there is a bug in your PR: the `base_address` is the opposite of `shift`
So every signs involving `base_address` should be switched
Are you ok with this?

su-vikas:
Woops. Didn't realize the sign is to be changed as well. Updated the PR.

serpilliere:
Thank you @su-vikas !

PR #987, opened 2019-02-27
Added implementation for aarch64 NEGS instruction

su-vikas:
I have added implementation for negs instruction. As per ARM ARM, NEGS is an alias of SUBS instruction. 

I have copied the CMP instruction, as CMP is also an alias of SUBS. Let me know if I got it right.


su-vikas:
For your testing purposes, instruction is "FF 03 0A 6B", IDA shows it as `CMP WZR, W10`, probably aliasing it. The encoding looks to be of NEGS.

commial:
> I have copied the CMP instruction, as CMP is also an alias of SUBS

Does CMP also update the destination?

su-vikas:
No, CMP discards the results and just updates the condition flags.

commial:
So it isn't an alias of SUBS, right? Or maybe I misunderstood your sentence

su-vikas:
As per ARM documentation, NEGS and CMP are alias of SUBS - but NEGS and CMP discard the result.  http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0802b/NEGS_SUBS_addsub_shift.html

If you ask me, I don't think alias is the right word to use. But documentation states so, thus I use it as well.

PR #998, opened 2019-03-10
Extended tests - reintroduce ls tests with extern repository

commial:


commial:
Thanks to #1024, the test now run further in Python 3. It still fails due to #1016

commial:
As #1016 has been fixed, it is no more WIP

PR #1002, opened 2019-03-13
Fix typo in docstring

Nutigoodpipes:


miasm/expression/expression.py
```diff
@@ -1270,7 +1270,7 @@ def is_slice(self, start=None, stop=None):
 class ExprCompose(Expr):
 
     """
-    Compose is like a hambuger. It concatenate Expressions
```
p-l-: You could have fixed the second typo (concatenate**s**) ;-)

PR #1003, opened 2019-03-13
Add CET instructions

WilliamBruneau:


miasm/arch/x86/arch.py
```diff
@@ -4620,6 +4620,14 @@ class ia32_aaa(mn_x86):
 
 addop("emms", [bs8(0x0f), bs8(0x77)])
 
+addop("incssp", [pref_f3, bs8(0x0f), bs8(0xae)] + rmmod(d5))
+addop("rdssp", [pref_f3, bs8(0x0f), bs8(0x1e)] + rmmod(d1))
```
serpilliere: It seems rdssp has a fixed mod=11
So it seems it only accepts registers.
Are you ok with this?

PR #1005, opened 2019-03-13
Arm: add parallel instructions

serpilliere:


miasm/arch/arm/sem.py
```diff
@@ -1285,6 +1285,97 @@ def uadd8(ir, instr, a, b, c):
...
+    return e, []
+
+
+def sadd8(ir, instr, a, b, c):
+    e = []
+    sums = []
+    ges = []
+    for i in range(0, 32, 8):
+        sums.append(b[i:i+8] + c[i:i+8])
+        ges.append(ExprOp("FLAG_SIGN_SUB", ExprInt(0, 9), b[i:i+8].signExtend(9) + c[i:i+8].signExtend(9)))
```
commial: Double checking: is using `_sub` on purpose? (because it seems to be an addition)

serpilliere:
yes it is.
But I need someone to read again the doc by himself to compare our "interpretation" of the documentation for those flags :smile: :There are some *interesting* points on flag computation !

PR #1008, opened 2019-03-16
'tuple' object has no attribute 'op'

jie-lin:
if len(left) is more than 1, it will occur 'tuple' object has no attribute 'op'


serpilliere:
Ho, I found the very same bug yesterday !
You are faster than me @jie-lin !
Thank you!

PR #1011, opened 2019-03-21
Add compose_and_mask simplification pass

icecr4ck:
First version of a simplification pass to simplify expressions like `ExprOp('&', ExprCompose(...), mask)`. This is the kind of output expected:
```
>>> from miasm.expression.simplifications import expr_simp
>>> from miasm.expression.expression import *
>>> expr_simp(ExprCompose(ExprInt(0, 24), ExprId('a', 8) & ExprInt(0xFF, 32))
0x0
```

### DIFF
--- a/miasm/expression/simplifications_common.py
+++ b/miasm/expression/simplifications_common.py
@@ -1570,8 +1570,8 @@ def simp_compose_and_mask(_, expr):
         return expr
     if not arg2.is_int():
         return expr
-    int2 = arg2.arg.arg
+    int2 = arg2.arg
     if (int2 + 1) & int2 != 0:
         return expr
-    mask_size = int2.bit_length() + 7 // 8
+    mask_size = int2.arg.bit_length() + 7 // 8
     return ExprSlice(arg1, 0, mask_size).zeroExtend(expr.size)
--- a/miasm/expression/simplifications_common.py
+++ b/miasm/expression/simplifications_common.py
@@ -5,7 +5,7 @@
 from future.utils import viewitems
 
 from miasm.expression.modint import mod_size2int, mod_size2uint
-from miasm.expression.expression import ExprInt, ExprSlice, ExprMem, \
+from miasm.expression.expression import ExprId, ExprInt, ExprSlice, ExprMem, \
     ExprCond, ExprOp, ExprCompose, TOK_INF_SIGNED, TOK_INF_UNSIGNED, \
     TOK_INF_EQUAL_SIGNED, TOK_INF_EQUAL_UNSIGNED, TOK_EQUAL
 from miasm.expression.expression_helper import parity, op_propag_cst, \
@@ -1556,9 +1556,9 @@ def simp_add_multiple(_, expr):
 
 def simp_compose_and_mask(_, expr):
     """
-    {X 0 8, Y 8 16} & 0xFF => X
-    {X 0 32} & 0xFFFF => X[0:16]
-    {X 0 8, Y 8 24, Z 24 32} & 0xFFFFFF => X|Y
+    {X 0 8, Y 8 32} & 0xFF => zeroExt(X)
+    {X 0 8, Y 8 16, Z 16 32} & 0xFFFF => {X 0 8, Y 8 16, 0x0 16 32}
+    {X 0 8, 0x123456 8 32} & 0xFFFFFF => {X 0 8, 0x1234 8 24, 0x0 24 32}
     """
     if not expr.is_op('&'):
         return expr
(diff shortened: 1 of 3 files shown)


commial:
Thanks for the PR! 

If you want to avoid the loop-style check, you can use the following formula: x is a mask if `(x+1)&x == 0` (for x != 0, but & 0 is already done in another simplification).

miasm/expression/simplifications_common.py
```diff
@@ -1554,3 +1553,24 @@ def simp_add_multiple(_, expr):
...
+    if len(expr.args) != 2:
+        return expr
+    arg1, arg2 = expr.args
+    if not arg1.is_compose():
+        return expr
+    if not arg2.is_int():
+        return expr
+    for i in range(8, arg1.size, 8):
+        if arg2.arg == ExprInt(0, i).mask.arg:
+            return ExprSlice(arg1, 0, i)
```
commial: A simplifier always returns an expression of the same size. Here, the `ExprSlice` will return a smaller expression. You can use `.zeroExtend(expr.size)` to fix it (trailing 0s)

icecr4ck: I just added the `.zeroExtend(expr.size)` and fix the loop, thanks for the comments!

miasm/expression/simplifications_common.py
```diff
@@ -1554,3 +1553,25 @@ def simp_add_multiple(_, expr):
...
+    # handle the case where arg2 = arg1.mask
+    if len(expr.args) != 2:
+        return expr
+    arg1, arg2 = expr.args
+    if not arg1.is_compose():
+        return expr
+    if not arg2.is_int():
+        return expr
+    int2 = arg2.arg.arg
+    if (int2 + 1) & int2 != 0:
```
serpilliere: Take care here:
- arg2 is an Expr
- arg2.arg is a moduint
- arg2.arg.arg is a python int

So in python int, 0xFFFFFFFF + 1 = 0x100000000
```python
>>> cst = ExprInt(0xFFFFFFFF, 32)
>>> print cst
0xFFFFFFFF
>>> print cst.arg
uint32(0xffffffff)
>>> print cst.arg.arg
4294967295
>>> print cst.arg.arg + 1
4294967296
>>>  
```

*but*
```python
>>> print cst.arg + 1
uint32(0x0)
>>> 
```
Maybe you want the second case here

serpilliere: Hum. Ok no. it works with and without the modulus computation.
So maybe you can just have:
```pyhton
int2 = int(arg2)
```
instead of:
```python
int2 = arg2.arg.arg
```

icecr4ck: Didn't see your latest comment earlier so I pushed a new commit replacing the python int by the moduint, do you want me to make the change with `int(arg2)`?

icecr4ck:
So as I was saying on gitter, I have this issue when I run the simplification tests.
```python
################################################################################
original:  {{a[0:8] 0 8, 0x0 8 64}[0:8] & 0x1 0 8, 0x0 8 64} new:  {a[0:8] & 0x1 0 8, 0x0 8 64}
################################################################################
original:  {{a[0:8] 0 8, 0x0 8 64}[0:32] & 0x1 0 32, 0x0 32 64} new:  {a[0:1] 0 1, 0x0 1 64}
  File "expression/simplifications.py", line 488, in <module>
    'bug in expr_simp_explicit simp(%s) is %s and should be %s' % (e_input, e_new, e_check)
ValueError: bug in expr_simp_explicit simp({{a[0:8] 0 8, 0x0 8 64}[0:32] & 0x1 0 32, 0x0 32 64}) is {a[0:1] 0 1, 0x0 1 64} and should be {{a[0:8] 0 8, 0x0 8 32} & 0x1 0 32, 0x0 32 64}
```
I guess this is due to my new simplification pass but I don't understand how it appends.

icecr4ck:
Giving my previous comment, do you think I should implement this simplification for other expressions like ExprSlice, ExprId and ExprMem?

miasm/expression/simplifications_common.py
serpilliere: Yes please: and sorry for the bad comment :disappointed:

icecr4ck: No problem :)

serpilliere:
In fact, it seems you slowly slip from simplification reductions to 'rewriting' rules:
I think the ` {X 0 8, Y 8 16} & 0xFF => X` is a reduction rule, as well as `{X 0 16, Y 16 32} & 0xFFFF => zeroExt(X)`.

But the `X & 0xFFFF => zeroExt(X[0:16])` is for me more a 'rewriting' rule, and I am not sure about its legitimacy here.
I mean, one day I can imagine a user doing a future PR with a new interesting reduction like:
`zeroExt(X[0:16]) => X & 0xFFFF` arguing the second one is simpler than the first one.

I know this is totally unfair and arbitrary here, but do you see my point?
It's a bit hard to decide which reduction rules are ok or not: maybe time shows us interesting ones.

icecr4ck:
Indeed, actually I was thinking this simplification is a bit too "agressive". Maybe I should keep it only for ExprCompose, and as you say, only for cases like `{X 0 8, Y 8 16} & 0xFF => X` or `{X 0 16, Y 16 32} & 0xFFFF => zeroExt(X)` but not for `{X 0 16, Y 16 32} & 0xFF => zeroExt(X[0:8])`. I guess that soon as it slices an ExprId, the simplification is not pertinent anymore.

serpilliere:
Ok :smiley: 
So can you rollback the commits to only include the `{X 0 8, Y 8 16} & 0xFF => X ` and the `{X 0 16, Y 16 32} & 0xFFFF => zeroExt(X)`?
Those ones seems legit to me!
Thank you for the update @icecr4ck

commial:
Thanks @icecr4ck for the contrib and taking into accounts our remarks!

PR #1017, opened 2019-03-28
Fix build on macOS with latest XCode

trufae:


trufae:
(didn't tried on previous versions)

serpilliere:
Hi @trufae !
Can you confirm that with this patch, Miasm builds on Mac ?

trufae:
Yep. Thats why i have submitted it :)

serpilliere:
A little question:
why do you swap the `__BYTE_ORDER` definition in windows environement?

trufae:
Just to be consistent in the way we define things on all the systems. Otherwise it hurts my eyes. Keep them in the same order. We can switch to alphabetical order too, but all those 3 defines should be happening in the same order in the 3 cases

trufae:
What's left to merge this?

serpilliere:
Sorry for the delay @trufae, and thank you for your fix!

trufae:
Thanks!

PR #1029, opened 2019-04-28
add LDMIA instructions

wzrybuddy:
im not ensure it right, but dis is ok.


wzrybuddy:
```
.text:0002BB80 8C C8                                         LDMIA           R0!, {R2,R3,R7}
.text:0002BB82 8C C1                                         STMIA           R1!, {R2,R3,R7}
.text:0002BB84 90 E8 CC 00                                   LDMIA.W         R0, {R2,R3,R6,R7}
.text:0002BB88 CC C1                                         STMIA           R1!, {R2,R3,R6,R7}

```

wzrybuddy:
```
.text:0001354C BA 46                                         MOVEQ           R10, R7
.text:0001354E 92 EA 03 0F                                   TEQ.W           R2, R3
.text:00013552 0B 46                                         MOV             R3, R1
```

serpilliere:
I am restarting CI here, as it seems the faulty test is due to bad travis.

wzrybuddy:
hi @serpilliere, i add test code for instruction, but why test/arch/arm/sem.py seems do not include armt semantic test? do you forget it or other reason? if i write some armt instruction semantic test, it' s a right place?

serpilliere:
Hi @wzrybuddy !
Currently in miasm, we have tests for the assembler/disassebmler in `test/arch/arm/arch.py` in which we try to assemble/disassemble each supported instruction (ok, maybe not *every* form of *every* instructions...)
It's a bit different for semantic: we only add regression tests for complex semantic or if the semantic is really prone to error.

To test the semantic, instead of testing the semantic directly, our preference goes to implement a little code and emulate it with special case, then check the registers/flags.

For example, here is the test for an instruction in aarch64, in `miasm/test/arch/aarch64/unit/mn_ubfm.py`:

```python
#! /usr/bin/env python2

import sys

from asm_test import Asm_Test
from pdb import pm


class Test_UBFM1(Asm_Test):
    TXT = '''
main:
       MOVZ    X0, 0x5600
       UBFM    X0, X0, 8, 15
       RET     LR
    '''
    def check(self):
        assert(self.myjit.cpu.X0 == 0x56)
        pass

class Test_UBFM2(Asm_Test):
    TXT = '''
main:
       MOVZ    X0, 0x56
       UBFM    X0, X0, 4, 55
       RET     LR
    '''
    def check(self):
        assert(self.myjit.cpu.X0 == 0x5)
        pass


if __name__ == "__main__":
    [test(*sys.argv[1:])() for test in [Test_UBFM1, Test_UBFM2 ]]

```


You have a bunch of these for x86 in miasm/test/arch/x86/unit/` if you want more examples!

serpilliere:
Ho, by the way, I will fix the tipo and relaunch your tests: it seems broken tests here are not your fault: it's only due to our bad english...

serpilliere:
I fix tipo with #1041 
Once it's ok and merged, I will hint you here, then you will be able to rebase your PR and fully pass regression tests.

serpilliere:
It seems some of your instructions have  a length of 31 bits.

miasm/core/cpu.py
```diff
@@ -1459,7 +1459,9 @@ def value(self, mode):
                     break
 
                 if f.value is not None and f.l:
-                    assert f.value <= f.lmask
+                    if f.value > f.lmask:
```
gamedbg: Why is this an assertion, which can be modified into a conditional statement?

serpilliere:
Hi @327135569 ,
Thank you for the PR update! (and sorry for the delay)
The PR seems ok. The only thing that may a problem is the semantic.
Maybe we could either:
- remove the semantic for the todo (which will create an error on run time if we encounter it =
- or explicitly add a raise` NotimplementedError, which will also raise an error at run time

I think this is necessary because if we let code empty, it will silently give an IR which will be wrong (as mnemonic will miss) and it is quite disturbing during analysis.

Are you ok to fix this or do you want me to do a pr on your pr ?

miasm/arch/arm/sem.py
```diff
@@ -974,31 +975,45 @@ def blx(ir, instr, a):
...
+        if a in dpregs_expr:
+            # two gp to one dp
+            # a[32:] = b
+            # a[:32] = c
+            e.append(ExprAssign(a, (c.zeroExtend(a.size) << ExprInt(32, 64) | b.zeroExtend(a.size))))
+        elif a in regs_expr:
+            # one dp to two gp
+            # a = c[32:]
+            # b = c[:32]
+            e.append(ExprAssign(a, (c & ExprInt(0xffffffff, 64))[0:32]))
```
serpilliere, line 992: Hey, you are close to a good semantic.
Just a remark here, as you are doing a [0:32] (which will extract the first 32 bits) you don't need to mask it before with the  `& 0xFFFFFFFF`

miasm/arch/arm/sem.py
```diff
@@ -974,31 +975,45 @@ def blx(ir, instr, a):
...
+            e.append(ExprAssign(b, ((c >> ExprInt(32, 64)) & ExprInt(0xffffffff, 64))[0:32]))
```
serpilliere, line 993: Same here, the bit truncation will do the mask job

serpilliere:
Hey @327135569 
It seems your are very close to the good semantic, I just added 2 remarks here are I think everything will be ok!

327135569:
Gosh, i am too slow to add semantic.. but seems OK, can you merge ?

PR #1034, opened 2019-05-07
[WIP] Armt: fix jitter it instr exception

gamedbg:
hi, when i run this PR's file jit_armt_itblock.py, i get a exception, so i write a patch for it.

the exception is throw by follow line
https://github.com/cea-sec/miasm/blob/142fb3411317e6ce63e9c3f9154530d9e65c9664/miasm/arch/arm/sem.py#L1758-L1759
because jitter run instruction one by one, so, the cond always TRUE.

i just test on armt arch.
this PR will slow jiter, becuase it add many work at every instr.
and maybe create other bug. 

can someone give me some tip, i will try fix again. thank you.

### DIFF
--- a/miasm/analysis/dse.py
+++ b/miasm/analysis/dse.py
@@ -50,6 +50,9 @@ Here are a few remainings TODO:
 from builtins import range
 from collections import namedtuple
 
+from miasm.arch.arm.sem import *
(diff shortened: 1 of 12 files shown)


gamedbg:
there is better way do this, i will try.

PR #1044, opened 2019-05-23
Fix jitter on macOS

### LINKED ISSUE #1038: Jitter: endianness for GCC and LLVM on macOS
Hello,

I tried to use the DSE on a x86-64 binary and I got this strange DriftException:
```
miasm.analysis.dse.DriftException: Drift of @64[0x13FFD0]: 1898898676 instead of 17646280302810628096
```

This happens when the instruction `PUSH 0x712EE4F4` is handled by the DSE and as you can see in the exception error, it looks more like an endianness issue:
* `hex(1898898676) = 0x712EE4F4`
* `hex(17646280302810628096) = 0xF4E42E7100000000` 

So I investigate a bit and the problem seems to come from this part of the code:
https://github.com/cea-sec/miasm/blob/4c2320b46250a8d6f8774e1218544b72a154cd8e/miasm/analysis/dse.py#L301-L308

Removing the `[::-1]` in `int(encode_hex(value_chr[::-1]), 16)` resolves the issue, however this is not the behaviour expected as x86 is a little endian architecture.

Also, looking at the following result makes me think the data is not stored correctly in memory as I get:
```
>>> data = sb.jitter.vm.get_mem(0x13FFD0, 8)
'\x00\x00\x00\x00q.\xe4\xf4'
>>> hex(unpack('<Q', sb.jitter.vm.get_mem(0x13FFD0, 8))[0])
'0xf4e42e7100000000'
```
And if I'm not mistaken, I should get:
```
>>> data = sb.jitter.vm.get_mem(0x13FFD0, 8)
'\xf4\xe4.q\x00\x00\x00\x00'
>>> hex(unpack('<Q', sb.jitter.vm.get_mem(0x13FFD0, 8))[0])
'0x712ee4f4'
```

Am I completely wrong (which is strongly possible) or is there a real issue?

Thanks!

icecr4ck:
PR to fix #1038 and "ld: can't link with bundle (MH_BUNDLE) only dylibs (MH_DYLIB)" error when using the Jitter on macOS.

### DIFF
--- a/miasm/jitter/vm_mngr.h
+++ b/miasm/jitter/vm_mngr.h
@@ -35,10 +35,8 @@
 
 #ifdef __APPLE__
 #define __BYTE_ORDER __BYTE_ORDER__
-#ifndef __BIG_ENDIAN
-#define __BIG_ENDIAN '>'
-#define __LITTLE_ENDIAN '<'
-#endif
+#define __BIG_ENDIAN BIG_ENDIAN
+#define __LITTLE_ENDIAN LITTLE_ENDIAN
 #elif defined(__NetBSD__) || defined(__OpenBSD__)
 #define __BYTE_ORDER _BYTE_ORDER
 #define __BIG_ENDIAN _BIG_ENDIAN
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@
 from __future__ import print_function
 from distutils.core import setup, Extension
 from distutils.util import get_platform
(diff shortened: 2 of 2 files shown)


serpilliere:
Hi @icecr4ck !
So, does this PR allows you to compile, install, and run tests with jitter on macOS?

icecr4ck:
Yep all the tests were passed successfully on my side.

serpilliere:
Thank you @icecr4ck !

PR #1047, opened 2019-06-17
fix typo in msp430's sem.py

Ninjananas:


serpilliere:
Thank you @Ninjananas !

PR #1049, opened 2019-06-17
Use llvmlite 0.29.0, which seems to fix the remaining issues under Wi…

aguinet:
…ndows

### DIFF
--- a/optional_requirements.txt
+++ b/optional_requirements.txt
@@ -1,3 +1,3 @@
 pycparser
 z3-solver==4.5.1.0
-llvmlite==0.26.0
+llvmlite==0.29.0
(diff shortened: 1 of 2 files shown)


XVilka:
Shouldn't be updated to 0.31.0 instead?  https://pypi.org/project/llvmlite/

serpilliere:
Currently, @aguinet have analyzed the div problem and opened a issue here:
https://github.com/numba/llvmlite/issues/495
We can upgrade the llvm, but it seems the div problem won't be solved until this open is resolved.

aguinet:
The div problem has been properly fixed here: https://github.com/cea-sec/miasm/pull/1153 . Closing this PR.

PR #1051, opened 2019-07-04
add information about the current bitstream in flen ('bs' and 'offset_b')

Ninjananas:


miasm/core/cpu.py
```diff
@@ -1085,7 +1087,10 @@ def guess_mnemo(cls, bs, attrib, pre_dis_info, offset):
             (l, fmask, fbits, fname, flen), vals = branch
 
             if flen is not None:
-                l = flen(attrib, fname_values)
+                try:
+                    l = flen(attrib, fname_values, bs, offset_b)
```
serpilliere: With this new code, flen can raise IOError as the user can get bits from the bs.
You must have the same behavior as getbits below in case of IOError.

miasm/core/cpu.py
```diff
@@ -1214,7 +1219,10 @@ def dis(cls, bs_o, mode_o = None, offset=0):
             total_l = 0
             for i, f in enumerate(c.fields_order):
                 if f.flen is not None:
-                    l = f.flen(mode, fname_values)
+                    try:
+                        l = f.flen(mode, fname_values, bs, offset_b)
```
serpilliere: Same here: flen may raise IOError, so you have to handle it like an IOError in getbits (I mean closing the atomic mode...)

miasm/core/cpu.py
Ninjananas: I fixed it in the latest commit.

miasm/core/cpu.py
Ninjananas: I fixed it in the latest commit too.

PR #1052, opened 2019-07-16
Update vm_mngr.c

qihuanpiaoliu:
Eliminate compile warnings on MAC.

MAC version = 10.12.6
clang version =  Apple LLVM version 9.0.0 (clang-900.0.37)

### DIFF
--- a/miasm/jitter/vm_mngr.c
+++ b/miasm/jitter/vm_mngr.c
@@ -226,7 +226,7 @@ static uint64_t memory_page_read(vm_mngr_t* vm_mngr, unsigned int my_size, uint6
 		}
 		switch(my_size){
 		case 8:
-			ret = ret;
+			/*ret = ret*/;
 			break;
 		case 16:
 			ret = set_endian16(vm_mngr, (uint16_t)ret);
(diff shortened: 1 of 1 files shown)


miasm/jitter/vm_mngr.c
```diff
@@ -926,7 +926,7 @@ void hexdump(char* m, unsigned int l)
       if (!(i%0x10) && i){
       last = i;
       printf("    ");
-      for (j=-0x10;j<0;j++){
+      for (j=-0x10;(int)j<0;j++){
```
serpilliere: Hey @qihuanpiaoliu !
Thank you for the fix.
For the last one, maybe we can declare an "int" (signed) `k` for example, and use it here to avoid the warning instead of casting j as unsigned. What do you thinl of this?

miasm/jitter/vm_mngr.c
```diff
@@ -226,7 +226,7 @@ static uint64_t memory_page_read(vm_mngr_t* vm_mngr, unsigned int my_size, uint6
 		}
 		switch(my_size){
 		case 8:
-			ret = ret;
+			/*ret = ret*/;
```
XVilka: Should be removed instead of commented I think?

PR #1057, opened 2019-08-08
Fix str/bytes handling in pe loader

carolineLe:
This is for Python 3 compatibility


serpilliere:
Thank you @carolineLe !

PR #1059, opened 2019-08-29
Travis: switch from trusty to xenial

WilliamBruneau:
see: https://travis-ci.community/t/cannot-apt-get-install-clang-5-0/3250


serpilliere:
Thank you @WilliamBruneau !

PR #1065, opened 2019-09-21
Fix bytes dllname

serpilliere:


miasm/analysis/sandbox.py
```diff
@@ -51,7 +51,9 @@ def __init__(self, fname, options, custom_methods=None, **kwargs):
         """
 
         # Initialize
-        self.fname = fname
+        if not isinstance(fname, bytes):
+            fname = fname.encode('utf8')
+        self.fname = bytes(fname)
```
p-l-: Is this needed? Either `fname` is a `bytes` instance, or it has been `.encode`d as UTF-8, which should result in bytes.

serpilliere: Correct!
Thank you :smile:

PR #1072, opened 2019-10-09
Fix IDA depgraph example

WilliamBruneau:


serpilliere:
Tahnk you @WilliamBruneau !

PR #1076, opened 2019-10-16
Jitter: android support

WilliamBruneau:


serpilliere:
Thank you for the fix @WilliamBruneau !

PR #1094, opened 2019-11-18
Fix semantics of REP prefix.

a-vincent:
Only REPE and REPNE exist at the binary level, but Miasm allows "REP" when
assembling x86 instructions. Make the semantics aware of this instead
of ignoring the REP prefix.
The bug could not occur when disassembling a binary program.


serpilliere:
Thank you @a-vincent

PR #1108, opened 2019-12-11
DSE: on mem_read, use parent behavior instead of no treatment

### LINKED ISSUE #1105: DSE mem_read issue
Hello,

I've noticed that DSE's symbolic execution engine treats some memory-reads in a different way than the pure one which results in some very "ugly" expressions and loss of precise values (I can provide examples if they are required). I found out that the cause is [here](https://github.com/cea-sec/miasm/blob/master/miasm/analysis/dse.py#L112). Changing the line to `return super(EmulatedSymbExec, self).mem_read(expr_mem)` fixes the issue. Is this a bug? If no, what are the reasons for this behavior?

commial:
Fix #1105 

credits @nofiv

### DIFF
--- a/miasm/analysis/dse.py
+++ b/miasm/analysis/dse.py
@@ -109,7 +109,7 @@ class ESETrackModif(EmulatedSymbExec):
 
     def mem_read(self, expr_mem):
         if not expr_mem.ptr.is_int():
-            return expr_mem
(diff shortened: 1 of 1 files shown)


serpilliere:
Thank you @commial  and @nofiv !

PR #1112, opened 2019-12-19
normalization of log messages

Sh3idan:
Messages from loggers have the same format.


serpilliere:
Thank you for the fix @Sh3idan !

PR #1119, opened 2020-01-03
Win_API: Fix str/bytes manipulations for set_mem/add_memory_page

WilliamBruneau:


serpilliere:
Thank you @WilliamBruneau !

PR #1120, opened 2020-01-06
Loader/PE: Initialize entry.originalfirstthunks if empty

carolineLe:


serpilliere:
Than you @carolineLe !

PR #1121, opened 2020-01-08
Win_API: Fix VirtualProtect on multiple areas

carolineLe:


serpilliere:
Hi!
It seems neither of us are correct on this case @carolineLe 
Your patch fixes some cases, but miss other ones. Example:
known memory range [0, 0x10000] and we do Virtual protect on 0x1000 with 0x10 bytes of size.
It breaks with this update.
(note that I agree with you that the current code is broken as well)

I think the "good" way to fix this is to mimic windows behavior:
```
The region of affected pages includes all pages containing one or more bytes in the range from the lpAddress parameter to (lpAddress+dwSize). This means that a 2-byte range straddling a page boundary causes the protection attributes of both pages to be changed.
```

The problem is that Miasm deals memory regions with start/stop addresses which are not backed by pagination.

So in case of changing the protection only on a subset of pages of an allocated region,  we may have to split the memory region using sizes and addresses of PAGES alignment, change their rights and re add those pages.

Are you ok with this or am I missing something?

carolineLe:
Hi !
Yes, I ended up having the same conclusion. 
Is there a way to free allocated memory areas in the jitter.vm ?

serpilliere:
Erf, not yet :smiley_cat: 
But maybe it's not very difficult to add it: it's a bit of C code and python.
Do you want to have a try at it?

(if you don't have time, tell me, I will add this)

serpilliere:
Thank you for the @carolineLe !

PR #1124, opened 2020-01-10
jitter.loader.pe: add docstrings for import/export accessors

williballenthin:


serpilliere:
Thank you @williballenthin !

Issue #1122, opened 2020-01-10
Add API Set Schema for Windows emulation


serpilliere:
Miasm needs to implement Windows 7 api set schema to support emulation in this environment.
Api redirections are stored in `ApiSetSchema.dll`.
We may add a kind of dictionary to store and use this information.

PR #1126, opened 2020-01-23
Change the returned value of OUT argument bool_ptr

stan1slas:
Hello!

I may be wrong, but I think that the value of bool_ptr is set to 1 in a case of a Win32 application running under SysWOW64. It's more detailed [here](https://docs.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-iswow64process).

Have a good day!

### DIFF
--- a/miasm/os_dep/win_api_x86_32.py
+++ b/miasm/os_dep/win_api_x86_32.py
@@ -1047,7 +1047,7 @@ def kernel32_GetSystemInfo(jitter):
 
 def kernel32_IsWow64Process(jitter):
     ret_ad, args = jitter.func_args_stdcall(["process", "bool_ptr"])
-    jitter.vm.set_u32(args.bool_ptr, 0)
+    jitter.vm.set_u32(args.bool_ptr, 1)
     jitter.func_ret_stdcall(ret_ad, 1)
 
 


serpilliere:
Hi @stan1slas 
I see the opposite of what you described:
```
If the process is running under 32-bit Windows, the value is set to FALSE.
```
Today, we mostly emulate 32 bits processes...
We didn't fix the #1122 for fully supporting 64 bits

stan1slas:
Well in my understanding, `32-bit Windows` is 32-bit Windows OS. However, on a 64 bits system OS, a 32 bits process is running under SysWOW64 so I guess we're in the case `A pointer to a value that is set to TRUE if the process is running under WOW64 on an Intel64 or x64 processor`.

I compiled that small program on a 64 bits windows as a 32 bits binary:
```
#include <windows.h>

int main(){
    int a, b;
    a = IsWow64Process(0xffffffff, &b);
    printf("%d %d", a, b);
    return 0;
}
```

The output: 
```
C:\Users\mlwr\Desktop>wow64.exe
1 1
```

So I guess that for 32 bits processes, the output will be 1 on a 64-bit system, and 0 on a 32-bit system.
I assumed that every person was running their 32-bits binaries on 64-bits systems, sorry!

serpilliere:
Anyway: maybe in the future we will support both systems (32 and 64 bits) so we will add this configuration in the `winobjs` object!
Are you ok to close the PR?

stan1slas:
Yep ! Thanks for your answer :)

PR #1127, opened 2020-01-27
Add 64bit structures for Windows (PEB, TEB, PEB_LDR_DATA)

Frky:
Minimal changes in `os_dep/win_XX_struct.py` and `os_dep/win_api_x86_XX_seh.py` to handle the loading of the following 64bit structures with the `PE` sandbox (`-y` option):
* `TEB`,
* `PEB`,
* `PEB_LDR_DATA`. 

A lot of additional changes would be required though, as many things were copied from the `32bit` version without any modification (...).

### DIFF
--- a/miasm/os_dep/common.py
+++ b/miasm/os_dep/common.py
@@ -71,15 +71,15 @@ class heap(object):
         self.addr &= self.mask ^ (self.align - 1)
         return ret
 
-    def alloc(self, jitter, size, perm=PAGE_READ | PAGE_WRITE):
+    def alloc(self, jitter, size, perm=PAGE_READ | PAGE_WRITE, cmt=""):
         """
         @jitter: a jitter instance
         @size: the size to allocate
         @perm: permission flags (see vm_alloc doc)
         """
-        return self.vm_alloc(jitter.vm, size, perm)
+        return self.vm_alloc(jitter.vm, size, perm=perm, cmt=cmt)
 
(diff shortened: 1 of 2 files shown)


serpilliere:
Lol, I updated a script *yesterday* with those structures... And said to myself "I may do a PR with this"...
Anyway: I will review your PR, and see if something differs from mine :smile: 
By the way, I have also done the `apiset`  with the `api-ms-xxx` stuff support for windows > 7 (apiset 8.1, 8.3 and win10).
But it needs some more tests.

I saw you add your files as |`win_seh_64`: I think if everything is ok, I will merge the PR like this: but if you are OK, later, I may move files in sub directories like `os_dep/win/win_xp` and `os_dep/win/win_7`  for example, because structures are not the same between a win 7 and a win 10 

May I ask you if the structures you gave here are from a win 7 or a win 10?

Frky:
I used this [this web page](http://bytepointer.com/resources/tebpeb64.htm) to get the structures,
Plus a binary file that is using these structures to get the list of imported DLLs.

According to the header of the web page, it should be the version for Windows Xp to 8, but I'm not sure if something differs in Win10 relatively to the fields I added (I only added the minimalistic fields used to get to the `PEB_LDR_DATA` structure).

Of course, it would make more sense to have these files in subdirectories as it is done for `linux`, I agree.

Final word: if you have time, you should definitely take a look at the ugly way I (we actually) did this, there probably is a much nicer way to do what I tried to do...

PR #1128, opened 2020-01-28
Fix bug in Microsoft x86-64 calling convention

Frky:
In Microsoft Windows 64bit, the `stdcall` calling convention allocates some space on the stack, called shadow space, corresponding to the parameters given in registers (cf [https://en.wikipedia.org/wiki/X86_calling_conventions](https://en.wikipedia.org/wiki/X86_calling_conventions#List_of_x86_calling_conventions)).

This means that, at the entry point of a function taking more than 4 parameters, the fifth parameter is at `RSP + 0x8 + (0x8 * 4)`.


serpilliere:
Thank you @Frky !

PR #1130, opened 2020-02-03
Add preliminary support for PPC MMU registers

IridiumXOR:
Hi, I have implemented a preliminary support for MMU registers in PowerPC 32bit. I have extended the operand MTSPR, MFSPR to support register SDR1 and BAT registers. I have also implemented the operands MTSR, MFSR to support SR  registers but I'm not able to implement MTSRIN and MFSRIN opcodes because destination register depends on the content of one of the arguments (see https://www.nxp.com/docs/en/reference-manual/MPCFPE32B.pdf pages 415 and 427)


serpilliere:
Hi @IridiumXOR !
Thank you for the PR.
A little question: today, on others architectures we raise an exception on *privileged*  instructions: this allow us to implement custom behavior in the handle responsible of privileged instructions.
For example, in x86, "mov cr0, eax"  will only raise an exception. I admin that this has its pro and cons. 

In your semantic, you really assign special registers assuming your code is run in supervised mode. This may raise an error if we are in user mode. 

I don't have a clean solution today:
- maybe we could add a "mode" flag to the cpu to raise or not an error in the semantic or something like?
- the semantic may become complex for those case, so may we could have a flag in the lifter in order to say is we are in user or privileged code?
- erf.

Anyone has an idea on this?

commial:
> Anyone has an idea on this?

Just a few thoughts:
* supporting a few privileged instruction could be nice, especially for architecture like ppc where it might allow us to run more firmware parts
* cr3 like and mmu related special register might not be so trivial to implement, given the way the memory work in Miasm
* the privileged status of an instruction could depends on its argument, like for the mov.

Given that, it wouldn't be too surprising for me to have conditional statements in the instruction semantic checking, for instance, the cpu current level. And depending on the result, having for now a exception raised, and for the few cases implemented, the expected behavior. IMHO, it would also provides a easy way for others to modify these specific behavior (rather than looking for how the code is lifted, etc.). It seems not so different of an exception raised because one of the argument is zero, for instance.

IridiumXOR:
I'm using miasm to emulate (in a symbolic way) part of a PPC kernel which operates on that registers and I want to catch which value is inserted in them. IMHO I support the idea of @commial of implementing the basic behavior of the instruction AND raise an exception in order to catch it and modify the current status of the emulation if it is needed. 

For the MTSRIN/MFSRIN instructions they work in this way: 

MTSRIN rS, rB  ====>
SR[rB[0:3]] <-- rS

the Segment register chosen for storing the value contained in rS depends only on the lowest 4 bits of the content of rB. The operation itself does not depends on other CPU flags. It is possible to implement using miasm's symbolic syntax? I'm not an expert of it.. If not, how can I catch the exception currently raised from miasm when this instruction is emulated?

a-vincent:
Hi,
the pull request is in line with the existing behaviour of Miasm on PPC and I'm fine with the pull request.

BTW, accesses to known-to-miasm common PPC SPR's are handled as assignments whether or not they require supervisor privilege. Accesses to unknown-to-miasm model-specific SPR's trigger an exception and fill spr_access with relevant information. Maybe that should be revisited someday to distinguish privileged accesses.

I think the discussion that started is much more ambitious than the patch: the patch is not trying to emulate a PPC MMU.

Regards,
 Aymeric

serpilliere:
Thank you @a-vincent for the feedback!
So let's merge this one, and keep the more complex implementation proposed by @commial for later.

@IridiumXOR : for the `MTSRIN` stuff, I am afraid there is no way to access to a register *through* another register number.
* One hackish way may be to do a kind of "memory mapped" registers for `SRs`. We could imagine a fake base address like `ExprId("SR_BASE_ADDRESS", 32)`, and replace each `SR` by it's corresponding memory lookup => `ExprMem(SR_BASE_ADDRESS + 4, 32)` for `SR1` for example. This will allow us to access any `SR` using a generic register, for example, `ExprMem(SR_BASE_ADDRESS + 4 * rB & 0xf)`. But the IR may be a bit more difficult to read and fake memory accesses will be created...
* We can generate IR code to "search" the good register. For your example `SR[rB[0:3]] <-- rS` the IR may be:
```
# -------------
loc_test_r0:
IRDst = (rb[0:3] == 0) ? (loc_do_r0, loc_test_r1)
# -------------
loc_do_r0:
sr0 = rs
IRDst = next
# -------------
loc_test_r1:
IRDst = (rb[0:3] == 1) ? (loc_do_r1, loc_else_r2)
# -------------
loc_do_r1:
sr1 = rs
IRDst = next
# -------------
...
```

But the IR will be a *mess*. The interesting point here is if in the code, the `rB` register comes from a constant, a pass of simplification of this code will delete *all* the bad branches. I think. :smile:

a-vincent:
@IridiumXOR, is it helpful for you if we merge the pull request as is for now, or do you prefer to amend it and also trigger an exception (like the one for mtspr) ? In the long run, it would be nice to distinguish between privileged and unprivileged instructions but I will not have time for this in the coming weeks, so it would rather be a different issue or pull request.

IridiumXOR:
It is helpful for me, at the moment, to pull the request as it. Thanks so much

a-vincent:
Thank you @IridiumXOR

PR #1131, opened 2020-02-04
win_api_x86_32: add further common windows API functions

williballenthin:
Found these while trying to get a naive "Hello world" VS2017 binary emulating under miasm. (doesn't yet work due to string table/MBS initialization. Not yet sure if I'll chase that down.)

### DIFF
--- a/miasm/os_dep/win_api_x86_32.py
+++ b/miasm/os_dep/win_api_x86_32.py
@@ -3034,3 +3034,244 @@ def msvcrt_strlen(jitter):
 
     s = get_win_str_a(jitter, args.src)
     jitter.func_ret_cdecl(ret_ad, len(s))
+
+
+def kernel32_QueryPerformanceCounter(jitter):
+    ret_ad, args = jitter.func_args_stdcall(["lpPerformanceCount"])
+    jitter.vm.set_mem(args.lpPerformanceCount, struct.pack('<Q', 0x1))
+    jitter.func_ret_stdcall(ret_ad, 1)
+
+
+def kernel32_InitializeCriticalSectionEx(jitter):
+    '''
+      LPCRITICAL_SECTION lpCriticalSection,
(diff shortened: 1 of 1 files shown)


serpilliere:
Hi @williballenthin , thank you for PR!
It's seems you are on your way to rewrite a complete windows in Python :smile: 

Just a few remarks: In `win_api_x86_32` there is a special object called the ` c_winobjs` (https://github.com/cea-sec/miasm/blob/master/miasm/os_dep/win_api_x86_32.py#L133)
This object represents the current "windows environment":  it stores various stuffs from registry keys to system date. This object is instantiated once (https://github.com/cea-sec/miasm/blob/master/miasm/os_dep/win_api_x86_32.py#L133) and APIs which needs one of those variables use the global variable `winobj`. This is a convenient way for a Miasm user to customize, let say a keyboard layout or the current date:
```python
from miasm.os_dep.win_api_x86_32 import winobjs
winobjs.hkey_handles.update({
    0x80000000: "hkey_classes_root",
...
})
```

As I can see, you have integrated some environment dependent variables in your APIs. Maybe It can be interesting to move them in it. 

Anyway, as you did the big part of the job, tell me if you want to give it a try or if you prefer I merge the PR like this (and we will patch this later) !

serpilliere:
Another remark:
In most cases, when we run "classic" windows binaries, we don't start at the entry point, but rather in the `winmain`: this avoid a lot of hard work (you did) like stack cookie initialisation. But this has some drawbacks: the heap pointer is not initialized and stuffs like that, so it has to be fixed 'manually'.

serpilliere:
I am merging the code like this: I will fix it in another PR

williballenthin:
Hopefully I'll get a chance to make the fixes before you do; though, it will be when my attention returns at work in a week or two.

williballenthin:
anyways, thanks for merging!

PR #1140, opened 2020-02-19
Modification of the SR subregisters offsets

6oclock:


serpilliere:
Thank you for the fix @6oclock !

PR #1141, opened 2020-02-19
Initial support for floating point and Altivec instructions in PPC arch

IridiumXOR:
Hello,
I have introduced all the floating points and Altivec registers in PPC and also some instructions. The instructions, at the moment, are NOT implemented (they call mn_do_nop_warn() ), however this commit permits to correctly disassemble some of the floating point and altivec instructions.


IridiumXOR:
The build fails for a misspelling (?)

serpilliere:
Yep.
We have codespell running on every PR. It thinks the `FPR` register is a typo of `FOR`.
You have to add `fpr` in the `.codespell_ignore` file (in the root of the repository) in your PR to fix codespell.

Can you do that, or do you want to fix this before your PR @IridiumXOR  ?

IridiumXOR:
Spelling fixed!

a-vincent:
Hi,

thank you. I'm fine with your patch except for the change in the semantics of "mftb": it is unrelated to floating point support.

However, it is an actual bug: it should be implemented by mfspr, not mfmsr (that's likely a typo on my side), and not a "no op" either. If you feel like it, don't hesitate to submit a separate PR, otherwise I'll do it.

Also, it would be nice if you could add a couple of tests for the disassembly of FPU instructions because, as you could guess, they don't pass through miasm often so there could be some regression there after a while.

Regards,
 Aymeric

IridiumXOR:
Please can you fix the bug for MFTB? 
Which type of test should I implement?

IridiumXOR:
I have inserted some regression test for floating point and Altivex instructions. Please can you merge the commit now? :)

a-vincent:
Hi, sorry this issue slipped off my mind. Thank you for your contribution!

serpilliere:
Thank you for the PR  @IridiumXOR  and for the review @a-vincent !

PR #1153, opened 2020-03-05
Include missing functions from llvm's compiler-rt into Jitllvm

aguinet:
It's still a hack to fix the "divsion" problem with LLVM, but seems to work under Linux! (haven't tried under Windows yet but hopefully it works)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,18 +11,22 @@ import fnmatch
 import io
 import os
 import platform
-from shutil import copy2
+from shutil import copy2, copyfile, rmtree
 import sys
+import tempfile
+import atexit
 
 is_win = platform.system() == "Windows"
 is_mac = platform.system() == "Darwin"
+is_64bit = platform.architecture()[0] == "64bit"
+if is_win:
+    import winreg
 
 def set_extension_compile_args(extension):
     rel_lib_path = extension.name.replace('.', '/')
     abs_lib_path = os.path.join(get_python_lib(), rel_lib_path)
     lib_name = abs_lib_path + '.so'
     extension.extra_link_args = [ '-Wl,-install_name,' + lib_name]
-
 
 class smart_install_data(install_data):
     """Replacement for distutils.command.install_data to handle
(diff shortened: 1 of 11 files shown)


serpilliere:
Hi @aguinet !
Sorry for the delay.
Thanks a lot for the idea: I tested on archlinux, we need one more function (divti3)
I did a PR on your PR to fix this, but maybe you prefer

aguinet:
I've squash your commit into mine.
It works for Linux, but IIRC under Windows the hack does not work.

The main remaining issue is how to have distutils build a shared library in a portable way. We might also fix this by launching command line by hand to generate this .so/.dll ...

serpilliere:
Can you rebase on master @aguinet ? I fixed the tipo erros in codespell to pass travis tests.
(sorry for that)

aguinet:
Done!

serpilliere:
Hope the Travis Gods will be on our side.

serpilliere:
The Travis Gods are pleased with your sacrifice. 
But maybe it's time to remove Win32 support.

serpilliere:
I am pushing a dummy PR to see if the windows 32 bit tests are due to this PR or if it's a coincidence.

aguinet:
So I will squash these commits into one!

About Windows 32, I will try on my local VM, but it seems to be a bigger issue.

There is also this https://ci.appveyor.com/project/cea-sec/miasm/builds/33068672/job/701prnol8kpicw4d, with an issue on dse_crackme.py, that can't delete a temporary file. Maybe a race between two tests?

serpilliere:
There is an issue on dse which appears at *runtime* randomly, but in most cases, the *build* is ok on appveyor.
See this #1235 I just did: https://ci.appveyor.com/project/cea-sec/miasm/builds/33069250
The windows tests passed on windows 32 bit.

But it seems your PR broke the test at *build time*, right?

aguinet:
Yes the issue on win 32 bits is at build time. I seem to be able to reproduce this locally. I'm looking into it.

aguinet:
So the issue under 32 bits is that clang still build 64 bits binaries. Indeed, the choice is normally made with the VS environment that is choosen. I will try and fix that properly :)

aguinet:
Other problem is that uint128_t does not exist for Windows 32-bit, even with Clang. IIRC it is the same for Linux 32 bits. I guess we have to disable the mn_div test for 32-bit builds!

aguinet:
Fixed by forcing the compilation with clang only for 64-bit Windows systems. I also disabled the mn_div.py test if we are not running on a 64-bit system.

aguinet:
All green @serpilliere !

setup.py
```diff
@@ -36,6 +40,48 @@ def run(self):
         ]
         return install_data.run(self)
 
+def win_get_llvm_reg():
+    REG_PATH = "SOFTWARE\\LLVM\\LLVM"
+    try:
+      return winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE, REG_PATH, 0, winreg.KEY_READ | winreg.KEY_WOW64_32KEY)
+    except FileNotFoundError:
+      pass
+    return winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE, "SOFTWARE\\LLVM\\LLVM", 0, winreg.KEY_READ)
```
serpilliere, line 49: Maybe we can use the variable `REG_PATH` here?

aguinet: Yes, copy/paste is a terrible feature :)

serpilliere:
Thanks a lot for the fix @aguinet !

PR #1154, opened 2020-03-06
Link to the IR Notebook from README

XVilka:
It was added in the repository, but remains largely "invisible".


serpilliere:
Thank you @XVilka !

Issue #1149, opened 2020-03-03
Add architecture, bitness, endianess and entrypoint in the loaders


XVilka:
A useful abstraction avoiding manual parsing of the headers.

serpilliere:
We may need something like a "projet"  or something like `angr` has.

XVilka:
I also think the `Machine` class should bitness and endianess fields.

XVilka:
I can send a patch with adding these properties if you are agree with this approach.

serpilliere:
Your proposition seems great for `Machine` yes @XVilka !

PR #1156, opened 2020-03-10
Add endianess and bitness in Machine

XVilka:
Addressing the suggestion from https://github.com/cea-sec/miasm/issues/1149


serpilliere:
May be I missed something here:
Why 16 bit for `armtl`? What does represent exactly your variable `bits?`

XVilka:
Sorry, might be a radare2 bias - we use 16bit for Thumb mode depiction. I will remove this part.

Issue #1162, opened 2020-03-21
Question on Linux x64 shellcode simulation


Te-k:
Hi,

I am analyzing a metasploit linux x64 shellcode. I would like to simulate it, I have used a copy of `x86_32.py` calling the x64 machine instead :

```py
myjit = Machine("x86_64").jitter(args.jitter)
myjit.init_stack()

data = open(args.filename, 'rb').read()
run_addr = 0x40000000
myjit.vm.add_memory_page(run_addr, PAGE_READ | PAGE_WRITE, data)
myjit.set_trace_log()
myjit.add_exception_handler(EXCEPT_INT_XX, exception_int)
myjit.run(run_addr)
```

When I run the simulation, I get an AssertionError on the syscall : 

```
R8  0000000000000000 R9  0000000000000000 R10 0000000000000022 R11 0000000000000000 R12 0000000000000000 R13 0000000000000000 R14 0000000000000000 R15 0000000000000000 zf 1 nf 0 of 0 cf 0
40000015 SYSCALL    
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
~/projects/shellcode/night/simu_sc_linux.py in <module>
     25 myjit.set_trace_log()
     26 myjit.add_exception_handler(EXCEPT_INT_XX, exception_int)
---> 27 myjit.run(run_addr)

~/tools/malware/miasm/miasm/jitter/jitload.py in run(self, addr)
    421         """
    422         self.init_run(addr)
--> 423         return self.continue_run()
    424 
    425 

~/tools/malware/miasm/miasm/jitter/jitload.py in continue_run(self, step)
    403         while self.run:
    404             try:
--> 405                 return next(self.run_iterator)
    406             except StopIteration:
    407                 pass

~/tools/malware/miasm/miasm/jitter/jitload.py in runiter_once(self, pc)
    371 
    372         # Exceptions should never be activated before run
--> 373         assert(self.get_exception() == 0)
    374 
    375         # Run the block at PC

AssertionError: 
```

I have a few questions here : 
* Is it possible to add an exception handler on syscall similarly to `EXCEPT_INT_XX` for `int 0x80` ?
* Is it possible to simulate syscalls with miasm ? (I am ideally trying to replicate scdbg with miasm)
* Is there a better place than here to ask for miasm questions ?

Thanks !

serpilliere:
Hi @Te-k !
[...]

PR #1163, opened 2020-03-22
Implements EXCEPT_SYSCALL

Te-k:
Hi,

Following our discussion #1162, I have implemented an exception for syscall `EXCEPT_SYSCALL` in a similar way than `EXCEPT_INT_XX`.  I have added an example of how to use it because it is honestly hard to navigate miasm features without it.

What do you think ?


serpilliere:
Hey, @Te-k ! Thanks for the PR!

It seems nearly good: Now the `syscall` raises the `SYSCALL` exception, you can remove the privilege instruction flag in the semantic!

Another point: I am afraid your example mislead the reader: If I understand it correctly, it will display the corresponding syscall name, but won't do anything. We could for example just support the "write to stdout" syscall, and fault on anything else (for example `raise Notimplemented` in other cases).

This way, if a user uses it to do other stuffs, he has a kind of skeleton and will modify it to support custom syscalls, and it won't 'fault' silently.

Are you ok with this?

serpilliere:
By the way: don't worry with the faulty scripts in the CI regressions tests, I will your PR to support the new syscall exception.

serpilliere:
Another point: For your tests, if you modify the semantic of an instruction, dont forget to remove your `/tmp/miasm_cache`  which contains the jitted cache modules. If you don't, miasm will use the *old* jitted basic blocks and you will wonder why your modifications are not taken into account.

Te-k:
So no need to raise `EXCEPT_PRIV_INSN` for the syscall instruction anymore ?

Regarding the example script, in many cases, just logging the syscall gives an idea of what the shellcode is doing, but indeed people may expect that it is actually emulating it. What about implementing one syscall, and writing "NotImplemented" clearly for others ? (so that it is clear but still works)

Te-k:
Here it is.

serpilliere:
Perfect!
Thanks a lot @Te-k .
I will complete your PR to fix the remaining regression tests
(Or feel free to do it if you want to give it a try)

Te-k:
I guess, here https://github.com/cea-sec/miasm/blob/master/miasm/os_dep/linux/syscall.py#L1081 it also needs to raise `EXCEPT_SYSCALL` ?

serpilliere:
Your analyze is correct @Te-k 
And another point is just here: https://github.com/cea-sec/miasm/blob/a040288c541f3b2d05fe1ce50f4c30a22572a3fb/miasm/os_dep/linux/syscall.py#L1005
We will have to clean the `EXCEPT_SYSCALL` instead of `EXCEPT_PRIV_INSN`

Te-k:
I can fix that later today.

Regarding the check failed, it does not seem to come from this PR, it also fails with the master branch. Here is the error :

```
  File "/home/travis/build/cea-sec/miasm/example/jitter/run_with_linuxenv.py", line 93, in <module>
    jitter.continue_run()
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/miasm/jitter/jitload.py", line 405, in continue_run
    return next(self.run_iterator)
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/miasm/jitter/jitload.py", line 373, in runiter_once
    assert(self.get_exception() == 0)
AssertionError
```

serpilliere:
Yes, it's linked to the PR:
With your PR, the syscall generates an exception which is not handled by the script (the script handles except_priv, not except_syscall)

That's why we should do the stuff I was telling about.

Te-k:
Here it is !

serpilliere:
Thanks a lot for the PR @Te-k!

PR #1168, opened 2020-03-30
Updt propagation expression algorithm

serpilliere:
This PR replaces the code responsible of "Expression propagation" used to simplify and deobfuscate code. The new code uses propagation of equivalence classes through basic blocks  in order to reveal possible simplifications.

For example:
```
    B = A+1
    C = A
    A = 6
    D = [B]
```
Will result in:
```
    B = A+1
    C = A
    A = 6
    D = [C+1]
```

To deal with pointer aliases, the code uses a "may interfer" which can be customized (sub classed) by the user.

This code is quicker, simpler, more correct than the actual `ExprPropagation`.

### DIFF
--- a/example/disasm/full.py
+++ b/example/disasm/full.py
@@ -10,8 +10,7 @@ from miasm.core.asmblock import log_asmblock, AsmCFG
 from miasm.core.interval import interval
 from miasm.analysis.machine import Machine
 from miasm.analysis.data_flow import \
-    DiGraphDefUse, ReachingDefinitions, \
-    replace_stack_vars, load_from_int, del_unused_edges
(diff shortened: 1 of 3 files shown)


serpilliere:
There is a bug in the `unssa` algorithm, but which is not introduced by this PR: the bug has been seen in the 0.1.3.
Here is a test case, the last step before unssa
ssa

And after the unssa
final


The problem is in 0x401008.
It may be linked to the fact that is this branch, after propagation, we have 2 living ssa variable of a same "Phi variable" (`EDI.0`and `EDI.1`) and maybe we have a bad test of live variable at this point.

serpilliere:
So I may merge this branch, as it's an improvement for the simplification engine, and track this bug down.

PR #1170, opened 2020-03-30
Syscalls linux32

Te-k:
Hi,

I have implemented a few more syscalls for Linux 32 (and some generic for Linux 64b too).
I am able to emulate several Linux 32b shellcodes with it, like :

```
[DEBUG   ]: socket(AF_INET, SOCK_STREAM, 0)
[DEBUG   ]: -> 3
WARNING: address 0x1240000 is not mapped in virtual memory:
[DEBUG   ]: socket_connect(fd, [AF_INET, 1234, 10.2.2.14], 102)
[DEBUG   ]: -> 0
Done
```

What do you think ?

### DIFF
--- a/miasm/os_dep/linux/environment.py
+++ b/miasm/os_dep/linux/environment.py
@@ -183,6 +183,9 @@ class FileDescriptorSocket(FileDescriptor):
         self.type_ = type_
         self.protocol = protocol
 
+    def read(self, count):
+        return b""
+
 
 class FileSystem(object):
     """File system abstraction
--- a/miasm/os_dep/linux/syscall.py
+++ b/miasm/os_dep/linux/syscall.py
@@ -68,6 +68,157 @@ def _dump_struct_stat_arml(info):
     return data
 
 
+def sys_x86_32_execve(jitter, linux_env):
+    # int execve(const char *pathname, char *const argv[],
+    #              char *const envp[]);
+    pathname_addr, argv_ptr, envp_addr = jitter.syscall_args_systemv(3)
+    pathname = jitter.get_c_str(pathname_addr)
+    argv = []
+    i = 0
+    argv_addr = jitter.vm.get_u32(argv_ptr)
+    while argv_addr != 0:
+        argv.append(jitter.get_c_str(argv_addr))
+        argv_ptr += 4
+        argv_addr = jitter.vm.get_u32(argv_ptr)
+    envp = []
+    i = 0
+    while envp_addr != 0:
+        argv.append(jitter.get_c_str(envp_addr))
+        i += 4
+        argv_addr = jitter.vm.get_u32(jitter.cpu.EDX+i)
+    log.debug("sys_execve(%s, [%s], [%s])", pathname,
+            ", ".join(argv), ", ".join(envp))
+    jitter.syscall_ret_systemv(0)
+
+
+def sys_x86_64_execve(jitter, linux_env):
+    # int execve(const char *pathname, char *const argv[],
+    #              char *const envp[]);
+    # TODO : merge that into a generic execve
+    pathname_addr, argv_ptr, envp_addr = jitter.syscall_args_systemv(3)
+    pathname = jitter.get_c_str(pathname_addr)
+    argv = []
+    i = 0
+    argv_addr = jitter.vm.get_u64(argv_ptr)
+    while argv_addr != 0:
+        argv.append(jitter.get_c_str(argv_addr))
+        argv_ptr += 8
+        argv_addr = jitter.vm.get_u64(argv_ptr)
+    envp = []
+    i = 0
+    while envp_addr != 0:
+        argv.append(jitter.get_c_str(envp_addr))
+        i += 8
+        argv_addr = jitter.vm.get_u64(jitter.cpu.EDX+i)
+    log.debug("sys_execve(%s, [%s], [%s])", pathname,
+            ", ".join(argv), ", ".join(envp))
+    jitter.syscall_ret_systemv(0)
+
+
+def sys_x86_32_socket(jitter, linux_env):
+    # int socketcall(int call, unsigned long *args)
+    # Redirect to several other socket syscalls
+    SOCKET_DOMAINS = {
+        0: "AF_UNSPEC",
+        1: "AF_UNIX",
+        2: "AF_INET",
+        3: "AF_AX25",
+        4: "AF_IPX",
+        5: "AF_APPLETALK",
+        6: "AF_NETROM",
+        7: "AF_BRIDGE",
+        8: "AF_AAL5",
+        9: "AF_X25", #Who cares ?
+        10: "AF_INET6",
+        11: "AF_MAX"
+    }
+
+    SOCKET_TYPE = {
+        1: "SOCK_STREAM",
+        2: "SOCK_DGRAM",
+        3: "SOCK_RAW"
+    }
+
+    if jitter.cpu.EBX == 1:
+        # int socket(int domain, int type, int protocol);
+        domain = jitter.vm.get_u32(jitter.cpu.ESP)
+        stype = jitter.vm.get_u32(jitter.cpu.ESP+4)
+        proto = jitter.vm.get_u32(jitter.cpu.ESP+8)
+        fd = linux_env.socket(domain, stype, proto)
+        log.debug("socket(%s, %s, %s)", SOCKET_DOMAINS[domain],
+                SOCKET_TYPE[stype], proto)
+        jitter.syscall_ret_systemv(fd)
+    elif jitter.cpu.EBX == 2:
+        # int bind(int sockfd, const struct sockaddr *addr,
+        #         socklen_t addrlen);
+        fd = jitter.vm.get_u32(jitter.cpu.ESP)
+        socklen = jitter.vm.get_u32(jitter.cpu.ESP+8)
+        sockaddr = jitter.vm.get_mem(jitter.cpu.ESP+4, socklen)
+        family = struct.unpack("H", sockaddr[0:2])[0]
+        if family == 2:
+            # IPv4
+            port = struct.unpack(">H", sockaddr[2:4])[0]
+            ip = ".".join([str(i) for i in struct.unpack("BBBB", sockaddr[4:8])])
+            log.debug("socket_bind(fd, [%s, %i, %s], %i)", SOCKET_DOMAINS[family],
+                    port, ip, socklen)
+        else:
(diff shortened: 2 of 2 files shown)


miasm/os_dep/linux/syscall.py
```diff
@@ -821,6 +985,14 @@ def sys_generic_setgid(jitter, linux_env):
         jitter.syscall_ret_systemv(0)
 
 
+def sys_generic_setreuid(jitter, linux_env):
+    # Parse arguments
+    ruid, euid = jitter.syscall_args_systemv(2)
+    log.debug("sys_setreuid(%x, %x)", ruid, euid)
```
serpilliere: We could use the current Linux env `uid/euid/gid` of the linux env here?

miasm/os_dep/linux/syscall.py
```diff
@@ -169,14 +176,22 @@ def sys_x86_32_socket(jitter, linux_env):
         #           socklen_t addrlen);
         fd = jitter.vm.get_u32(jitter.cpu.ESP)
         socklen = jitter.vm.get_u32(jitter.cpu.ESP+8)
-        # Not the exact size because shellcodes won't provide the full struct
-        sockaddr = jitter.vm.get_mem(jitter.vm.get_u32(jitter.cpu.ESP+4), 8)
+        try:
```
serpilliere: What about using the `socklen` instead of a fixed length?

miasm/os_dep/linux/syscall.py
```diff
@@ -68,6 +68,157 @@ def _dump_struct_stat_arml(info):
...
+        family = struct.unpack("H", sockaddr[0:2])[0]
+        if family == 2:
+            # IPv4
+            port = struct.unpack(">H", sockaddr[2:4])[0]
+            ip = ".".join([str(i) for i in struct.unpack("BBBB", sockaddr[4:8])])
+            log.debug("socket_connect(fd, [%s, %i, %s], %i)", SOCKET_DOMAINS[family],
+                    port, ip, socklen)
+        else:
+            log.debug("socket_connect(fd, sockaddr, socklen)")
+        jitter.syscall_ret_systemv(0)
```
serpilliere: We could return a new generated socket here instead of 0

miasm/os_dep/linux/syscall.py
Te-k: In some cases, a shellcode would not have a full sockaddr struct but just the needed fields, what I have done in the next commit is getting the socklen and if it fails, I only get the first 8 bytes. Is that an ok trick to have it work ?

Te-k:
All good points, thanks. I will fix these later this week.

miasm/os_dep/linux/syscall.py
serpilliere: In fact, maybe we should behave like the kernel does so we will be close to a real environment.
If the kernel is ok with semi structures, maybe your patch is ok.

serpilliere:
Thank you for you PR @Te-k !
If you really want to be sure we won't break anything in the future, maybe we could add a regression test of one of your shellcode (if you can share them, obvisouly), but put it in the https://github.com/cea-sec/miasm-extended-tests repository. Those tests are currently executed by the Miasm travis file.
The reason is simple:
Some times ago, we put a shellcode directly in the main repository, and the travis environment has flagged Miasm as malware and refused to run regression tests.
Maybe we should definitively not commit any shellcode/malware in the main repo, as it may be flagged as malware by PIP or distributions.

Another reason is to not add too many weight to the main repo.

miasm/os_dep/linux/environment.py
```diff
@@ -183,6 +183,9 @@ def __init__(self, number, family, type_, protocol):
         self.type_ = type_
         self.protocol = protocol
 
+    def read(self, count):
+        return b""
+
```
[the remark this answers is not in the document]
Te-k: I like the idea, but it seems hard to subclass, it means you have to implement a subclass of `FileDescriptorSocket`, `Network` and `LinuxEnvironment`, and make all this work together, right ?

miasm/os_dep/linux/syscall.py
```diff
@@ -68,6 +68,157 @@ def _dump_struct_stat_arml(info):
...
+    while argv_addr != 0:
+        argv.append(jitter.get_c_str(argv_addr))
+        argv_ptr += 4
+        argv_addr = jitter.vm.get_u32(argv_ptr)
+    envp = []
+    i = 0
+    while envp_addr != 0:
+        argv.append(jitter.get_c_str(envp_addr))
+        i += 4
+        argv_addr = jitter.vm.get_u32(jitter.cpu.EDX+i)
```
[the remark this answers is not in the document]
Te-k: Yep

miasm/os_dep/linux/syscall.py
```diff
@@ -68,6 +68,157 @@ def _dump_struct_stat_arml(info):
...
+    i = 0
+    argv_addr = jitter.vm.get_u32(argv_ptr)
+    while argv_addr != 0:
+        argv.append(jitter.get_c_str(argv_addr))
+        argv_ptr += 4
+        argv_addr = jitter.vm.get_u32(argv_ptr)
+    envp = []
+    i = 0
+    while envp_addr != 0:
+        argv.append(jitter.get_c_str(envp_addr))
```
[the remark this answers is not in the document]
Te-k: Also yes

miasm/os_dep/linux/syscall.py
```diff
@@ -68,6 +68,157 @@ def _dump_struct_stat_arml(info):
...
+    i = 0
+    argv_addr = jitter.vm.get_u64(argv_ptr)
+    while argv_addr != 0:
+        argv.append(jitter.get_c_str(argv_addr))
+        argv_ptr += 8
+        argv_addr = jitter.vm.get_u64(argv_ptr)
+    envp = []
+    i = 0
+    while envp_addr != 0:
+        argv.append(jitter.get_c_str(envp_addr))
```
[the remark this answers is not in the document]
Te-k: Damn copy paste

miasm/os_dep/linux/syscall.py
Te-k: I have added a first get_mem for the full structure and fallback to 8 bytes if the memory if not that large, it should be close to what the kernel is doing I guess

miasm/os_dep/linux/syscall.py
Te-k: So socket is already created a few lines above when `socket` is called, I am not sure what should be done here, because there is only so much we can emulate on that, right ?

Te-k:
I have made some fix based on your suggestions, two are still unresolved :
* Whether or not to implement read on sockets
* Should it create a socket on `connect` ? (not sure why it would)

Just one warning : I have added a change on uid and euid in `sys_generic_setreuid` and it does not check for privileges to do that, should I implement privileges here ?

Let me know what you think

Te-k:
And I have added a script in the examples to emulate Linux shellcodes, which is needed to add test cases to `miasm-extended-tests`

Te-k:
And here is the PR for the test https://github.com/cea-sec/miasm-extended-tests/pull/1 along with the update of travis config file (I have not tested it but it should be simple enough to work)

miasm/os_dep/linux/syscall.py
serpilliere: I totally miss read the switch case. The return case of *connect* seems good to me.
Sorry for this

miasm/os_dep/linux/environment.py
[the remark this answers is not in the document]
serpilliere: Kind of. It will give something like:
```python

class CustomFileDescriptorSocket(FileDescriptorSocket):
    def read(self, count):
        print("Turlututu")

class CustomNetworking(Networking):
    def socket(self, family, type_, protocol):
        fd = self.linux_env.next_fd()
        fdesc = CustomFileDescriptorSocket(fd, family, type_, protocol)
        self.linux_env.file_descriptors[fd] = fdesc
        return fd


class CustomLinuxEnvironment(LinuxEnvironment):
    def __init__(self):
        super(CustomLinuxEnvironment, self).__init__()
        self.network = CustomNetworking(self)
```

But maybe there is better: we could modify those classes to have a class variable which embed their needs. For example, for Networking:

```python
class Networking(object):
    """Network abstraction"""

    fd_generator = FileDescriptorSocket
    def __init__(self, linux_env):
        self.linux_env = linux_env

    def socket(self, family, type_, protocol):
        fd = self.linux_env.next_fd()
        fdesc = self.fd_generator(fd, family, type_, protocol)
        self.linux_env.file_descriptors[fd] = fdesc
        return fd

```

So the "overhead" may just be:
```python
class CustomNetworking(Networking):
    fd_generator = CustomFileDescriptorSocket
```


But I am not really sure if this is a suitable python pattern.
Or maybe `Networking` should take it's generator as `init` argument ?
It's a problem we already face in the `SandBox` object, which depends on os, arch, ...

@commial @p-l- I am interested if you have some feed on this.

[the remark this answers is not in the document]
commial: I'm not sure this is the same problem than the multiple inheritance of Sandbox. IMHO, it looks like more what we've done in `Jitcore` with `Cgen` or `SymbExecClass`, which reflects your last proposal.

In my opinion, the question is "what we want to provides, and what customization should be reasonably easy to implements?".
I agree with the fact that it should be easy to modify what the socket returns, its state, etc. i'm not sure that the more global Networking part needs that kind of customization possibility.

A pattern we can use would be to provides a kind of "socket factory" (sorry for this word, but it is what it is) that the Network would use to creates its sockets.
It could be a function, taking as input the socket parameters and returning an instance with the socket "interface", ie. a subclass of the socket fd.
It could also be a class, taking as `__init__` these parameters, and asked just after for successful creation or not (to keep the possibility to easiliy deny socket creation). I rather prefer the function solution, as it could be easier to return default implementation or several socket families implementation.

This "factory function" is then an attribute of the Networking class, and could be replaced with a dedicated function / property.

If this pattern become more frequent for the Linux kernel stub implementation, we could have a "config-like" class containing several factories functions, or hooks.

What do you think?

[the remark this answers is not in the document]
Te-k: So the socket factory would be an attribute of Network ?

Using it would be something like :

```py
class CustomFileDescriptorSocket(FileDescriptorSocket):
    def read(self, count):
        print("Chapeau pointu")

env = environment.LinuxEnvironment_x86_32()
env.network.socket_class = CustomFileDescriptorSocket
```

Is that correct ?

PR #1177, opened 2020-04-07
Minor improvement in rebuild_edges

nofiv:
If one deletes all the direct predecessors of a pending block, it dosn't get freed even thought it's not required anymore. Since all the pendings are recalculated here, there's no reason not to cover these cases here IMHO by clearing them prior the recalculation. Another solution is to add a watcher to del_block().


serpilliere:
Sounds great to me!

serpilliere:
Thank you @nofiv

PR #1184, opened 2020-04-14
Fix cpu.set_mem() -> vm.set_mem()

WilliamBruneau:


serpilliere:
Thank you @WilliamBruneau !

PR #1193, opened 2020-04-21
DepGraph minor improvement

nofiv:
Reaching a head with predecesors didn't stop dependency check(only yielded a result) and there's also no need to explicitly check for empty predecessors since there won't be any new states to process and the loop will just continue anyway I think, but I could be wrong. I'll also add that I recon that nodes with no predecessors are always included in the heads(it's required to yield a result in them now).

### DIFF
--- a/miasm/analysis/depgraph.py
+++ b/miasm/analysis/depgraph.py
@@ -615,12 +615,9 @@ class DependencyGraph(object):
             if done_state in done:
                 continue
             done.add(done_state)
-            if (not state.pending or
-                    state.loc_key in heads or
-                    not self._ircfg.predecessors(state.loc_key)):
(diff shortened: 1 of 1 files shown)


carolineLe:
Hi !
Thank you for your PR and sorry for the late answer.
I think the variable names might be a bit misleading here: 
- the 'heads' set contains the node where we want a result to be yielded
- it does not always contains the "real" heads of the CFG

The reason for not including the real CFG heads by default in what we call the "heads" set is that sometimes we don't want a result to be yielded when reaching the real head of the graph.

I hope it's clearer for you, don't hesitate to ask more information if not :)

nofiv:
Hi,

Thank you for the explanation, it's clear for me now. I'm closing the PR.

PR #1198, opened 2020-04-25
Rework the unpack_upx example

commial:
Hi,

The code in `update_binary` is actually linked to the "PE rebuilding", so move it to the "PE Rebuilding" part.
Also:
* add a few comments to explain what is happening
* write the explicit code to import and use `vm2pe`

In addition, IMHO it would be simpler for people to understand and base their script on a version without the detailed PE reconstruction, but only with the `vm2pe`.
If we want to illustrate the fine grain manipulation, it could be done in another dedicated example, likely in the `loader` section.

Thoughts ?

### DIFF
--- a/example/jitter/unpack_upx.py
+++ b/example/jitter/unpack_upx.py
@@ -1,9 +1,8 @@
 from __future__ import print_function
 import os
 import logging
-from pdb import pm
-from miasm.loader import pe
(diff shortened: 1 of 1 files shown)


commial:
PR Updated to actually remove the "explicit" part, and only use `vm2pe` (after IRL conversation with @serpilliere )

serpilliere:
Thank you @commial !

PR #1199, opened 2020-04-27
Fix jitter on Windows

jbgalet:
On Windows with Python38 the Jitter fails to compile mostly because of the native extension filenames `.cp38-win_amd64.lib`.
Therefore:
- during setup, some libs are misplaced
- bad extension in `jitcore_cc_base.py` (`sysconfig.get_config_var('EXT_SUFFIX')` is `.pyd` but we need `.lib` here)
- use hardcoded `python27.lib`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@ from distutils.sysconfig import get_python_lib, get_config_vars
 from distutils.dist import DistributionMetadata
 from distutils.command.install_data import install_data
 from tempfile import TemporaryFile
+import fnmatch
 import io
 import os
 import platform
@@ -248,7 +249,8 @@ def buil_all():
         for lib in libs:
             filename = os.path.basename(lib)
             dst = os.path.join(build_base, lib_dirname, "miasm", "jitter")
-            if filename not in ["VmMngr.lib", "Jitgcc.lib", "Jitllvm.lib"]:
+            # Windows built libraries may have a name like VmMngr.cp38-win_amd64.lib
+            if not any([fnmatch.fnmatch(filename, pattern) for pattern in ["VmMngr.*lib", "Jitgcc.*lib", "Jitllvm.*lib"]]):
                 dst = os.path.join(dst, "arch")
             dst = os.path.join(dst, filename)
             if not os.path.isfile(dst):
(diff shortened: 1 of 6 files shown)


serpilliere:
Hi @jbgalet !
If windows with python38 is now supported, can you upgrade the `.appveyor.yml` file in this PR to add the regression test for this platform?

serpilliere:
If you are not familiar with appveyor yml, don't hesitate to tell me, I will PR your PR with the fix!

serpilliere:
Thanks a lot @jbgalet!

Don't worry for the travis bugs, it's a llvm bug we are facing for some times.
Just a last request if you are ok: Can you merge your first and second commit ?
(the second one is just a fixup)

jbgalet:
Should be good, note that I had to update the llvmlite requirement to 0.31.0 to make it work, but it breaks travis (since llvmlite 0.31 requires llvm 7).
I have a .travis.yml that almost works here: https://github.com/jbgalet/miasm/blob/test-travis/.travis.yml
I can merge it as needed.

serpilliere:
Thank you @jbgalet !
If you are ok, you can PR the fix for the travis!

serpilliere:
@jbgalet , with your travis file, did you manage to make llvm tests pass again on travis ?

setup.py
```diff
@@ -248,7 +249,8 @@ def buil_all():
         for lib in libs:
             filename = os.path.basename(lib)
             dst = os.path.join(build_base, lib_dirname, "miasm", "jitter")
-            if filename not in ["VmMngr.lib", "Jitgcc.lib", "Jitllvm.lib"]:
+            # Windows built libraries may have a name like VmMngr.cp38-win_amd64.lib
+            if not any([fnmatch.fnmatch(filename, pattern) for pattern in ["VmMngr.*lib", "Jitgcc.*lib", "Jitllvm.*lib"]]):
```
p-l-, line 253: I know it's too late, but you don't need to build a list in such situations; this works:
```suggestion
            if not any(fnmatch.fnmatch(filename, pattern) for pattern in ["VmMngr.*lib", "Jitgcc.*lib", "Jitllvm.*lib"]):
```

jbgalet:
> @jbgalet , with your travis file, did you manage to make llvm tests pass again on travis ?

Not yet. I've talked to @commial and it seems there are some issues related to #1153

serpilliere:
Oh ok!
I thought you have found it was fixed in the last version of llvm.
Noproblemo.

PR #1200, opened 2020-04-27
upck32be vs upck16be

itsacoderepo:
I guess it is a copy & paste error :)


serpilliere:
Oh!
Nice one @itsacoderepo 
Thanks a lot!

PR #1202, opened 2020-04-28
Add MIPS MMU registers

IridiumXOR:
I have defined registers related to MMU. I have also add some registers related to CPU configuration (ConfigX, PRiD), exception table (EBase) and used in kernel mode (KScratchX)


IridiumXOR:
There are some errors related to x86 arch bu I haven't touched it...

serpilliere:
Yes, we are facing a llvm bug with compiler-rt.
Don't worried about this.

serpilliere:
Thank you @IridiumXOR !

PR #1203, opened 2020-04-28
Add new MIPS opcodes

IridiumXOR:
I have introduced the following working opcodes:
LH BNEL BEQL BLEZL BGTZL BC1TL BC1FL BLTZL
with the exception of LH, which loads an halfword, the other opcodes are the "likely" version of existing branch instructions: if the branch is not taken they do not execute the delay slot instruction.

I have introduced also these opcodes but I have implemented them with "nop" because I'm not so confident with IR expressions:
SWL SWR CFC1 CFT1

### DIFF
--- a/miasm/arch/mips32/arch.py
+++ b/miasm/arch/mips32/arch.py
@@ -47,8 +47,8 @@ class additional_info(object):
         self.except_on_instr = False
 
 br_0 = ['B', 'J', 'JR', 'BAL', 'JAL', 'JALR']
-br_1 = ['BGEZ', 'BLTZ', 'BGTZ', 'BLEZ', 'BC1T', 'BC1F']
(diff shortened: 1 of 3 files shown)


serpilliere:
Hi @IridiumXOR ,
Thank you for the PR!
Maybe it's better not to implement an empty mnemonic than implementing an empty one.
This way, if someone emulates code or translate it into intermediate language, it raises an error. In my opinion it's better than a silent "fail".
Can you modify your PR to remove the mnemonics with empty semantic ?
(just remove the semantic description not the disasm/asm description)

Another point: Can you update the regression test for the asm/disasm of the freshly added mnemonics ? (maybe you have bytes example of them)

Thank you again!

serpilliere:
Thank you @IridiumXOR !

PR #1204, opened 2020-04-28
Fix bug 'ins' MIPS opcode and PPC "byte_swap()"

IridiumXOR:
"ins" MIPS opcode and PPC byte_swap() use the old API for ExprCompose() which produces a "TypeError: unhashable type: 'list'" exception. I have convert them to use the new API.


serpilliere:
Thank you @IridiumXOR !

PR #1205, opened 2020-04-28
Fix ExprInt() without size in PPC

IridiumXOR:
Two expression in PPC sem.py using ExprInt() do not have size and raise "__new()__ missing 1 required positional argument: 'size'"


serpilliere:
Thank you @IridiumXOR !

PR #1210, opened 2020-04-29
Back out recent change from x.arg to int(x) for ExprMem's

a-vincent:
Also, use the newer x.ptr instead of x.arg.


a-vincent:
duplicate of #1211

PR #1211, opened 2020-04-29
Fix mn_do_lmw() and mn_do_stmw()

IridiumXOR:
PPC mn_do_lmw() and mn_do_stmw() raise exception because try to extract the source/destination address using int() from an ExprMem object.


a-vincent:
Thanks. I fixed it this morning, the pull request is still pending. :-)

serpilliere:
Thank you @IridiumXOR !

PR #1217, opened 2020-05-03
Fix win ldr first entry's blink

carolineLe:
In the Loader linked list initialization, the blink of the first entry was NULL in the following lists:
InMemoryOrderLinks
InInitializationOrderLinks
InLoadOrderModuleList

This PR fixes the blink to point to the real last entry of the linked list.


serpilliere:
(dont take the llvm's tests into account for the pr)

serpilliere:
Thank you @carolineLe !

PR #1223, opened 2020-05-12
Implementation of all Coprocessor 15 registers in ARM

IridiumXOR:
Hi,
I have implemented all the 32-bit coprocessor 15 registers of the ARM architecture. I have also introduced the opcode necessary to load/retrieve values from them (MCR/MRC) and fix a the opocodes for DSB/ISB for ARM mode


serpilliere:
Hi @IridiumXOR !
Can you add a disasm test for the `mcr` and `mrc` mnemonic ?

IridiumXOR:
I have added test units for MCR/MRC

serpilliere:
Thanks a lot @IridiumXOR !

PR #1224, opened 2020-05-14
Win_api: set lasterror on FindNextFile + encoding options

carolineLe:


serpilliere:
Thank you @carolineLe !

PR #1225, opened 2020-05-14
Win_api: fix argv index in shell32_CommandLineToArgvW

carolineLe:


serpilliere:
Thank you @carolineLe !

PR #1228, opened 2020-05-15
Jitter: new feature run_until

carolineLe:


Frky:
+1, this would be handy.

miasm/jitter/jitload.py
```diff
@@ -422,6 +428,18 @@ def run(self, addr):
...
 
+    def run_until(self, addr):
+        """PRE: init_run.
+        Continue the run of the current session until iterator returns, run is
+        set to False or addr is reached.
+        /!\ A previously set breakpoint at addr will be removed
+        Return the iterator value"""
+
+        def stop_exec(jitter):
+            jitter.remove_breakpoints_by_address(jitter.pc)
```
commial, line 439: Is it possible to remove the `stop_exec` function from the breakpoint callback instead of all the breakpoints at `jitter.pc`?

It would avoid weird behavior if a breakpoint is already registered, like:
```
jitter.add_breakpoint(0x1234, dump_pe)
jitter.run_until(0x1234)
```

carolineLe: Hi! Thanks for you review !
Unfortunately, it's not possible because I don't think it's possible to get a reference to the current function (it's a different object generated every time). 

Another option could be to define the callback outside, but it would mean that everytime a run_until breakpoint is reached, then all run_until breakpoints are removed, and that's not what is expected either.

In my view, `run_until` is an interactive feature, and not something we should use in a script. It's a convenient way to replace:
```
sb.jitter.vm.add_breakpoint(xxx, lambda _: False)
sb.jitter.continue_run()
```
So, in the case you mentionned, you would just need to make the function `dump_pe` return False and would likely not use `run_until`.

commial: Well, I'm not so sure about that, giving the following code:
```Python
def toto():
    print(toto)

toto()
```

Outs, in Python2 and Python3:
```
<function toto at 0x7f9059e918c0>
```

So it might actually works :) (Python being interpreted, there are more facilities :D )

I agree with you for the "spirit" of the feature; however, we always see a function re-used in several context, so if we can avoid this problem, it would be better. I also agree that it is probably an edge case.

carolineLe: Oh you're right ! I didn't think about that. Let me update this :)

serpilliere:
Thank you @carolineLe !

PR #1229, opened 2020-05-15
Mips new opcodes + bugfix

IridiumXOR:
Hi,
I have implemented this opcodes for MIPS
TNE, CLZ, LL, SC, SYNC, PREF, TLBR, MTHI, MTLO

I have also introduced this opcodes (only disasm)
CACHE, ERET

I have fix a typo (KSCRATCH0 -> KSCRATCH) and a little bug on opcode TEQ


miasm/arch/mips32/sem.py
```diff
@@ -488,6 +509,66 @@ def ei(arg1):
...
+@sbuild.parse
+def mthi(arg1):
+    R_HI = arg1
+
+@sbuild.parse
+def mtlo(arg1):
+    R_LOW = arg1
+
+@sbuild.parse
+def clz(rs, rd):
```
serpilliere, line 526: Maybe you can use the miasm operator "cntlz". For example, in arm, we have:
```python
def clz(ir, instr, a, b):
    e = []
    e.append(ExprAssign(a, ExprOp('cntleadzeros', b)))
    return e, []
```

IridiumXOR: I have replace it with 'cntleadzeros'

IridiumXOR: However, a "manual" implementation of clz (from which I have copied my past implementation) it is present in arch/mep/sem.py

serpilliere: Oh! I think I did a bad review of the mep before merging it.

serpilliere:
@IridiumXOR : the PR seems nearly ok to me:
Can you:
- merge your 2 commits together (to avoid having the fixing commit in the master)
- add another commit with some asm/disasm regression tests?

Tell me if you don't have time to do this: I will merge your PR and add tests later.

IridiumXOR:
Sorry, I was not able to merge commits :( I'm a noob with git..

serpilliere:
Noproblemo
Thank you for the fix @IridiumXOR !

PR #1230, opened 2020-05-18
Win_api: fix kernel32_MultiByteToWideChar

carolineLe:
According to the documentation, if cchWideChar is 0, then
the required size for the buffer indicated by lpWideCharStr
is returned.


serpilliere:
Thank you @carolineLe !

PR #1231, opened 2020-05-21
simp_cmp_bijective_op improvement

nofiv:
Input `expr` such as `(ESP_init + 0xFFFFFFA8) == (ESP_init + @32[ESP_init + 0xFFFFFECC] * 0x2 + 0xFFFFFFA8)` resulted in empty `args_a` which caused "sanitycheck: ExprOp args must have same size!" in `arg_a = ExprOp(op, *args_a)`.


serpilliere:
Nice catch @nofiv !

PR #1237, opened 2020-05-24
simplifications_common: added explicit assertion messages for division by 0

mrphrazer:
Hi!

`assert int(int2)` seems not to be the clearest way to prevent a division by 0. Took me some time to realize that it actually is no bug in the constant propagation engine. This PR is a proposal to be more specific here.

Best,

phrazer

### DIFF
--- a/miasm/expression/simplifications_common.py
+++ b/miasm/expression/simplifications_common.py
@@ -71,26 +71,28 @@ def simp_cst_propagation(e_s, expr):
                 shifter = int(int2) % int2.size
                 out = (int(int1) << shifter) | (int(int1) >> (int2.size - shifter))
             elif op_name == '/':
+                assert int(int2), "division by 0"
                 out = int(int1) // int(int2)
             elif op_name == '%':
+                assert int(int2), "division by 0"
                 out = int(int1) % int(int2)
             elif op_name == 'sdiv':
-                assert int(int2)
+                assert int(int2), "division by 0"
                 tmp1 = mod_size2int[int1.size](int(int1))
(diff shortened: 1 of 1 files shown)


serpilliere:
Hi @mrphrazer 
This should not happen as each architecture ensures a non zero division in the IR code.
Let take x86 for example:
div_graph

And the generated IR:
div_ir

So, if you enable constant propagation, for this example:
div_0_graph

And the propagation deletes the unused branch, so the result is:
div_0_propag


Have I missed something @mrphrazer ?

mrphrazer:
Hi @serpilliere,

I agree, in this case it should work fine. In our case, we are symbolically executing a path and get a purely symbolic expression. Then, we replace a variable with a constant and trigger the constant propagation via `expr_simp`:

[snippet.zip](https://github.com/cea-sec/miasm/files/4676704/snippet.zip)

In these cases, this triggers a division by zero at some point. Still, due to its assertion error, it first seems to triggering a bug in the constant propagation (but it does not)

Best

phrazer

serpilliere:
Ok, I see. But we may replace the assertion by something like:
```python
if XXX == 0:
    raise ZeroDivisionError("Division by 0 in cst propagation")
```
This way, we may let a chance to the user to handle this kind of error.
Or do you prefer the way you fixed it?
(maybe the zerdivision is overkill?)

mrphrazer:
Yeah I think that would be way better than assertions (with or without comment) :)

serpilliere:
Thank you for the fix @mrphrazer !

PR #1238, opened 2020-05-25
Jitter: add a trace option to 'run'

carolineLe:
Hi!
I added this option because I kept doing:

```
sb.jitter.set_trace_log()
sb.jitter.continue_run(step=True)
sb.jitter.set_trace_log(False, False, False)
```

And I think it's more convenient this way. What do you think ?


serpilliere:
Thank you @carolineLe !

PR #1239, opened 2020-05-26
Fix bug in kernel32_GetStartupInfo

Frky:
Extend the structure size to fit with the documentation : structure should be `0x44` bytes. All of these bytes should be initialized.

TODO : initialize correctly the fields of the structure, but I leave that to a future PR.


serpilliere:
Thank you for the fix @Frky !

PR #1242, opened 2020-05-27
Add apply_simp to DiGraphDefUse

nofiv:
Prevents simple FPs such as `xor eax, eax` just like in the `DependencyGraph`


serpilliere:
Thank you for the feature @nofiv !

PR #1243, opened 2020-05-28
Minimalistic implementation of kernel32_GetStringTypeW

Frky:


serpilliere:
Hi @Frky 
it seems that the windows tests didn't pass:
```
 from _curses import *
ImportError: No module named _curses
```
Maybe _curses doesn't exists on windows?

Frky:
Indeed. I use curses to get the set of characters for each type (*e.g.*, `C1_BLANK` or `C1_PUNCT`). I suggest that I hard-code the set of characters of each type instead, is that ok for you?

serpilliere:
Thank you @Frky !

PR #1245, opened 2020-05-30
Fix bug in WideCharToMultiByte

Frky:


serpilliere:
Thank you for the fix @Frky

PR #1247, opened 2020-06-02
Fix type error in msvcrt_wcsrchr and msvcrt_strrchr for Python 3

icecr4ck:
When using `rfind` in Python 3, it raises a TypeError when the type of the string and the type of the character searched are different.


serpilliere:
Thank you for the fix @icecr4ck !

PR #1248, opened 2020-06-03
Fix MultiByteToWideChar when cbMultiByte = -1

Frky:
When cbMultiByte = -1, the MultiByteStr input string is null-terminated.


serpilliere:
Thank you @Frky !

PR #1253, opened 2020-06-10
Fix collections

XVilka:
Most modules were moved from the root namespace of `collections` module to `collections.abc` since Python 3.2, and will stop work as is in the next Python 3.9 https://docs.python.org/3/library/collections.html

### DIFF
--- a/miasm/core/utils.py
+++ b/miasm/core/utils.py
@@ -3,7 +3,11 @@ import sys
 from builtins import range
 import struct
 import inspect
-from collections import MutableMapping as DictMixin
+
+try:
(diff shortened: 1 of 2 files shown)


serpilliere:
It seems your PR breaks python2.7
For now, we didn't take the decision to remove this support: Ghidra uses jython, which only supports python2.7

XVilka:
@serpilliere should be compatible now with both.

XVilka:
Any updates on this?

serpilliere:
Thank you for the fix @XVilka !
(and sorry for the delay !)

Issue #1255, opened 2020-06-15
miasm.core.types.BitField : incorrect backing num size check


w4kfu:
When a class `BitField` is initialized, there is a check on the sum of all bits used against the `backing_num` instance.

Here is a portion of the code from the `__init__` function: https://github.com/cea-sec/miasm/blob/975a527b4f268c4e41ab5ab259442c0941da6709/miasm/core/types.py#L997

This check should be this (or I am missing something?):

    if offset > self._num.size * 8:

Here is an example in order to reproduce the issue:

    from miasm.core.types import BitField, Num, MemUnion

    # THIS ONE SHOULD *NOT* RAISE AN EXCEPTION
    class TestBitStruct(MemUnion):
        fields = [
            ("ValueB", BitField(Num("<Q"), [
                ("field_00", 32),
                ("field_01", 32),
            ])),
            ("Value", Num("<Q")),
        ]

    # THIS ONE SHOULD RAISE AN EXCEPTION
    class TestBitStruct(MemUnion):
        fields = [
            ("ValueB", BitField(Num("<Q"), [
                ("field_00", 32),
                ("field_01", 32),
                ("field_02", 1),
            ])),
            ("Value", Num("<Q")),
        ]

serpilliere:
Nice catch @w4kfu !
Can you make a fix?
(or do you prefer me to fix this? tell me)

w4kfu:
I was waiting for your go, I just made a PR https://github.com/cea-sec/miasm/pull/1256 with the fix, it also contains the two test mentioned in this issue.

serpilliere:
Noproblemo @w4kfu 
Thank you for the fix!

PR #1256, opened 2020-06-16
Fix issue #1255 ; Add unit test

w4kfu:
Hello,

Here is a fix for the issue #1255, I also added two test, let me know if it's ok


serpilliere:
Thank you for the Fix @w4kfu !

PR #1264, opened 2020-07-08
Bug fix - tuples need to be converted to str before using force_bytes

nofiv:
to prevent TypeError during the subsequent byte object formatting

### DIFF
--- a/miasm/analysis/dse.py
+++ b/miasm/analysis/dse.py
@@ -250,7 +250,7 @@ class DSEEngine(object):
     def add_lib_handler(self, libimp, namespace):
         """Add search for handler based on a @libimp libimp instance
 
-        Known functions will be looked by {name}_symb in the @namespace
+        Known functions will be looked by {name}_symb or {name}_{ord}_symb in the @namespace
         """
         namespace = dict(
             (force_bytes(name), func) for name, func in viewitems(namespace)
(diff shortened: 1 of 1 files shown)


miasm/analysis/dse.py
```diff
@@ -258,11 +258,12 @@ def add_lib_handler(self, libimp, namespace):
 
         # lambda cannot contain statement
         def default_func(dse):
-            fname = b"%s_symb" % force_bytes(libimp.fad2cname[dse.jitter.pc])
+            fname = libimp.fad2cname[dse.jitter.pc]
+            fname = b"%s_symb" % force_bytes(str(fname))
```
p-l-, line 262: When `fname` is already a `bytes` object, this would raise a `TypeError`, which is probably not what you want.

marc-etienne:
This new patch shouldn't change the existing behavior when `libimp.fad2cname[dse.jitter.pc]` is `bytes` or `str` (or anything that isn't a `tuple`).

The docstring too was updated too to reflect the change.

serpilliere:
Hi @marc-etienne !
Thank you for the fix.
@nofiv : Did you test with the @marc-etienne fix ?
Is it ok for you ?

nofiv:
Hi @serpilliere , I've actually authored the fix and it works well for me. @marc-etienne maintains https://github.com/eset and reviewed the PR

marc-etienne:
Like he said. All credit goes to @nofiv. I'm simply the maintainer for the ESET GitHub organisation, trying to facilitate publishing stuff from ESET to GitHub.

serpilliere:
Thank you @nofiv and @marc-etienne for the fix!

PR #1269, opened 2020-07-24
output_interval was never updated and hence always empty

nofiv:
which was making the overlap check meaningless


serpilliere:
Oh! Nice catch.
Thanks a lot for the PR @nofiv !

PR #1272, opened 2020-08-18
Fix overflow if virtual memory page size >= 4GiB

IridiumXOR:
VmMngr suffers of a buffer overflow (which crashes the python interpreter) if you try to allocate a virtual memory page of size >= 4GiB.


serpilliere:
Hi @IridiumXOR,
Thank you for the report: Nice catch !
I made a mistake in the size test of an *INT_MAX*:
```
	if (buf_size > SIZE_MAX) {
		      fprintf(stderr, "Size too big\n");
		      exit(EXIT_FAILURE);
	}
```

serpilliere:
Tahnk you for the fix @IridiumXOR

PR #1273, opened 2020-08-18
Fix get_export_name_addr_list when there are no exports

Ana06:
When there are no exports, `get_export_name_addr_list` raises an exception at `miasm/jitter/loader/pe.py`, line 152:

```
AttributeError: 'DirExport' object has no attribute 'f_names'
```

Return an empty list instead.


serpilliere:
Thank you for the fix @Ana06 !

Issue #1129, opened 2020-01-29
run_at not working


Munroc:
I'm trying to run symbolic execution but run_at only returns the address. I traced the error to ircfg.get_block(addr) allways returning none.
Here is my code:

`from miasm.core.bin_stream                 import bin_stream_str
from miasm.arch.x86.disasm                 import dis_x86_32
from miasm.arch.x86.ira                    import ir_a_x86_32
from miasm.arch.x86.regs                   import all_regs_ids, all_regs_ids_init
from miasm.ir.symbexec                     import *
from miasm.expression.simplifications      import expr_simp

from miasm.analysis.machine import Machine
from miasm.core.locationdb import LocationDB
from miasm.arch.x86.arch import mn_x86

loc_db = LocationDB()
machine = Machine('x86_32')

# Binary path and offset of the target function
offset = 0x450 #A947B

# Get Miasm's binary stream
with open("adder", "rb") as binaryfile :
    bin_stream = bin_stream_str(binaryfile.read())

# Disassemble blocks of the function at 'offset'
mdis = machine.dis_engine(bin_stream)
asmcfg = mdis.dis_multiblock(offset)

ira = machine.ira(loc_db)
ircfg = ira.new_ircfg_from_asmcfg(asmcfg)

# Create symbolic execution engine
sb = SymbolicExecutionEngine(ira, machine.mn.regs.regs_init)
symbolic_pc = sb.run_at(ircfg, offset)
print ircfg.get_block(offset) #none
print symbolic_pc #offset`

serpilliere:
Hi @Munroc !
Did you try to graph the asmcfg?
(to dot file for example)
If it's ok, did you tryc to graph the ircfg? (dot as well)
Can you run the "run_at" with the argument step=True?

Munroc:
Hello! thanks for your answer,
Both graphs seems okay. 
[ircfg.dot.pdf](https://github.com/cea-sec/miasm/files/4131122/ircfg.dot.pdf)
[asmcfg.dot.pdf](https://github.com/cea-sec/miasm/files/4131124/asmcfg.dot.pdf)
Also I tried with step=True but same output unfortunately.

serpilliere:
What is the value of the returned symbolic pc?
What is the 'error'?

serpilliere:
> Also I tried with step=True but same output unfortunately.

You mean there is no additionnal logs duing the execution when using this flag?

Munroc:
[...]

Issue #873, opened 2018-10-21
Symbolic Execution does not terminate


sashs:
Hi,

I tried to analyse a small piece of code.
```
add byte ptr [rax - 0x77], cl; or byte ptr [r9], sil; rcr byte ptr [rbx + 0x5d], 0x41; pop rsp;               
```
However, the symbolic exection does not terminate. 

This is a simple script that reproduce that issue. 
```
from miasm2.analysis.machine import Machine                                                                          
from miasm2.core.locationdb import LocationDB                                                                        
from miasm2.analysis.binary import Container                                                                         
from miasm2.ir.symbexec import SymbolicExecutionEngine                                                               

code = "004889410831c05b5d415c".decode('hex')                                                                      
container = c = Container.from_string(code)                                                                          
loc_db = LocationDB()                                                                                                
                                                                                                                     
machine = Machine('x86_64')                                                                                          
mdis = machine.dis_engine(container.bin_stream)                                                                      
asmcfg = mdis.dis_multiblock(0)                                                                                      
print asmcfg                                                                                                         
ira = machine.ira(loc_db)                                                                                            
ircfg = ira.new_ircfg_from_asmcfg(asmcfg)                                                                            
for lb, irb in ircfg.blocks.items():                                                                                 
    print irb                                                                                                                                                                                        
sb = SymbolicExecutionEngine(ira)                                                                                    
                                                                                      
symbolic_pc = sb.run_at(ircfg,ircfg.blocks.keys()[0])    
```

I've also found other simple code pieces that leads to of an recursion depth exceedance.

cheers

serpilliere:
Hi!
I see the problem, but I am a bit in trouble to solve it:
The source is simple: you are creating a new `LocationDB` for the machine.ira, but the `dis_engine` has already created its own `LocationDB`. This messes the traduction from native to IR.

so I may give you this to fix your code,
```python
ira = machine.ira(mdis.loc_db)  
```
but I think we can do a fix in Miasm in order to avoid/detect this.
You can only manipulate objects linked to the very same LocationDB in miasm.
The point is that for now, a location_key has no link to its DB (on purpose). So we don't detect usage of a key belonging to another DB.

Am I clear?

sashs:
Yes, you are clear. Thank you!
That was too easy. :)

PR #1274, opened 2020-08-19
Avoid generate default locationdb

### LINKED ISSUE #1268: README example of symbolic execution not working
I tried following the steps of the README:

```python
from miasm.analysis.binary import Container
from miasm.arch.x86.arch import mn_x86
from miasm.core.locationdb import LocationDB
from miasm.ir.symbexec import SymbolicExecutionEngine

EXAMPLE_BYTES = b'\x8dI\x04\x8d[\x01\x80\xf9\x01t\x05\x8d[\xff\xeb\x03\x8d[\x01\x89\xd8\xc3'

def main():
    loc_db = LocationDB()
    c = Container.from_string(EXAMPLE_BYTES)
    machine = Machine("x86_32")
    mdis = machine.dis_engine(c.bin_stream)
    asmcfg = mdis.dis_multiblock(0)
    ira = machine.ira(loc_db)
    ircfg = ira.new_ircfg_from_asmcfg(asmcfg)
    sb = SymbolicExecutionEngine(ira)
    symbolic_pc = sb.run_at(ircfg, 0, step=True)
    print(symbolic_pc)

    sb = SymbolicExecutionEngine(ira, machine.mn.regs.regs_init)
    print(sb.run_at(ircfg, 0, step=True))
```

But I only get the following:
```
0
0
```

The README tells me that I should expect something like `((ECX + 0x4)[0:8] + 0xFF)?(0xB,0x10)` for the first line, and many more lines after that.

I'm running this on Arch Linux, I installed miasm via the [AUR package python-miasm-git](https://aur.archlinux.org/packages/python-miasm-git/) which installed the latest master (currently https://github.com/cea-sec/miasm/commit/a1c2dcbf6c4f3191984a781119d131c7c7f8d9ad).

serpilliere:
The `LocationDB` in Miasm everywhere: IR, loader, sandbox, ...
If some code share locations (`loc_key`), for example between the `IRCFG` and the `AsmCFG`, those locations *must absolutely* be part of the *same* `LocationDB`. If not, this may fail in a terrible and undetectable way. I mean, in a ghostbusters style, like "don't cross the streams" (https://www.youtube.com/watch?v=wyKQe_i9yyo)

The problem was that in Miasm, some objects take a `LocationDB` in their constructors, and some take an *optional* `LocationDB`. In the last case, the object was creating a new `locationDB` which must be used later.

This behavior is bad, as the user never really know when he has to create a `LocationDB` or where he has to find a default created one.

This PR fixes this behavior. No more object in Miasm will create a default `LocationDB`. If the `LocationDB` is needed, it will be explicit in the  object's constructor.

This will avoid #1268 #1129 #873 
Fixes #1268

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ Import the shellcode thanks to the `Container` abstraction:
 
 ```pycon
 >>> from miasm.analysis.binary import Container
->>> c = Container.from_string(s)
+>>> c = Container.from_string(s, loc_db)
 >>> c
 <miasm.analysis.binary.ContainerUnknown object at 0x7f34cefe6090>
 ```
@@ -182,7 +182,7 @@ Disassembling the shellcode at address `0`:
 ```pycon
 >>> from miasm.analysis.machine import Machine
 >>> machine = Machine('x86_32')
->>> mdis = machine.dis_engine(c.bin_stream)
+>>> mdis = machine.dis_engine(c.bin_stream, loc_db=loc_db)
 >>> asmcfg = mdis.dis_multiblock(0)
 >>> for block in asmcfg.blocks:
 ...  print(block.to_string(asmcfg.loc_db))
@@ -208,7 +208,7 @@ RET
 Initializing the Jit engine with a stack:
 
 ```pycon
->>> jitter = machine.jitter(jit_type='python')
+>>> jitter = machine.jitter(loc_db, jit_type='python')
 >>> jitter.init_stack()
 ```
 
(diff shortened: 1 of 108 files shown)


williballenthin:
I think this is a good enhancement that leads users to interact with miasm in a more correct way.

serpilliere:
The PR is getting bigger, but the `loc_db` usage should be a bit less messy.

serpilliere:
Travis callback seems to be broken, but regression tests are ok:
https://travis-ci.com/github/serpilliere/miasm/builds/181490981

PR #1277, opened 2020-08-20
Add offset() to miasm.core.asmblock.AsmBlock

Ana06:
I find myself calculating the offset of a block often with a code like `block.lines[0].offset`, which doesn't look really nice. I can write an `offset()` method in my code, but I think this function belongs to miasm. Can `miasm.core.asmblock.AsmBlock` have an `offset()` method/attribute? 🙏 

please

I have implemented the function with `min([line.offset for line in block.lines])` instead of `block.lines[0].offset`, because I think new lines could be added at the front with `addline()` and the lines wouldn't be ordered in that case.

### DIFF
--- a/miasm/core/asmblock.py
+++ b/miasm/core/asmblock.py
@@ -117,6 +117,15 @@ class AsmBlock(object):
     def addline(self, l):
         self.lines.append(l)
 
+    def offset(self):
+        """
+        returns the offset (as int) of the first line in the block
+        """
+        try:
+            return min(line.offset for line in block.lines)
+        except ValueError:
+            return None
+
     def addto(self, c):
         assert isinstance(self.bto, set)
         self.bto.add(c)


miasm/core/asmblock.py
```diff
@@ -117,6 +117,12 @@ def __str__(self):
     def addline(self, l):
         self.lines.append(l)
 
+    def offset(self):
+        """
+        returns the offset (as int) of the first line in the block
+        """
+        min([line.offset for line in block.lines])
```
p-l-, line 124: You don't need to create a list then compute the minimum value:
```suggestion
        min(line.offset for line in block.lines)
```

serpilliere:
Ok, maybe I have misunderstood your needs, but if you have disassembled a function, and have retrieved a block of it, you can use:
```python
block.loc_key
```
which is the location of the block.
And you can use:
```python
loc_db.get_location_offset(block.loc_key)
```
which will return the offset of the block, or None it the block is not 'pinned' to a specific location.
Does it respond to your needs ?

Ana06:
@serpilliere thanks for the detailed explanation! That solution sounds reasonable. I close this PR then. :+1:

PR #1283, opened 2020-08-31
Add cntleadzeroes, cnttrailzeroes call operators

IridiumXOR:
With this patch the python IR translator when reaches a 'cntleadzeros' or a 'cnttrailzeros' converts them in a call to a function with tha same name as the operator

cntleadzeors -> cntleadzeros(size, arg)
cnttrailzeros -> cnttrailzeros(size, arg)

### DIFF
--- a/miasm/ir/translators/python.py
+++ b/miasm/ir/translators/python.py
@@ -85,6 +85,9 @@ class TranslatorPython(Translator):
 
             return "((%s | %s) &0x%x)" % (part1, part2, int(expr.mask))
 
+        elif expr.op in ['cntleadzeros', 'cnttrailzeros']:
+            return "%s(0x%x, %s)" % (expr.op, expr.args[0].size, self.from_expr(arg))
+
         raise NotImplementedError("Unknown operator: %s" % expr.op)
 
     def from_ExprAssign(self, expr):


serpilliere:
Hi @IridiumXOR 
The goal of the Python translator is that we can take it's output and paste it in a python shell, after importing the expression module. Thus, if we apply your patch, this won't be possible anymore as the function `cntleadzeros` is not known in any module.

IridiumXOR:
I have assumed to use the same technique used for memory access (cal for the memory() function implemented in python) and also I followed the same strategy already used for the C translator. Otherwise, how can we implement that operands in python translator?

serpilliere:
Hi @IridiumXOR .
What do you mean by:
```I have assumed to use the same technique used for memory access (call the memory() function implemented in python) ``` ?

IridiumXOR:
Look here [https://github.com/cea-sec/miasm/blob/218492cd10b339a8d47d2fdbd61953fcf954fb8b/miasm/ir/translators/C.py#L178](https://github.com/cea-sec/miasm/blob/218492cd10b339a8d47d2fdbd61953fcf954fb8b/miasm/ir/translators/C.py#L178) the C translator converts the cntleadzeros in a function call (a call to bignum() function). In python I have implemented the same strategy.

PR #1288, opened 2020-09-09
Mandate passing loc_db to from_stream(). From_string() needs it.

a-vincent:


serpilliere:
Nice catch.
Thank you @a-vincent !

PR #1289, opened 2020-09-20
fix usage of asm_resolve_final and parse_txt

pcy190:
Seems the return value of `parse_txt` api was used in a wrong way.
Two changes:
1. obtain both asmcfg and loc_db of the `parse_txt`   
2. fill the lost parameter of the calling `asmblock.asm_resolve_final`

Anyway, probably there're many usage of `asm_resolve_final` which lost its third parameter, i.e. loc_db.

### DIFF
--- a/example/asm/simple.py
+++ b/example/asm/simple.py
@@ -8,7 +8,7 @@ from miasm.core.locationdb import LocationDB
 
 # Assemble code
 loc_db = LocationDB()
-asmcfg = parse_asm.parse_txt(
+asmcfg, loc_db = parse_asm.parse_txt(
     mn_x86, 32, '''
 main:
    MOV    EAX, 1
@@ -30,7 +30,7 @@ loop:
 loc_db.set_location_offset(loc_db.get_name_location("main"), 0x0)
 
 # Spread information and resolve instructions offset
-patches = asmblock.asm_resolve_final(mn_x86, asmcfg)
+patches = asmblock.asm_resolve_final(mn_x86, asmcfg, loc_db)
 
 # Show resolved asmcfg
 for block in asmcfg.blocks:


serpilliere:
Hi @pcy190 
There has been a API update on the LocationDB usage (see https://github.com/cea-sec/miasm/commit/80e40a3d2ca735db955807ad0605b43ca22e4e35, https://github.com/cea-sec/miasm/commit/6f5cb3bef180400001541f56f704f77da0d73429, https://github.com/cea-sec/miasm/commit/24ce193b8bad352853a9c5589f6fdcf5177d5466)
The regression tests on `simple.py` are ok on the master regarding this API modification.
It seems you are reverting the API update in this PR, is it correct ?

pcy190:
Well, the regression test looks fine.
I think I'm reverting the API unnecessarily.  
However, if I don't add the `loc_db` to receive the result of `asm_resolve_final`, I couldn't get the correct usage of asmcfg ( `asm_resolve_final`'s return value is a tuple containing asmcfg and loc_db )

FYI, I find and change this when I encounter the error, running the `example/asm/simple.py` individually.

serpilliere:
Maybe I have misunderstood something: Starting from the commits I gave, the result of `asm_resolve_final` is not a tupple any more:
https://github.com/cea-sec/miasm/blob/06239dde95cd984548deb40e9945d8bd85d83425/miasm/core/asmblock.py#L1131
It's a "path" set:
https://github.com/cea-sec/miasm/blob/06239dde95cd984548deb40e9945d8bd85d83425/miasm/core/asmblock.py#L1158

pcy190:
Indeed, perhaps my version is a little bit old.
After the commit you given, the asm_resolve_final is pure `patches` and example usage is correct.
( after I fetch the newest source, the problem disappear )

PR #1311, opened 2020-11-19
Add possibility to diff states in dumping symbexec state

sparkyparrot:
Add possibility to diff states in dump func of symbexec

### DIFF
--- a/miasm/ir/symbexec.py
+++ b/miasm/ir/symbexec.py
@@ -965,14 +965,15 @@ class SymbolicExecutionEngine(object):
                     continue
                 yield mem, value
 
-    def dump(self, ids=True, mems=True):
+    def dump(self,init_state_engine=None, ids=True, mems=True):
         """
         Display modififed variables
+        @init_state_engine: StateEngine instance
         @ids: display modified ids
         @mems: display modified memory
         """
 
-        for variable, value in self.modified(None, ids, mems):
+        for variable, value in self.modified(dict(init_state_engine if init_state_engine is not None else {}), ids, mems):
             print("%-18s" % variable, "=", "%s" % value)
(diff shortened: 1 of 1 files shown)


miasm/ir/symbexec.py
```diff
@@ -965,14 +965,15 @@ def modified(self, init_state=None, ids=True, mems=True):
...
+    def dump(self,init_state_engine=None, ids=True, mems=True):
         """
         Display modififed variables
+        @init_state_engine: StateEngine instance
         @ids: display modified ids
         @mems: display modified memory
         """
 
-        for variable, value in self.modified(None, ids, mems):
+        for variable, value in self.modified(dict(init_state_engine if init_state_engine is not None else {}), ids, mems):
```
serpilliere, line 976: Maybe the patch be better with:
```python
 def dump(self,init_state=None, ids=True, mems=True)
    ...
    if init_state is None:
        init_state ={}
    ...
    for variable, value in self.modified(init_state, ids, mems):
        ....
```

Is it ok for you ?

sparkyparrot: Hmmm just note that modified doesnt take in a state instance as the name implies but a base dictionary as its commented in the source code (so a dict(state) )
`@init_state: a base dictionary linking variables to their initial values`

So I guess passing dict() of a StateEngine or correcting the modified func as well.

serpilliere: Well, as the state is a kind of dictionary, I may keep the dictionary as argument if you are ok with this (the `.symbols` return a dictionary as well)

sparkyparrot: All goodie then !

Issue #1312, opened 2020-11-19
[AARCH64] Wrong disas of 'extr' instruction leading to assert when translating to IR


w4kfu:
Hello,

Maybe I'm missing something but I don't understand why the arguments of the instruction 'extr' in arm64 is using a signed integer for the 'imms' parameter: https://github.com/cea-sec/miasm/blob/bcb5fdf783909338eaa1246796200406d90f552b/miasm/arch/aarch64/arch.py#L2141
'simm6' is defined like this here:
https://github.com/cea-sec/miasm/blob/bcb5fdf783909338eaa1246796200406d90f552b/miasm/arch/aarch64/arch.py#L1735

When running the following test:
```python
from __future__ import print_function
from miasm.analysis.binary import Container
from miasm.analysis.machine import Machine
from miasm.core.locationdb import LocationDB

instrs = b"\x8C\x81\xCC\x93"        # EXTR            X12, X12, X12, #0x20
instrs += b"\xC0\x03\x5F\xD6"       # RET

loc_db = LocationDB()
cont = Container.from_string(instrs, loc_db)
machine = Machine("aarch64l")
mdis = machine.dis_engine(cont.bin_stream, loc_db=loc_db)
asm_block = mdis.dis_block(0)
print(asm_block)
ira = machine.ira(mdis.loc_db)
ircfg = ira.new_ircfg()
ira.add_asmblock_to_ircfg(asm_block, ircfg)
```
I got the following assert:
```
Traceback (most recent call last):
  File "fix_extr.py", line 17, in <module>
    ira.add_asmblock_to_ircfg(asm_block, ircfg)
  File "/usr/local/lib/python3.8/dist-packages/miasm/ir/ir.py", line 818, in add_asmblock_to_ircfg
    split = self.add_instr_to_current_state(
  File "/usr/local/lib/python3.8/dist-packages/miasm/ir/analysis.py", line 82, in add_instr_to_current_state
    assignblk, ir_blocks_extra = self.instr2ir(instr)
  File "/usr/local/lib/python3.8/dist-packages/miasm/ir/ir.py", line 745, in instr2ir
    ir_bloc_cur, extra_irblocks = self.get_ir(instr)
  File "/usr/local/lib/python3.8/dist-packages/miasm/arch/aarch64/sem.py", line 2305, in get_ir
    instr_ir, extra_ir = get_mnemo_expr(self, instr, *args)
[...]

PR #1313, opened 2020-11-19
[AARCH64] Fix 'extr' disassembly

w4kfu:
Hello,

Here is a fix for the issue : https://github.com/cea-sec/miasm/issues/1312.

Let me know what you think.

Thx,


serpilliere:
Hi @w4kfu !
Thank you for the fix!

PR #1314, opened 2020-11-19
[IR] Add signExt & zeroExt operator to python translator

w4kfu:
Hello,

While trying to translate to python an expression with the signext or zeroext operator I got the following assert:

```
  File "expr_ext.py", line 10, in <module>                                                               
    xpy = Translator.to_language('python').from_expr(x)                                                  
  File "/usr/local/lib/python3.8/dist-packages/miasm/ir/translators/translator.py", line 123, in from_expr
    ret = handler(expr)
  File "/usr/local/lib/python3.8/dist-packages/miasm/ir/translators/python.py", line 88, in from_ExprOp
    raise NotImplementedError("Unknown operator: %s" % expr.op)
NotImplementedError: Unknown operator: signExt_64
```

Here is a test script to reproduce the issue:

```python
from miasm.ir.translators.translator import Translator
from miasm.expression.expression import ExprId

x = ExprId('x', 64)
print(x)
x = x[:63].zeroExtend(64)
print(x)
x = x[:63].signExtend(64)
print(x)
xpy = Translator.to_language('python').from_expr(x)
print(xpy)
```
The changes are exactly the same that are present in the file "translator/C.py": https://github.com/cea-sec/miasm/blob/bcb5fdf783909338eaa1246796200406d90f552b/miasm/ir/translators/C.py#L156-#L175

### DIFF
--- a/miasm/ir/translators/python.py
+++ b/miasm/ir/translators/python.py
@@ -1,5 +1,6 @@
 from builtins import map
-from miasm.expression.expression import ExprInt
+from miasm.expression.expression import ExprInt, ExprCompose, ExprCond
+from miasm.core.utils import size2mask
 from miasm.ir.translators.translator import Translator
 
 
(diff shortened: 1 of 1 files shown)


serpilliere:
Hi @w4kfu !
I am torn by this PR.
In Miasm, there are a lot of operators: simple ones, like `+` or `*`, and operator that are a bit more high level, like `CC_EQ`, `zero_ext`, and even higher level, like `<u`.

We could implement all of them in translators for all languages (`python`, `C`, `z3`, ...). But what we did until now is to add 'simplification passes' which can translate high level operators into simple one.

For example, the https://github.com/cea-sec/miasm/blob/master/miasm/expression/simplifications_explicit.py contains rules which translate high level operators into very simple ones.

The fact is that today, if someone wants to have `IR` translated into python, the user has to first apply those rules in order to have a the "explicit" version of those operators, and then run the translation.
This avoid re implementing the pass which translate in explicit version in multiple location in Miasm code.

But, here is a counter argument: in certain languages, say `z3` for example, the destination language has the capacity to translate high level operators, for example the `<u` operator. So we may not want to first translate `<u` into explicit version, then translate this to `z3`.

Tell me if I am clear in those explanations.

w4kfu:
Hello, @serpilliere,

Thanks for taking the time to respond.

I was missing something and you point me in the right direction about how high level operator are translated into simple ones. I got your point, so it's totally normal that every operator has not been implemented in multiple locations.

Closing the PR.

serpilliere:
I am adding another precision here:
The semantic of each instruction is coded using high level operators (and not low level operators) as it's easy to translate high level operators into low level operators, but it can be very hard to translate low level operator into high level operators.

PR #1316, opened 2020-11-26
Fix propagate expression infinite loop

serpilliere:


miasm/analysis/data_flow.py
```diff
@@ -1755,6 +1755,12 @@ def __init__(self):
         self.equivalence_classes = UnionFind()
         self.undefined = set()
 
+    def __str__(self):
+        out = []
+        out.append(str(self.equivalence_classes))
+        out.append(str(self.undefined))
+        return "\n".join(out)
```
p-l-, line 1762: Hi @serpilliere!

FWIW
```suggestion
        return "%s\n%s" % (self.equivalence_classes, self.undefined)
```
Or:
```suggestion
        return "{0.equivalence_classes}\n{0.undefined}".format(self)
```

serpilliere: Oh! Thank you @p-l-.

PR #1318, opened 2020-11-30
Add missing == operator in translators

serpilliere:


miasm/ir/translators/smt2.py
```diff
@@ -226,6 +228,8 @@ def from_ExprOp(self, expr):
                     res = bv_rotate_left(res, arg, expr.size)
                 elif expr.op == ">>>":
                     res = bv_rotate_right(res, arg, expr.size)
+                elif expr.op == "==":
+                    ret = self.from_expr(ExprCond(expr.args[0] - expr.args[1], ExprInt(0, 1), ExprInt(1, 1)))
```
p-l-, line 232:
```suggestion
                    res = self.from_expr(ExprCond(expr.args[0] - expr.args[1], ExprInt(0, 1), ExprInt(1, 1)))
```

serpilliere: Oh. Double thanks @p-l- , as it seems we need a regression test here: this PR should have failed here

PR #1330, opened 2020-12-23
Fix del dummy phi

serpilliere:
Remove phi nodes using equivalence class algorithm


tly000:
you forgot to remove the debug outputs

serpilliere:
Oups. Correct. Thank you @tly000

PR #1350, opened 2021-01-14
Example/loader: add loc_db to Container.from_stream

carolineLe:


serpilliere:
Nice catch @carolineLe 
Thank you for the fix!
I must add this example to the refression tests as well

PR #1355, opened 2021-02-04
Add future to install_requires in setup

Ana06:
The future library is used in `miasm/core/bin_stream.py`: `from future.utils import PY3`

When running miasm (with Python 3.9.1) without future installed, I get the following error:
`ModuleNotFoundError: No module named 'future'`

I assume that future needs to be installed to run miasm (and consequently needs to be added to the dependencies). Or am I doing something wrong?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -242,6 +242,7 @@ def buil_all():
                         "VERSION"
                     ]
                 },
+                install_requires=['future', 'pyparsing'],
                 cmdclass={"install_data": smart_install_data},
                 ext_modules = ext_modules,
                 # Metadata


serpilliere:
Hi @Ana06 
In fact, `future` is already present in the `requirements.txt`
I am a little bit confused here between the `requirements.txt` and the `intall_requires` of the setup
@p-l- have you got an advise on where must we add the dependency ?

Ana06:
@serpilliere 

> In fact, future is already present in the requirements.txt
> I am a little bit confused here between the requirements.txt and the intall_requires of the setup

According to [Python's documentation](https://packaging.python.org/discussions/install-requires-vs-requirements), `install_requires` should only include the libraries that miasm minimally needs to run correctly, while `requirements.txt` could include more development libraries and it is not analyzed by pip during an install. It seems like the `requirements.txt` file is used for Docker, AppVeyor and travis: https://github.com/cea-sec/miasm/search?q=requirements, so I think it is just needed to add `install_requires` to `setup.py`.

Btw, I think that `pyparsing` should also be included in `install_requires`. I didn't get an error for `pyparsing` but it seems it is also needed to run miasm. Isn't it?

serpilliere:
Thanks for the explanation @Ana06 ! 
Yes it's a hard dependence as the assembler uses it.

p-l-:
Hi all,
Disclaimer: this is what I understand, I am not certain this is the complete truth...
I do understand the same than @Ana06 that you should add "hard" requirements to the `install_requires` value (the name looks weird to me, as it sounds like "needed to install the project", while it means "needed to use the project").
You can also add optional dependencies to `extra_require`.
Also `requirements.txt` is used by `pip` (and only by it).

serpilliere:
Ok, so, if I understand correctly, the goal here will be to mirror the content of `install_requires` and the `requirements.txt` right?
@Ana06, are you ok to fix your PR with this? Or do you prefer I merge this one and I PR another fix with the remaining dependence ?

p-l-:
@serpilliere that's what I think.

Ana06:
@serpilliere 
> @Ana06, are you ok to fix your PR with this? Or do you prefer I merge this one and I PR another fix with the remaining dependence ?

Updated! 😉

p-l- (approved):

serpilliere:
Thank a lots for the fix!

PR #1359, opened 2021-03-17
irb_out_ndes -> irb_out_nodes

yasong:


serpilliere:
Thank you for the fix @yasong

PR #1362, opened 2021-03-23
Fix sem.py in MIPS32

THUzxj:
- Fix `sb` & `sh` instruction
- Fix m2_expr import

But there are still warnings in my editor, like this
```python
@sbuild.parse
def jal(arg1):
    "Jumps to the calculated address @arg1 and stores the return address in $RA"
    PC = arg1
    ir.IRDst = arg1
    RA = m2_expr.ExprLoc(ir.get_next_break_loc_key(instr), RA.size)
```
It warns that `ir` & `instr` are undefined. I don't know whether it can work normally with `@sbuild.parse`.


serpilliere:
Thanks a lot for the PR  @THUzxj !

Issue #1367, opened 2021-05-03
Run or Run ?


Summus-git:
Hello, 
I juste wanted to point on a duplicate name that is quite dangerous :
https://github.com/cea-sec/miasm/blob/master/miasm/jitter/jitload.py#L415

```python
class Jitter(object):
    # ...
    def init_run(self, pc):
        # ...
        self.run = True

    def continue_run(self, step=False, trace=False):
        # ...
        while self.run:
        # ...

    def run(self, addr):
        """
        Launch emulation
        @addr: (int) start address
        """
        self.init_run(addr)
        return self.continue_run()
```

Used in https://github.com/cea-sec/miasm/blob/master/example/jitter/x86_32.py#L34  
`Jitter.run` is firstly a method that calls `Jitter.init_run` that overwrite `Jitter.run` as a boolean  
So that, a second call to `myjit.run(run_addr)` will lead into an exception   
`TypeError: 'bool' object is not callable`

```python
myjit = Machine("x86_32").jitter(loc_db, args.jitter)
# ...
if myjit.run: print("Running")  # -> "Running"
myjit.run(run_addr) # -> works
print(myjit.run) # -> False
myjit.run(run_addr) # exception
```

You could rename the boolean `run` to `running` or find another name for (or remove) the `run method`  
I can do a PR if you tell me your choice :)

serpilliere:
That's a long one that we wanted to fix, but always forgot.
You are walcome for a PR with the propose renaming (run -> running for the bool)

Thank you @Summus-git !

PR #1368, opened 2021-05-05
Replace jitter.run boolean by jitter.running

Summus-git:
Linked to #1367


serpilliere:
Tank you for the fix @Summus-git !

PR #1370, opened 2021-06-01
Symbols are str instead of bytes

serpilliere:


miasm/core/parse_asm.py
```diff
@@ -174,7 +175,7 @@ def parse_txt(mnemo, attrib, txt, loc_db):
         # label
         match_re = LABEL_RE.match(line)
         if match_re:
-            label_name = match_re.group(1).encode()
+            label_name = force_str(match_re.group(1))
```
p-l-, line 178:
```suggestion
            label_name = match_re.group(1)
```
Unless I am missing something here, since `LABEL_RE` is a `Pattern[str]`, `match_re` is a `Match[str]` and `match_re.group(1)` is a `str`.

serpilliere: Correct @p-l- !
Thank you for the feed back. I will update the PR soon

serpilliere:
Thank you for the review @p-l- !

PR #1371, opened 2021-06-04
fixed a bug when building on Python 2.7 on Windows

herosi:


serpilliere:
Hi @herosi 
I am a bit confused: I didn't see any errors on the regression tests launched on Appveyor, for python 2.7 or python 3.8
So, in which cases the bug you are fixing occurs?

herosi:
Thanks for asking me.

Hmm, I'm not sure why it works on Appveyor.
However, on Python 2, there is no module name "winreg". It is for Python 3.
Please see this.

https://docs.python.org/2.7/library/_winreg.html
`Note The _winreg module has been renamed to winreg in Python 3. The 2to3 tool will automatically adapt imports when converting your sources to Python 3.
`
And when I tried to import winreg in my environment, it failed with an ImportError exception.
```
> py -2
Python 2.7.18 (v2.7.18:8d21aa21f2, Apr 20 2020, 13:25:05) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import _winreg
>>> import winreg
  File "<stdin>", line 1, in <module>
ImportError: No module named winreg
>>>

```

And the original setup.py gets the same error in my environment.
And "except FileNotFoundError" did not work on python 2 as well because there are no names on python 2.

Could you check it on your environment?
thanks!

williballenthin:
image

looks like the [future](https://github.com/PythonCharmers/python-future) library provides `winreg` under `winreg`, too. miasm depends on `future`: https://github.com/cea-sec/miasm/blob/master/requirements.txt#L2 so i think it can rely on `winreg` being present without this PR.

herosi:
Ah, I see. Thanks @williballenthin and @serpilliere .
I compiled it on the environment and I use it on a different environment.
And the different environment has the feature module, but another one does not have.
Sorry to bother you guys.
And thanks again.

serpilliere:
That you for the hint @williballenthin !

PR #1373, opened 2021-06-09
Updated IDA API call for IDA 7.x

su-vikas:
`idc.GetReg` API call is changed to `idc.get_sreg` in IDA 7.x

https://hex-rays.com/products/ida/support/ida74_idapython_no_bc695_porting_guide.shtml


serpilliere:
Thank you @su-vikas

Issue #1375, opened 2021-06-13
x86_64 Cannot disasm mov with segment


KKomarov:
Failed to disasm `mov qword ptr es:[rax], rdx` (`26 48 48 89 10` in hex)
Tested on recent master https://github.com/cea-sec/miasm/commit/42528eff158475d1eca930b06754003242d5f1c1

```
from miasm.analysis.machine import Machine

machine = Machine('x86_64')
machine.mn.dis(bytearray.fromhex('26 48 48 89 10'), 64)


Traceback (most recent call last):
  File "<input>", line 3, in <module>
  File "..\miasm-repo\miasm\core\cpu.py", line 1276, in dis
    raise Disasm_Exception('cannot disasm at %X' % offset_o)
miasm.core.utils.Disasm_Exception: cannot disasm at 0
```

serpilliere:
Hi @KKomarov 
If I read correctly your assebmyl, it's:
```
26 48 48 89 10 
```
which is composed of:
- 26: the x86 legacy prefix (for es segment use)
- 48 the REX prefix
- and here is the strange thing: you have a *second* REX prefix ?
- the remaining bytes seem legit to me, for the mov instruciton.

Maybe I am wrong here, but I think you can have only *one* (at maximum) REX prefix, so this instruction seems not to be legit as it's composed of two REX.

Have I missed something @KKomarov ?

serpilliere:
Hum, I just get this on https://wiki.osdev.org/X86-64_Instruction_Encoding#REX_prefix:
```
The use of multiple REX prefixes is undefined, although processors seem to use only the last REX prefix. 
```

The current Miasm implementation only take one prefix. Maybe we could to some modifications in order to interpret the last REX in case of multiple REX

KKomarov:
@serpilliere I just take this from real binary. It seems like a part of anti-analysing feature. 
I tested Ida and Zydis they treat this fine

KKomarov:
@serpilliere Is it an easy fix? If you can tell what should be done I can make a PR

serpilliere:
Hi @KKomarov 
yep: the code here: https://github.com/cea-sec/miasm/blob/master/miasm/arch/x86/arch.py#L754 takes the last byte read, (which is not a prefix, so may be a rex) and if this is a rex it modify the instruciton value.
Maybe it's around there. We may modify this to do a second loop and parse while it's a REX.


If you want to try the modification, go for it @KKomarov.
If not, I may look t this by the end of the week

[...]

PR #1376, opened 2021-07-01
x86_64 Fix multiple REX prefix instruction disasm

KKomarov:
Example of problem instruction `mov qword ptr es:[rax], rdx` (`26 48 48 89 10` in hex where `48` is REX prefix)
Issue: https://github.com/cea-sec/miasm/issues/1375

### DIFF
--- a/miasm/arch/x86/arch.py
+++ b/miasm/arch/x86/arch.py
@@ -751,14 +751,18 @@ class mn_x86(cls_mn):
                 break
             pre_dis_info['prefix'] += c
             offset += 1
-        if mode == 64 and c in b'@ABCDEFGHIJKLMNO':
-            x = ord(c)
+        rex_prefixes = b'@ABCDEFGHIJKLMNO'
+        if mode == 64 and c in rex_prefixes:
+            while c in rex_prefixes:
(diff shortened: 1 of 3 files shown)


serpilliere:
Hey @KKomarov 
- The fix seems ok
- You have added regression tests
- You have put a comment in the code to explain the special case.
Everything if perfect :+1: 

Can you merge your commits into one commit ?

KKomarov:
@serpilliere Can we merge it with squash strategy? It will result into one commit in master

serpilliere:
Yes I could, but the commit will stack directly on the master.
Usually, we create a merge to have a 'merge commit' instead of a squash in order to show the fix in the commit tree.

KKomarov:
If you don't mind please do `merge with squash` - one commit with fine naming of PR title
At work people in their PRs often do tons of commits like "test test", "wip". If we merge such PRs without squash master history would become not readable.

serpilliere:
Yes, that's why I was asking you to merge your commits into only one commit.
And I prefer to have a clean fix *and* a merge pull request in order to keep the commit tree with a merge commit on top of your fix.
But never mind. I will squash it.

PR #1380, opened 2021-08-31
Fix kernel32_WideCharToMultiByte

tdrrdt:


serpilliere:
Thank you for the fix @tdrrdt !

PR #1382, opened 2021-09-03
Clean encode's value check

327135569:
Seems we can do it in one pleace in cpu.py. it simple than check value and mask in every encode() ?

### DIFF
--- a/miasm/arch/aarch64/arch.py
+++ b/miasm/arch/aarch64/arch.py
@@ -1797,8 +1797,6 @@ class op0_value(aarch64_uint64):
         v = self.encodeval(v)
         if v is False:
             return False
-        if v > self.lmask:
-            return False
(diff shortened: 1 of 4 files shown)


serpilliere:
Hi @327135569 
Yep, it seems to greatly simplify the code.
What makes me think twice what the fact that with this patch, the `self.value` is left 'unattended  until the return.
But yes, it seems ok after all.

327135569:
I am not very clear, is this mask check used to filter out mismatched mnemonics? If yes, I think it is necessary to do the check in each encode().  but now, not every encode() do this check.

serpilliere:
Yes, it's there to check if the encoded value is correct. That's why your fix seems legit

serpilliere:
Thank you for the PR @327135569 !

Issue #1396, opened 2021-11-12
"Concurrent access on same bit not allowed"


0xeb:
Trying to symbolically execute a block that starts with "pop sp" in a PE+ executable.

PR #1397, opened 2021-11-12
Fixed "POP reg_sp" in all modes

0xeb:
POP SP/ESP were broken in x64 mode.
Same for POP SP in x32 mode.
etc.

Now, we don't increment reg_sp at all if it is the target of the POP.

Fixes issue #1396


serpilliere:
Hi @0xeb !
The fix seems ok for me!
Why did you closed your pr ?

0xeb:
I did not close the PR, seems to have been closed automatically. Let me try to re-open it / re-do it again.

PR #1398, opened 2021-11-16
Fixed "POP reg_sp" in all modes

0xeb:
POP SP/ESP were broken in x64 mode.
Same for POP SP in x32 mode.
etc.

Now, we don't increment reg_sp at all if it is the target of the POP.


0xeb:
image

The PR was closed by itself and the auto CI is failing all the time for some reason.

0xeb:
@serpilliere please check now and merge if possible.

serpilliere:
Thank you for the fix @0xeb !

PR #1399, opened 2021-11-29
Fixed symbexec example to work for newer IDA (7.6)

0xeb:


serpilliere:
Thank you for the fix @0xeb !

PR #1403, opened 2021-12-06
Fix html; Add reg test

### LINKED ISSUE #1402: arm shift handling causes problems with the html output
in line 483 in arch/arm/arch.py is a special handling for shift expressions:
```python
if isinstance(s, ExprOp) and s.op in expr2shift_dct:
    s = ' '.join(
        str(x)
        for x in (
                color_expr_html(s.args[0], loc_db),
                utils.set_html_text_color(expr2shift_dct[s.op], utils.COLOR_OP),
                color_expr_html(s.args[1], loc_db)
        )
    )
```
this code converts `s` into a str but this `s` is later used in `color_expr_html` calls that expect a ExprOp. this leads to a ValueError in translator.py line 126.

serpilliere:
Fix #1402


tly000:
thanks, looks good to me 🙂

serpilliere:
Thank you for the issue @tly000 !

PR #1409, opened 2022-01-14
Implement syscall handler for mips32b + minimal Linux environment

icecr4ck:
Hello,

As mentioned in the title, this PR aims to implement a minimalist Linux environment for the mips32b architecture in order to handle syscalls.

There is also an implementation of the `socket` syscall (to give an example).


serpilliere:
Thank you @icecr4ck 
The only point I see is maybe the os doesn't set each ret value at each syscall, so we may opt them.
(like, in x86, eax is used to ret the first value, and optionally it can use edx to add more ret value, but edx may not be set if eax is sufficient)
Anyway: Thank you for the PR!

PR #1422, opened 2022-03-23
Unittest/Pytest compatibility

CLOVIS-AI:
This PR introduces compatibility with Unittest and PyTest without removing any functionalities (the old way of calling the tests is unchanged).


CLOVIS-AI:
Removed type hint

serpilliere:
Thank you for the update @CLOVIS-AI !

PR #1423, opened 2022-03-25
Fix memory breakpoint display in debugger

WilliamBruneau:


serpilliere:
Thank you @WilliamBruneau !

PR #1431, opened 2022-05-27
CI: add support for Python 3.10

commial:
This PR adds, in CI:
* Matrix test for Python 3.10
* use of LLVM 8 as expected by the currently used `llvmlite` version


serpilliere:
Thanks @commial !

Issue #1404, opened 2021-12-11
I can't seem to install it on Windows, getting errors


jt0dd:
From what I can tell it's looking for something in the directory "build" but that's not a standard directory. I'm confused on how to proceed. I tried this both on my bare metal Windows 11 machine and a Windows 10 VM. I've installed things like GCC with MinGW and added the bin to PATH. What's going wrong here?

```
PS C:\Users\jonat> python3 -m pip install miasm
Collecting miasm
  Downloading miasm-0.1.3.tar.gz (554 kB)
     |████████████████████████████████| 554 kB 2.2 MB/s
  Preparing metadata (setup.py) ... error
  ERROR: Command errored out with exit status 1:
   command: 'C:\Users\jonat\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\python.exe' -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'C:\\Users\\jonat\\AppData\\Local\\Temp\\pip-install-caz1p30p\\miasm_8310d01b47764a7bacbdaf01f6d6e702\\setup.py'"'"'; __file__='"'"'C:\\Users\\jonat\\AppData\\Local\\Temp\\pip-install-caz1p30p\\miasm_8310d01b47764a7bacbdaf01f6d6e702\\setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base 'C:\Users\jonat\AppData\Local\Temp\pip-pip-egg-info-ibw_wrr1'
       cwd: C:\Users\jonat\AppData\Local\Temp\pip-install-caz1p30p\miasm_8310d01b47764a7bacbdaf01f6d6e702\
  Complete output (19 lines):
  building
  build with 'all'
  running egg_info
  creating C:\Users\jonat\AppData\Local\Temp\pip-pip-egg-info-ibw_wrr1\miasm.egg-info
  writing C:\Users\jonat\AppData\Local\Temp\pip-pip-egg-info-ibw_wrr1\miasm.egg-info\PKG-INFO
[...]

PR #1435, opened 2022-10-14
Allow building on Windows without a Visual Studio command prompt

mrexodia:
This still requires you have Visual Studio and LLVM installed, but might prevent some issues like #1404

This was tested with LLVM versions 4.0.0-15.0.2, 32-bit python and 64-bit python. Additionally I added support for detecting Visual Studio's Clang-cl so you don't have to install LLVM separately.

### DIFF
--- a/miasm/jitter/vm_mngr.h
+++ b/miasm/jitter/vm_mngr.h
@@ -19,7 +19,9 @@
 #define CODENAT_H
 
 #if defined(_WIN32) || defined(_WIN64)
+#ifndef _CRT_SECURE_NO_WARNINGS
 #define _CRT_SECURE_NO_WARNINGS
+#endif
 #endif
 
 #if _WIN32
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ def win_get_llvm_reg():
     except FileNotFoundError:
       pass
     return winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE, REG_PATH, 0, winreg.KEY_READ)
-  
+
 def win_find_clang_path():
     try:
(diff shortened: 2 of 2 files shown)


serpilliere:
Hi @mrexodia !
Thanks for the fix. If we want to mimic the unix behavior, we can just let the jit compilation error: the only working jitter will then be the python one.

mrexodia:
I guess the issue is that setup.py will fail and therefore pip will abort the installation, even if technically it will “work”. Will try to fail compilation on linux and see what happens exactly.

There is still an issue with the linker, for some reason it appears lld-link is not compatible with Visual Studio’s link.exe command line on AppVeyor. Need some time to fix that.

mrexodia:
Yeah, it is my understanding that similar to clang-cl there should be a command line-compatible version of lld. Perhaps this is incorrect though, just weird it was working on my test environment…

mrexodia:
Still polishing this to output all the debug information necessary to diagnose future problems, but it should be pretty much ready.

I changed it so that on Windows it will first try to detect a “valid” build environment by checking if cl.exe and link.exe are in the PATH and it will not even attempt the build if not found. The user will get a warning with instructions to set up a suitable build environment, but the installation will succeed without native JIT module as you described.

mrexodia:
Is python 2.7 still supported by miasm or is it an old test?

serpilliere:
No, and I will remove the python2.7 build & tests to avoid  errors on regression tests

serpilliere:
Oh @mrexodia !
Thanks for the "Seek & Destroy" patch :smile:

serpilliere:
Thank you for the fix @mrexodia !

PR #1437, opened 2022-11-03
Remove Python 2.7 from CI

mrexodia:
It would be nice if Miasm could slowly migrate to more modern python (f-strings, type annotations). According to earlier discussions 2.7 isn't supported anymore so I took the liberty to remove it from CI.

Also a discussion about the minimal Python 3 version would be good. There's nice updates to typing in more recent versions.


serpilliere:
I agree with you: It's time to drop python2.7!

PR #1453, opened 2023-07-08
fix compilation on cygwin

frank2:
Hello! I am unsure if it was intentional not to target Cygwin, but based on what the modified line was targeting, it appears the line is attempting to detect compilation environments. Cygwin is a Windows-based Unix-like POSIX environment. So while it compiles on Windows (i.e.. `defined(_WIN32)`, for MSVC), it ultimately gets compiled with Unix tools like GCC or Clang on a Windows environment, and can be detected with the `__CYGWIN__` preprocessor define.

Compilation of the C Python bindings for miasm fails because it does not detect the Cygwin environment on compilation. The fix is pretty simple. Miasm is working great on my system now.

Let me know if it was intentional not to target Cygwin, I can see why you might not want to have so many platforms to support on your plate, but if you're targetting MinGW32, I don't see why you couldn't also support Cygwin.

### DIFF
--- a/miasm/runtime/int_lib.h
+++ b/miasm/runtime/int_lib.h
@@ -48,7 +48,7 @@
 #define XSTR(a) STR(a)
 #define SYMBOL_NAME(name) XSTR(__USER_LABEL_PREFIX__) #name
 
-#if defined(__ELF__) || defined(__MINGW32__) || defined(__wasm__)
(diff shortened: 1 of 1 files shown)


serpilliere:
Thank you for the fix @frank2 !
(and sorry for the delay)

Issue #1456, opened 2023-09-05
Implement carry flag for basic shifters


W0ni:
Modified instructions:
- ASRS, LSLS, LSRS, RORS, RRXS
- MOVS, MOVNS with shifter

Maybe is it possible to factor all the computations in one "update_shift_c" function? Or we have to split each shift type.

See: https://developer.arm.com/documentation/ddi0597/2023-06/Base-Instructions/MOV--MOVS--register---Move--register--?lang=en

serpilliere:
Hi @W0ni 
Maybe I have missed something, but it seems the flags are present and modified for those instructions:
in miasm/arch/arm/sem.py:
```python

def rors(ir, instr, a, b):
    e = []
    r = ExprOp(">>>", a, b)
    e.append(ExprAssign(a, r))

    e += [ExprAssign(zf, ExprOp('FLAG_EQ', r))]
    e += update_flag_nf(r)

    dst = get_dst(a)
    if dst is not None:
        e.append(ExprAssign(ir.IRDst, r))
    return e, []
...

def asrs(ir, instr, a, b, c=None):
    e = []
    if c is None:
        b, c = a, b
    r = ExprOp("a>>", b, c)
    e.append(ExprAssign(a, r))

    e += [ExprAssign(zf, ExprOp('FLAG_EQ', r))]
    e += update_flag_nf(r)

    dst = get_dst(a)
    if dst is not None:
        e.append(ExprAssign(ir.IRDst, r))
    return e, []
...
```

W0ni:
Hey @serpilliere !

Indeed, the N and Z flags are correctly modified by this code. However, these instructions should also modify the C flag according to this documentation: https://developer.arm.com/documentation/ddi0597/2023-06/Base-Instructions/MOV--MOVS--register---Move--register--?lang=en

If the shift is 0, the carry flag is not affected. If it's not 0, a new block is required to compute the carry flag.

PR #1457, opened 2023-09-06
[ARM] compute cf for shift/rotate

W0ni:
This partially addresses #1456

- Moves rrx computation inside the functions, this allows to do the carry flag computation related to RRX
- Adds a shifter/rotate template for the carry flag similar to the one in x86

It is still (at least) left to:
- compute the carry flag of some instructions with immediates.
Here is an example A32ExpandImm_C (https://developer.arm.com/documentation/ddi0597/2023-06/Shared-Pseudocode/aarch32-functions-common?lang=en#impl-aarch32.A32ExpandImm_C.2). Used by ANDS, BICS, EORS, MOVS, MVNS, ORRS, TEQ, TST with immediates.
The equation is something like `cf= ROR(zExt32(imm[0:8], imm[8:12] << 1).MSB()` (if `2*imm[8:12] != 0`). This is tricky because the semantics don't know the rotate amount that was used. Maybe the additional_info object on instructions could store useful data like rotate amount and maybe the encoding category (A1, A2, T1, T2...)?

- ~~Update the semantic tests to include the carry flag.~~

Related additional work:
- Maybe we could also clean up a bit the code (use flag functions more often, replace isinstance for Expr* by is_XXX...)
- Maybe handle the exception when setflags AND dst=PC?

### DIFF
--- a/miasm/arch/arm/arch.py
+++ b/miasm/arch/arm/arch.py
@@ -1148,8 +1148,12 @@ class arm_op2(arm_arg):
             shift_op = ExprInt(amount, 32)
         a = regs_expr[rm]
         if shift_op == ExprInt(0, 32):
+            #rrx
             if shift_type == 3:
                 self.expr = ExprOp(allshifts[4], a)
+            #asr, lsr
+            elif shift_type == 1 or shift_type == 2:
+                self.expr = ExprOp(allshifts[shift_type], a, ExprInt(32, 32))
             else:
                 self.expr = a
         else:
--- a/miasm/arch/arm/sem.py
+++ b/miasm/arch/arm/sem.py
@@ -405,6 +405,8 @@ def adc(ir, instr, a, b, c=None):
     e = []
     if c is None:
         b, c = a, b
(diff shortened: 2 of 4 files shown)


serpilliere:
Hi @W0ni,
Thanks for the PR: We will look into it soon!
(nice tipo catch on the "compote".)

W0ni:
Hello @serpilliere ,
While adding the tests for MOVS, I noticed a bug in the disassembly of the shifters. I fixed it in the commit 71e4e09da6816a72905adeea3c0864714353c406 along with the related tests.
I also added the tests for the MOVS instuctions in 1885f3c.

serpilliere:
Hi @W0ni !

I was looking at your patch regarding the disassembling fix, more precisely the support of ASR/LSR and RRX during instruction decoding. I think we should add a regression test on the disassemble engine itself (test/arm/arch.py), in order to test the freshly added code.

Do you want to add it to the PR, or do you prefer I add it?
(I was thinking of an asm/disasm of the fixed versions)

PR #1464, opened 2024-01-06
Python 3.12 support

mrexodia:
The main issue is the removal of the deprecated `imp` module (I replaced it with `importlib.util` and matched the existing behavior).

Other than that I made a few minor fixes of warnings, most notably fixing some broken regular expressions by switching to regex literals.


serpilliere:
Hey, thank you @mrexodia !

PR #1465, opened 2024-01-08
document windows compilation

gogo2464:


mrexodia:
Using setup.py directly is now deprecated by Python. Everything should work if you run `pip install .` from a developer prompt.

gogo2464:
my bad! I can fix it

serpilliere:
I think we have someone who is currently updating to the setup.py replacement

mrexodia:
Technically it’s not necessary to get rid of setup.py (and honestly I wouldn’t recommend it). I tried locally to create a spec-compliant project.toml that uses the current setup.py, but everything exploded because Python doesn’t properly support some of the constructs used anymore 🥲

Redhpm:
Since there is a better build system and a gh action for windows now (thanks to mrexodia here), I don't think this is relevant now.

PR #1468, opened 2024-01-16
Improve CreateFile flags parsing

W0ni:
Add flag 2 for write, similarly to flag 1 for read.
Fix open mode according to flags.


serpilliere:
Thank you for the fix @W0ni !!

PR #1474, opened 2024-03-18
Fix typos found by codespell

DimitriPapadopoulos:


miasm/analysis/data_flow.py
```diff
@@ -1908,9 +1908,9 @@ def __ne__(self, other):
         # required Python 2.7.14
         return not self == other
 
-    def may_interfer(self, dsts, src):
+    def may_interfere(self, dsts, src):
```
mrexodia, line 1911: I think for this function you should provide a wrapper with the old name (marked as deprecated). This ensures any usages won’t straight up break…

serpilliere:
Thank you for the fix @DimitriPapadopoulos !

PR #1476, opened 2024-03-18
Fix typo found by codespell - API breaking change

DimitriPapadopoulos:


serpilliere:
Hi @DimitriPapadopoulos !
Thanks for the patch!
As @mrexodia said, I think we may add:
```python
import warnings
....
    def may_interfer(self, dsts, src):
        warnings.warn("This api is deprecated, use `may_interfere` instead")
        return self.may_interfere(dsts, src)
```
Can you add it to the PR or may I fir your PR @DimitriPapadopoulos ?

serpilliere:
@DimitriPapadopoulos : Can you rebase this pr?
(I think I should have merged this one *before* the tipo fixing :sob: )

PR #1487, opened 2024-04-27
Fix: `bin_stream::get_u8` method returning bytes

psyirius:
`bin_stream::get_u8` supposed to be returning an int.
but instead it returns a bytes object

> Not sure how the dependent callers are expecting!


psyirius:
@serpilliere Is there any way you can review it and merge as soon as possible.
Workaround code for essential stuff is not gonna look good in production. :)

serpilliere:
Arg! I forgot this one Thank you for the fix @psyirius !!!

PR #1496, opened 2024-08-03
Fix typo in aarch64_immhi_page decode

### LINKED ISSUE #1495: Failing to assemble ADR X29 0xFFFFFFFFFFFFFAC8 due to missing offset
I am trying to modify a block inside an arm64 binary.
My steps are as follows:

1. Load the binary inside miasm
```python
cont = Container.from_stream(open('blabla.so', 'rb'), loc_db=location_db)
dis_engine = machine.dis_engine(cont.bin_stream, loc_db=location_db)
```
3. Disassemble blocks
```python
asmcfg = dis_engine.dis_multiblock(...)
```
4. Attempt to reassemble the blocks (without any modifications)
```python
dinterval = interval(blk.get_range() for blk in asmcfg.blocks)
dinterval_full = interval([dinterval.hull()])

assembled = asm_resolve_final(machine.mn, asmcfg, dst_interval=dinterval_full)
```
5. Error
```
site-packages\miasm\core\cpu.py", line 1427, in asm
    raise ValueError(
ValueError: cannot asm 'ADR' ['X29', '0xFFFFFFFFFFFFFAC8']
```

To give some context, this is the instruction it is failing on.

```asm
0xC7100: ADR X29, sub_C6BC8 ; Instruction in hex [5D D6 FF 10]
```

The instruction `adr x29, #0xfffffffffffffac8` would be correct, if it were placed at offset `0x0`. 
However it is assembled for the location `0xC7100` and should be `adr x29, #0xc6bc8`.

Looking through the miasm codebase I can not see any support for this.
Am I missing something?

AeonLucid:
This fixes a typo in the decoding of `aarch64_immhi_page`.
I assume this code was copied from `aarch64_immhip_page` and someone forgot to change the shift afterwards.

The value `524020` (and immlo.value `0`) decodes to `ExprInt(0xFFFFFFFFFFFFFAC8, 64)`.

Before it would fail to re-encode this due to incorrectly assuming an overflow (`v > (1 << 19) - 1`).
After the change it re-encodes back to `524020`.

---

Fixes #1495.

### DIFF
--- a/miasm/arch/aarch64/arch.py
+++ b/miasm/arch/aarch64/arch.py
@@ -1424,7 +1424,7 @@ class aarch64_immhi_page(aarch64_imm_32):
     def encode(self):
         v = int(self.expr)
         if v & (1 << 63):
-            v &= (1 << 33) - 1
+            v &= (1 << 21) - 1
         self.parent.immlo.value = v & 3
         v >>= 2
         if v > (1 << 19) - 1:
--- a/test/arch/aarch64/arch.py
+++ b/test/arch/aarch64/arch.py
@@ -1846,7 +1846,10 @@ reg_tests_aarch64 = [
     ("XXXXXXXX    TLBI       0x0, c7, 0x0, XZR",
      "1F8708D5"),
     ("XXXXXXXX    YIELD      ",
-     "3F2003D5")
+     "3F2003D5"),
+
+    ("XXXXXXXX    ADR        X29, 0xFFFFFFFFFFFFFAC8",
+     "5DD6FF10"),
 ]
 
 


serpilliere:
Hi @AeonLucid 
Thank you for the fix.
I saw in the issue that you have an example:
```
0xC7100: ADR X29, sub_C6BC8 ; Instruction in hex [5D D6 FF 10]
```
Can you add it to the regression test? Or may and it to this pr?

AeonLucid:
> Hi @AeonLucid 
> 
> Thank you for the fix.
> 
> I saw in the issue that you have an example:
> 
> ```
> 
> 0xC7100: ADR X29, sub_C6BC8 ; Instruction in hex [5D D6 FF 10]
> 
> ```
> 
> Can you add it to the regression test? Or may and it to this pr?
> 
> 

I couldn't find any regression tests where I could add this instruction to. Can you point me to them?

serpilliere:
Hi @AeonLucid 

There are asm/disasm regressions tests in `/miasm/test/arch/aarch64/arch.py`


For example
```python
reg_tests_aarch64 = [

    ("XXXXXXXX    MOV        W1, WZR",
     "E1031F2A"),
...
```

which means "MOV W1, WZR" gives bytes "E1031F2A", and vice versa

Tell me if you want to give it a try (I will add it in the other case)

Thankt you!

AeonLucid:
I have added a regression test, let me know if you want any other changes. @serpilliere

serpilliere:
Thank you for the fix @AeonLucid !!

PR #1501, opened 2024-08-13
Fixes #1500

### LINKED ISSUE #1500: Test cases status - Windows
after successfully bulding the full package (including llvm) I did run some subset under windows
the quesitons are
1) why do we have different results when running with default options and using sequential run (-m flag)
2) here is the failed TC when running with -m

```
c:\Dev\re\miasm\test>python test_all.py -o python -m
 TEST\ARCH x86/sem.py python
DONE x86/sem.py python 0s
 TEST\ARCH x86/unit/mn_stack.py python
DONE x86/unit/mn_stack.py python 1s
 TEST\ARCH x86/unit/mn_das.py python
DONE x86/unit/mn_das.py python 3s
 TEST\ARCH x86/unit/mn_pshufb.py python
DONE x86/unit/mn_pshufb.py python 0s

....

Result: 136/137 pass
ERROR jitcore.py python
b'Traceback (most recent call last):\r\n  File "c:\\Dev\\re\\miasm\\test\\jitter\\jitcore.py", line 12, in <module>\r\n    jitter.cpu.RAX = -1\r\n    ^^^^^^^^^^^^^^\r\nTypeError: Arg too big for uint64_t\r\n'

```

I'm curious how come that this speicific TC fails... is -1 value bigger then uint64_t type?

greenozon:
ref to issue #1500 for details

the added code makes the negative values  unsigned integers (64 or 32 bit) thus allowing to set it into uint64_t value of RAX/EAX

### DIFF
--- a/test/jitter/jitcore.py
+++ b/test/jitter/jitcore.py
@@ -9,16 +9,16 @@ jitter = machine.jitter(loc_db, sys.argv[1])
 jitter.cpu.RAX = 16565615892967251934
 assert jitter.cpu.RAX == 16565615892967251934
 
-jitter.cpu.RAX = -1
+jitter.cpu.RAX = -1 & 0xffffffffffffffff
 assert jitter.cpu.RAX == 0xffffffffffffffff
 
-jitter.cpu.RAX = -2
+jitter.cpu.RAX = -2 & 0xffffffffffffffff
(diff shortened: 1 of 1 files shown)


serpilliere:
Thank you for the fix @greenozon !

PR #1502, opened 2024-08-19
fix expression division

W0ni:
`__div__` is not bound to the division operator in Python3. This leads to an error when trying to divide expressions.


serpilliere:
Thank you @W0ni !

PR #1504, opened 2024-12-03
add ldaxr 

0xjacklove:
Add the ldaxr instruction.


serpilliere:
Hi @0xjacklove !
Thank you for the patch!
If I dont mess up, in your code the ExprMem has already the size of arg1.size, so you don't need to zeroExtend it.
Do you agree ?

0xjacklove:
I agree with you.

0xjacklove:
But I wrote it by imitating the function def ldxr(ir, instr, arg1, arg2):

PR #1515, opened 2025-12-12
remove some unnecessary files

Jvlegod:
1. example: remove unnecessary result files
2. .gitignore: add .bin and .exe files


Jvlegod:
There were some result files that appeared to be in the source code, which I deleted. I also modified the .gitignore file.

serpilliere:
Hi @Jvlegod 
If I don't make mistakes, files (simple_test.bin & x86_32_sc.bin) are not *results* but raw files used in regression tests
So we need to keep them in the git tree

For example, we can see the dependencies here:
```bash
miasm]$ git grep x86_32_sc.bin
test/jitter/jitload.py:# Initial data: from 'example/samples/x86_32_sc.bin'
test/test_all.py:for script, dep in [(["x86_32.py", Example.get_sample("x86_32_sc.bin")], []),
```

Jvlegod:
yes, I think you are right.

Jvlegod:
hi @serpilliere
also, how about this PR https://github.com/cea-sec/miasm/pull/1516

What are your future plans for software? I'd love to know! :)

PR #1519, opened 2026-01-30
adjust to removed APIs in ast module for python >=3.14

### LINKED ISSUE #1518: Compatibility with python >= 3.14
Starting with python 3.14 the deprecated classes  `ast.Str` and `ast.Num` have been removed. They are currently in use in `miasm/core/sembuilder.py`

Switching to using `ast.Constant` will, however, likely break compatibility with python versions lower than 3.8, which started representing all constants as `ast.Constant`.

I'm willing to provide a PR, but I'm not sure what the goal regarding compatibility is for newer miasm versions.

fkil:
ast.Num and ast.Str were deprecated with python 3.8 and are now removed with python3.14.

This commit replaces uses of these APIs using ast.Constant. Consequentely, the behavior for python versions <3.8 will break

Resolves https://github.com/cea-sec/miasm/issues/1518

### DIFF
--- a/miasm/core/sembuilder.py
+++ b/miasm/core/sembuilder.py
@@ -9,6 +9,8 @@ from future.utils import PY3
 import miasm.expression.expression as m2_expr
 from miasm.ir.ir import IRBlock, AssignBlock
 
+def _is_ast_string(value):
+    return isinstance(value, ast.Constant) and isinstance(value.value, str)
 
 class MiasmTransformer(ast.NodeTransformer):
     """AST visitor translating DSL to Miasm expression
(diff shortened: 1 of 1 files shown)


crimsonskylark:
There are also problems due to the incompatibility of `python-future` with >=3.13 used [here](https://github.com/cea-sec/miasm/blob/2a15c60712b326b541d42ac48f372f97023547e7/miasm/core/utils.py#L16), amongst other places. 

It would be good to have some feedback from the developers with regards to what types of changes, breaking or otherwise, are acceptable for PRs.

serpilliere:
Hi @fkil  & @crimsonskylark 

Yep, maybe it's time to drop some "old" versions of python.
Maybe we can consider the reverse of fun binaries is done on "full featured" distribution :smile:

mrexodia:
https://endoflife.date/python:

image

I would drop everything below 3.10 honestly 🤷🏻‍♂️

fkil:
I've looked into the failing regression test, and the cause seems to be an outdated z3 version. Updating to the most recent z3 version in optional_requirements.txt passes the test.

I've also locally ran the same tests for python3.14 and they pass as well, with a small limitation: llvmlite does not provide wheels for python3.14 and its `setup.py` is not compatible with the most recent `setuptools` version (which is used by default with pip). As a workaround I installed an older setuptools (>=v81.0 breaks, so I used v80.10.2) and then use the `--no-build-isolation` option of pip. I feel like it would be pretty unclean to add these to the tester though.

Should I bump the z3 version in this PR as well, or should that be a separate one?

fkil:
@crimsonskylark Regarding python-future: While newer versions aren't officially supported, they do not break 3.13 or 3.14 yet. While the dependency should probably be removed, I believe it should be a separate issue and patch set.

crimsonskylark:
Yes, of course - I didn't mean to hijack your PR. Sorry!

serpilliere:
I agree with you @mrexodia : dropping everything below 3.10 is fine.
I did some login to some machine here and there, and there are a lot of machines that still have python 3.11 (I didn't find 3.10 by the way)

Maybe we can bump the z3 in another PR.

Thank you for the tests @fkil

PR #1520, opened 2026-02-14
Add support for VSCode Dev Containers

crimsonskylark:
This commit add supports for building a dev container for using miasm inside a Ubuntu-based Docker container. This decouples miasm usage from the user's host system and should prevent issues amongst systems by using pinned dependencies instead.

Note that the dev container feature requires docker/podman installed and running.

I have tested these changes on the following systems:

- Manjaro (Kernel 6.12.64-1)
- Windows 11 (25H2 26200.7840)

On both machines, all tests except `unpack_upx.py` pass. It seems to be stuck somewhere. I have not yet debugged why that is, but it seems unrelated to the container.


mrexodia:
There is this magic thing called `uv`, you might enjoy it 😄

crimsonskylark:
Did not work well on my system. This was the lowest friction solution I found (albeit not very good, as it requires both Docker and a specific editor) for my use case.

If there's no interest in having it on the main project I can close the PR. It's not a problem.

crimsonskylark:
Nevermind. Tried again after updating my system and it built miasm with no problems.

Thanks for the heads-up.

PR #1521, opened 2026-03-14
Fix outofssa algorithm: lost update in insert_parallel_copy and insructions after conditional jmp

Explosiontime202:
I have noticed that for certain larger IRCFGs the `UnSSADiGraph` produced an invalid output. This is caused by `insert_parallel_copy` in certain constellations as it would add parallel copies to the end of a `IRBlock` in one loop iteration and then in a later one add parallel copies to the start of the same `IRBlock`. But as the iteration is over the `IRBlock`s itself, the first update was lost.

Secondly, I noticed that the code produces the following sequence of instructions at the end of a basic block:
```
IRDst = A?(loc_1, loc_2)
C = D
```
As i would interpret the IRDst assignment as a jump, this seems not correct. Please correct me there if the semantics of the `IRDst` assignments differ from my assumptions.
As mentioned in the paper this algorithm stems from, if we need to insert parallel copies before the conditional jump instructions, we need to check for interference with the condition computation. As for a quick fix, this is currently done with an `assert`, which is less than ideal. But I don't know how to implement that interference check into the given algorithm.

### DIFF
--- a/miasm/analysis/outofssa.py
+++ b/miasm/analysis/outofssa.py
@@ -5,7 +5,6 @@ from miasm.ir.ir import IRBlock, AssignBlock
 from miasm.analysis.ssa import get_phi_sources_parent_block, \
     irblock_has_phi
 
-import logging
 
 class Varinfo(object):
     """Store liveness information for a variable"""
@@ -89,13 +88,7 @@ class UnSSADiGraph(object):
             for parent, parallel_copies in viewitems(parent_to_parallel_copies):
                 parent = ircfg.blocks[parent]
                 assignblks = list(parent)
-                # insert before IRDst, TODO: might need to check for interference as mentioned in the paper
-                # sanity checks for now:
-                jmp_block = parent[-1]
-                for jmp_src in jmp_block.values():
-                    jmp_read_values = set(filter(lambda expr: expr.is_id(), jmp_src.get_r(mem_read=True)))
-                    assert len(jmp_read_values & set(parallel_copies.keys())) == 0, "Interference between jmp instruction read set and parallel copy write set"
-                assignblks.insert(-1, AssignBlock(parallel_copies, jmp_block.instr))
+                assignblks.append(AssignBlock(parallel_copies, parent[-1].instr))
                 new_irblock = IRBlock(parent.loc_db, parent.loc_key, assignblks)
                 ircfg.blocks[parent.loc_key] = new_irblock
 
--- a/miasm/analysis/outofssa.py
+++ b/miasm/analysis/outofssa.py
@@ -5,6 +5,7 @@ from miasm.ir.ir import IRBlock, AssignBlock
 from miasm.analysis.ssa import get_phi_sources_parent_block, \
     irblock_has_phi
 
+import logging
 
 class Varinfo(object):
     """Store liveness information for a variable"""
@@ -61,7 +62,8 @@ class UnSSADiGraph(object):
         """
         ircfg = self.ssa.graph
 
-        for irblock in list(viewvalues(ircfg.blocks)):
+        for block_loc in list(ircfg.blocks.keys()):
+            irblock = ircfg.get_block(block_loc)
             if not irblock_has_phi(irblock):
                 continue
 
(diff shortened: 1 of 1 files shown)


serpilliere:
Hi @Explosiontime202 !

The assignment of IRDst is here "only an assignment". Here are the rules:
- every irblock has exactly one assignment to IRDst
- the assignment can be anywhere in the irblock

So you can have other assignments later in the irblock: this is legit.

For the first point, have you a little example to show the faulty behavior?

Explosiontime202:
Thanks for your quick answer :) 
Then I am going to revert the change regarding the `IRDst` assignment and manage ensure that invariant in another way (because I need that for my stuff to work correctly :) )

Regarding an example: I have an example which triggers this, but it is quite lengthy. I put it below, but I am now sure, how it is helpful to you. The difference in the example is in `loc_1c9b0f` where `R11.6 = R11.54` is missing.
I think that the explanation is easier to understand: In one iteration inside the `insert_parallel_copy`, a copy is added to a `IRBlock` and the block is added to the `IRCFG`. In an iteration following it, another copy is inserted into the same block (for example because it is the parent of another block). But the copy is added to the old version of the block, which is then added to the `IRCFG` and thereby loosing the previous copy instruction.

Before OutOfSSA:
example_bug_before

After correct OutOfSSA:
example_bug_after

After buggy OutOfSSA:
example_bug_after_buggy


here is also a diff between the `.dot` files (left fixed vs. right buggy): https://www.diffchecker.com/39K40dRm/
(this still includes my patch to move the assignments before IRDst assignments)

miasm/analysis/outofssa.py
Redhpm, line 1: Why not just `for irblock in ircfg.blocks.values()` ?
How does that fix anything ? I do prefer this way of writing but it should be equivalent functionally.

Explosiontime202: I also prefer this way of writing it, but it is just not the same, because updating a dict while iterating over a view over it (which .values() is) is a no-no.
Example:
```Python
d = {1: 1}
for e in d.values():
    d[2] = 2
```
results in:
`RuntimeError: dictionary changed size during iteration`

The fix that we no longer use the stale block when updating it and inserting it back into the blocks dict.

Redhpm: Ah ok, I cannot verify now but shouldn't `list(d.values())` work then ?

Explosiontime202: Then we have the same buggy code as before. Note that `viewvalues()` is the Python2 compatible way of `.values()`. So, no.

Redhpm: Correct me if I'm wrong but the only way 

```python
for v in list(d.values()):
```

and 

```python
for k in list(d):
    v = k.get(d)
```

are different is if multiple keys map to the same value.

That would mean the fix is applying the alg possibly multiple times to the same IRBlock ?
In your example, do you have multiple keys to the same block ?

Redhpm: Failed to see that at first

Redhpm: Or the ordering, might be a weird bug like that

Explosiontime202: No, we are not iterating over blocks multiple times. But in every iteration we might update more than one block: We update the current block and its parent blocks.
Therefore, when we iterate over the list of IRBlocks we do not see the updates to the parent blocks in previous iterations and thus drop these previous updates.

So to present the correctly situation with the snippets that you gave:
```Python
d = {1: 1, 2: 2, 3: 3}
for v in list(d.values()):
  print(v)
  d[v+ 1] = 8
```

vs.

```Python
d = {1: 1, 2: 2, 3: 3}
for k in list(d):
  v = k.get(d)
  print(v)    
  d[v + 1] = 8
```

These do not produce the same output.

There should never be multiple keys to the same block, this would be an invalid IR. (and there aren't, as far as I am aware)

Redhpm: Ah right, maybe we should add a comment to explain the behavior as it's a bit subtle

Explosiontime202: I mean we can, but it is pretty much standard behavior and common knowledge, I'd say, and already used in more placed within the codebase.
I'll updated the PR to add the comment.

PR #1530, opened 2026-05-31
expression: weak-reference the singleton intern table to bound memory

mrexodia:
Claude found this when my deobfuscation script was growing to 50+GB memory usage, works really well!


serpilliere:
Hey @mrexodia !
I don't see any downsides here so let's merge this

PR #1536, opened 2026-09-14
Apply relocs for ifuncs

kStor2poche:
This implements ifunc relocation for ELFs containing them by running the ifunc on a jitter that copies the currently-being-loaded file's memory.
While this should technically be the job of the libc init functions, it made sense for my application to apply this relocation along with the other types of relocation.
This only works for ifuncs in statically and dynamically loaded executables, as per the [glibc manual](https://sourceware.org/glibc/manual/latest/html_node/Indirect-Functions.html#When-IFUNC-Resolvers-Run).

As it is a bit intensive, it could make sense to put it behind a separate option in `vm_load_eld`, this is up for debate.


miasm/jitter/loader/elf.py
```diff
@@ -244,13 +252,28 @@ def apply_reloc_x86(elf, vm, section, base_addr, loc_db):
...
+            end_addr = 0x1337beef
+            def _code_sentinelle(jitter):
+                jitter.running = False
+                return False
+            ifunc_jitter.add_breakpoint(end_addr, _code_sentinelle)
+            if elf.size == 32:
+                ifunc_jitter.push_uint32_t(end_addr)
+            elif elf.size == 64:
+                ifunc_jitter.push_uint64_t(end_addr)
+
```
Redhpm, line 272: Maybe add `else` with error

miasm/jitter/loader/elf.py
```diff
@@ -206,14 +208,20 @@ def apply_reloc_x86(elf, vm, section, base_addr, loc_db):
...
+            else:
+                # B + A
+                addr = base_addr + r_addend
         elif reloc.type == elf_csts.R_X86_64_64:
             # S + A
-            addr_symb = loc_db.get_name_offset(symbol_entry.name)
+            addr_symb = loc_db.get_name_offset(symbol_entry.name.decode())
             if addr_symb is None:
-                log.warning("Unable to find symbol %r" % symbol_entry.name)
+                log.warning("Unable to find symbol %r" % symbol_entry.name.decode())
```
Redhpm: Since we're at it, just like you did further down, might as well switch to an f-string

Redhpm:
And a more general comment, maybe we should hide this behind a parameter like `run_ifuncs=False` for the `apply_reloc_x86` function, and another one for the `vm_load_elf` function.

We also need a test, as it needs quite realistic examples, might as well add a full on script in `example/jitter` with a real sample

miasm/jitter/loader/elf.py
Redhpm: since it's a logger, we can also keep the `%r` syntax and put the name as an argument

kStor2poche: Sure, I switched to f-strings right now but if you feel like % encoded in logs is more idiomatic in this project I can to that too !

PR #1537, opened 2026-09-14
Allow simplifying the following expr: {@n[id], @m[id][n:m]} with m>n

kStor2poche:
As per the title of the PR, it allows the simplification of `{..., @n[id], @o[id][n:m], ...}` with `o>m>n` into `{..., @m[id], ...}` if `o==m` and `{..., @o[id][0:m], ...}` otherwise.


Redhpm:
We should add a unit test for this in test/expression/simplifications.py

